PR #163, opened 2024-08-15
docs: add link to contact page

BogdanDanilescu:
image

### DIFF
--- a/apps/docs/src/app/contact/page.tsx
+++ b/apps/docs/src/app/contact/page.tsx
@@ -1,11 +1,19 @@
 import { Heading, Paragraph } from '@govie-react/ds';
+import { config } from '@/lib/config';
 
 export default function ContactPage() {
   return (
     <div>
       <Heading>Contact</Heading>
       <Paragraph>
-        You can contact the design system team via email. Email coming soon.
+        You can contact the design system team using{' '}
+        <a
+          target="_blank"
+          className="underline text-blue-700 hover:decoration-md"
+          href={config.signUpFormUrl}
+        >
+          this form.
+        </a>
       </Paragraph>
     </div>
   );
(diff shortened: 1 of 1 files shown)


apps/docs/src/app/contact/page.tsx
```diff
@@ -1,11 +1,19 @@
...
       <Heading>Contact</Heading>
       <Paragraph>
-        You can contact the design system team via email. Email coming soon.
+        You can contact the design system team using{' '}
+        <a
+          target="_blank"
+          className="underline text-blue-700 hover:decoration-md"
+          href={config.signUpFormUrl}
+        >
+          this form.
```
ianlnf, line 15: period can be moved outside of link

apps/docs/src/app/contact/page.tsx
```diff
@@ -1,11 +1,19 @@
...
+import { config } from '@/lib/config';
 
 export default function ContactPage() {
   return (
     <div>
       <Heading>Contact</Heading>
       <Paragraph>
-        You can contact the design system team via email. Email coming soon.
+        You can contact the design system team using{' '}
+        <a
```
ianlnf, line 10: could add a TODO to replace with the `Link` component when it supports target (for styling consistency)

hamza14khan, line 10: nitpick: You can use `<Link>` component from `govie-react/ds`.

BogdanDanilescu: Good idea. Though if we use the `Link` component from `govie-react/ds` we won't be able to open the form in a new tab as it doesn't have a `target` prop yet

BogdanDanilescu: maybe we should add a TODO as Ian was suggesting

ianlnf (approved):

PR #165, opened 2024-08-15
ci: add uat and staging environments

hamza14khan:
Addition of `uat` and `staging` environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,17 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Select deployment environment"
+        required: true
+        type: choice
+        options:
+          - uat
+          - staging
+          - production
+        default: uat
 
 permissions:
   contents: write
@@ -25,7 +36,7 @@ jobs:
       id-token: write
       contents: read
 
-    environment: dev
+    environment: development
 
     steps:
       - name: Checkout code
@@ -43,6 +54,9 @@ jobs:
         run: |
           git config user.name "${GITHUB_ACTOR}"
           git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+      - name: Set environment variable
+        run: echo "CONFIGURATION=${{ github.event.inputs.environment }}" >> $GITHUB_ENV
 
       - name: Build and test
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,17 +7,6 @@ on:
   pull_request:
     branches:
       - main
-  workflow_dispatch:
-    inputs:
-      environment:
-        description: "Select deployment environment"
-        required: true
-        type: choice
-        options:
-          - uat
-          - staging
-          - production
-        default: uat
 
 permissions:
   contents: write
@@ -54,9 +43,6 @@ jobs:
         run: |
           git config user.name "${GITHUB_ACTOR}"
           git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
-
-      - name: Set environment variable
-        run: echo "CONFIGURATION=${{ github.event.inputs.environment }}" >> $GITHUB_ENV
 
       - name: Build and test
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       id-token: write
       contents: read
 
-    environment: development
+    environment: dev
 
     steps:
       - name: Checkout code
--- a/.github/workflows/deploy-docs-prod.yml
+++ b/.github/workflows/deploy-docs-prod.yml
@@ -3,10 +3,15 @@ name: Deploy docs to Production
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        description: "Git tag to deploy"
+      environment:
+        description: "Select deployment environment"
         required: true
-        default: "last-release"
+        type: choice
+        options:
+          - staging
+          - uat
+          - production
+        default: uat
 
 jobs:
   build:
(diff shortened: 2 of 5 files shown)


apps/docs/project.json
```diff
@@ -34,6 +34,22 @@
...
+          "push": true,
+          "build-args": ["DEPLOY_ENV=uat"],
+          "tags": [
+            "975050122570.dkr.ecr.eu-west-1.amazonaws.com/life-events-design-system:uat",
+            "975050122570.dkr.ecr.eu-west-1.amazonaws.com/life-events-design-system:latest"
+          ]
+        },
+        "staging": {
+          "push": true,
+          "build-args": ["DEPLOY_ENV=staging"],
```
hamza14khan, line 48: Should this be `sta` instead of `staging`?

@ianlnf

apps/docs/src/lib/config.ts
```diff
@@ -4,16 +4,26 @@ function isProduction() {
...
+function isUat() {
+  return process.env.NODE_ENV === ('uat' as any);
+}
+
+function isStaging() {
+  return process.env.NODE_ENV === ('staging' as any);
+}
+
 export type DocumentSiteConfiguration = {
   isProduction: () => boolean;
```
ianlnf, line 16: These helpers could be removed, as they are not currently used, `isProduction`, `isUat`, `isStaging`

apps/docs/project.json
```diff
@@ -34,6 +34,22 @@
             "891377307993.dkr.ecr.eu-west-1.amazonaws.com/life-events-design-system:prd",
             "891377307993.dkr.ecr.eu-west-1.amazonaws.com/life-events-design-system:latest"
           ]
+        },
+        "uat": {
```
ianlnf, line 38: can define `staging` before `uat`, so it matches order of deployment (staging deployments before uat)

apps/docs/src/lib/config.ts
```diff
@@ -4,16 +4,26 @@ function isProduction() {
...
+  isStaging: () => boolean;
   isGitHubPages: () => boolean;
   showDrafts: () => boolean;
   buildingBlocksHomeUrl: string;
   feedbackFormUrl: string;
   signUpFormUrl: string;
 };
 
-type DocumentSiteEnvironment = 'development' | 'production';
+type DocumentSiteEnvironment = 'development' | 'uat' | 'staging' | 'production';
```
ianlnf, line 26: can define `staging` before `uat`

apps/docs/src/lib/config.ts
```diff
@@ -55,6 +79,17 @@ function getConfiguration(
   );
 }
 
+const getDeployEnvironment = () => {
```
ianlnf, line 82: would favour function definitions (`function ...`) over function expressions (`const .. = `)

ianlnf: could also type the return as `DocumentSiteEnvironment`

apps/docs/src/lib/config.ts
```diff
@@ -55,6 +79,17 @@ function getConfiguration(
...
+  if (process.env.DEPLOY_ENV === 'production') {
+    return 'production';
+  } else if (process.env.DEPLOY_ENV === 'staging') {
```
ianlnf: can remove redundant `else` throughout, or use `switch`

hamza14khan:
> I spoke with Tamara earlier today about which signup url are we supposed to use depending on the environment. She said we should use https://www.forms.uat.gov.ie/en/664b6de45f7c9800231daf22 for non-prod environments and https://www.forms.gov.ie/en/664ccbf2b644d000246cfd78 for the prod environment. Can we make these changes for the newly added environments ?

Yes correct, let me review it with her. Thanks for the headsup!

.github/workflows/ci.yml
```diff
@@ -7,6 +7,17 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
ianlnf, line 10: This will need to be in `deploy-docs-prod.yml`. The `ci.yml` deploys to `dev` already on merges to `main`.

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       id-token: write
       contents: read
 
-    environment: dev
+    environment: development
```
ianlnf, line 28: this will need to be `dev` as it refers to the GitHub environments (i.e. where to retrieve secrets)

.github/workflows/deploy-docs-prod.yml
```diff
@@ -3,10 +3,15 @@ name: Deploy docs to Production
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        description: "Git tag to deploy"
```
ianlnf, line 7: still need the existing inputs (which tag to deploy, which branch)

.github/workflows/deploy-docs-prod.yml
```diff
@@ -3,10 +3,15 @@ name: Deploy docs to Production
...
+      environment:
+        description: "Select deployment environment"
         required: true
-        default: "last-release"
+        type: choice
+        options:
+          - staging
+          - uat
+          - production
+        default: uat
```
ianlnf, line 14: default: staging

apps/docs/project.json
ianlnf: staging

apps/docs/src/lib/config.ts
```diff
@@ -42,6 +40,18 @@ function getConfiguration(
 
   const configurationOverrides: Configurations = {
     development: {},
+    staging: {
+      // TODO Review the following paths
```
ianlnf, line 44: comments can be removed

ianlnf (approved):

PR #166, opened 2024-08-15
docs: add dynamic title and description

BogdanDanilescu:


apps/docs/src/app/[...slug]/layout.tsx
```diff
@@ -15,6 +15,30 @@ type DocumentLayoutProps = {
   };
 };
 
+export async function generateMetadata({
+  params,
+}: {
+  params: { slug: string[] };
+}) {
+  const document = documents.getBySlug({ slug: params.slug });
```
Comment (ianlnf) [inline apps/docs/src/app/[...slug]/layout.tsx:23]: could add a `documents.getMeta({ slug })` and `documents.getTitle({ title })` or similar to centralise the suffix

Comment (BogdanDanilescu) [reply apps/docs/src/app/[...slug]/layout.tsx:23]: Good idea @ianlnf . I'll create the helpers. But I'm wondering if it wouldn't make more sense to add the suffix directly in the frontmatter since that's the place where we have the metadata content ?

Comment (ianlnf) [reply apps/docs/src/app/[...slug]/layout.tsx:23]: yes that's another idea for MDX files, if a calculated title field can be used, but the title helper would still be required to use in custom pages and the calculated field

ianlnf (approved):

PR #179, opened 2024-08-22
ci: run storybook pipeline on html package changes

hamza14khan:
- Verify if `govie-frontend/storybook` package was affected by the associated dependencies and make appropriate deployment.
- Also, enable manual deployment.


.github/workflows/deploy-html-storybook-prod.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "packages/html/**"
```
ianlnf, line 8: The issue with this approach is that any changes in the downstream dependencies of the html packages will not trigger a new Storybook deployment, e.g. tokens package, theme package etc. You'd be better to use the `nx affected` command and add an Nx deploy script for each storybook instance that programatically pushes to S3.

hamza14khan: Good catch! Let me fix this.

.github/workflows/deploy-html-storybook-prod.yml
```diff
@@ -4,26 +4,73 @@ on:
...
+
+      - name: Install dependencies
+        run: |
+          corepack enable
+          pnpm install --frozen-lockfile --ignore-scripts
+
+      - name: Determine affected projects
+        id: affected
+        run: |
+          npx nx affected:libs --base=origin/main~1 --head=HEAD > affected.txt
```
pregno, line 39: This command in local doesn't work, does `affected:libs` need to be present in package.json?

ianlnf (approved):

.github/workflows/deploy-html-storybook-prod.yml
```diff
@@ -4,26 +4,72 @@ on:
...
+
+      - name: Install dependencies
+        run: |
+          corepack enable
+          pnpm install --frozen-lockfile --ignore-scripts
+
+      - name: Determine affected projects
+        id: affected
+        run: |
+          nx show projects --affected --base=origin/main~1 --head=HEAD > affected.txt cat affected.txt
```
pregno, line 39: Is `cat affected` needed?
Shouldn't it be on a new line?

pregno (approved):

PR #181, opened 2024-08-23
test: add nunjuck validation test

ianlnf:


examples/wagtail/core/static/js/govie-frontend.umd.js
pregno, line 1: Can we have a symlink to the `dist` file?

PR #186, opened 2024-08-30
feat: enable tailwind in react components

pregno:
Adopt tailwind for React components


examples/vite/src/application.tsx
```diff
@@ -9,17 +13,49 @@ import {
...
+
+        <Link
+          href="https://www.google.com"
+          external={true}
+          noUnderline={true}
+          noVisited={true}
+        >
+          Link
+        </Link>
+
```
hamza14khan, line 42: non-blocking: Can we remove these extra lines/spacing? If its intentional let me know.

pregno: done

packages/react/ds/src/link/link.tsx
```diff
@@ -15,7 +14,6 @@ export function Link({
   noUnderline,
   noVisited,
   external,
-  dark = false,
```
BogdanDanilescu: if we don't use the `dark` property anymore I think we also need to `link.schema.ts`

BogdanDanilescu (approved):

hamza14khan (approved):
Awesome work!

PR #194, opened 2024-09-03
chore: add chromatic

pregno:


BogdanDanilescu:
I'm not sure how chromatic works as I'm not familiar with it but is there some way we can run it locally as well ? perhaps a script in `package.json` ?

BogdanDanilescu (approved):

hamza14khan (approved):

PR #196, opened 2024-09-04
ci: fix storybook deployment actions

pregno:


.github/workflows/deploy-html-storybook-prod.yml
```diff
@@ -82,8 +82,6 @@ jobs:
     name: Deploy to Amazon S3
     runs-on: ubuntu-latest
     needs: build
-    # Run if html is affected or manually triggered
-    if: ${{ needs.determine-affected.outputs.html_affected == 'true' || github.event_name == 'workflow_dispatch' }}
```
hamza14khan, line 86: Dont we need this here for manual deployment? Happy to fix it for you.

.github/workflows/deploy-react-storybook-prod.yml
```diff
@@ -9,8 +9,46 @@ env:
...
+            echo "react_affected=true" >> $GITHUB_OUTPUT
+          else
+            echo "react_affected=false" >> $GITHUB_OUTPUT
+          fi
+
   build:
     runs-on: ubuntu-latest
+    needs: determine-affected
+    # Run if html is affected or manually triggered
+    if: ${{ needs.determine-affected.outputs.react_affected == 'true' || github.event_name == 'workflow_dispatch' }}
```
hamza14khan, line 51: Given that this check is here, we might need manual trigger here :D

.github/workflows/deploy-html-storybook-prod.yml
pregno: It depends on `build`, which has the same reference.
Plus is doesn't reference `determine-affected` so it wouldn't work anyway.

.github/workflows/deploy-react-storybook-prod.yml
pregno: thanks, forgot about it

hamza14khan (approved):
Great stuff 🥇

PR #198, opened 2024-09-04
chore: update paragraph classes and tests

hamza14khan:
- Global html paragraph class update and associated tests.
- Classes now synced with react paragraph component.

### DIFF
--- a/packages/html/ds/src/paragraph/paragraph.html
+++ b/packages/html/ds/src/paragraph/paragraph.html
@@ -1,19 +1,19 @@
 {% macro govieParagraph(props) %}
   {% macro getSizeClass(size) %}
     {% if size == 'lg' %}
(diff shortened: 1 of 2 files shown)


packages/html/ds/src/paragraph/paragraph.html
```diff
@@ -1,19 +1,19 @@
 {% macro govieParagraph(props) %}
   {% macro getSizeClass(size) %}
     {% if size == 'lg' %}
-      xs:gi-text-lg md:gi-text-lg xl:gi-text-xl
+      gi-text-md xs:gi-text-md md:gi-text-md xl:gi-text-md
```
pregno, line 4: You have the correct sizes already; we must update them on React.

pregno (changes requested):

PR #204, opened 2024-09-06
docs: update docs to react library

pregno:


apps/docs/src/app/page.tsx
```diff
@@ -51,24 +50,22 @@ function Benefit({
...
 }
 
 function Benefits() {
   return (
     <div className="flex flex-col gap-4xl items-center">
-      <Text
+      <Heading
         as="h2"
-        className="text-center font-semibold lg:px-6xl max-w-[50ch]"
+        // className="text-center font-semibold lg:px-6xl max-w-[50ch]"
```
hamza14khan, line 64: suggestion: Can we remove this?

packages/react/ds/src/text/text.tsx
```diff
@@ -11,7 +11,7 @@ function getTextClass({ as, size }: { as: TextAs; size?: TextSize }) {
     const sizes = {
       sm: 'xs:gi-text-sm md:gi-text-sm xl:gi-text-md',
       md: 'xs:gi-text-md md:gi-text-md xl:gi-text-lg',
-      lg: 'xs:gi-text-lg md:gi-text-lg xl:gi-text-xl',
+      lg: 'gi-text-lg',
```
hamza14khan, line 14: question: Should html components be updated as well?

hamza14khan (approved):
Looks good. Just few questions.

packages/react/ds/src/text/text.tsx
pregno: yes, you're right. I'm doing it

PR #205, opened 2024-09-06
ci: update actions version

pregno:


BogdanDanilescu (approved):

.github/workflows/deploy-react-storybook-prod.yml
```diff
@@ -84,11 +85,8 @@ jobs:
       contents: read
     environment: prod
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
       - name: Download artifact
-        uses: actions/download-artifact@v3.0.2
+        uses: actions/download-artifact@v4
```
hamza14khan, line 89: question: Should we do similar changes to `deploy-docs-gh-pages.yml`?

pregno: `deploy-docs-gh-pages.yml` look ok to me

PR #209, opened 2024-09-12
feat: use icon font for material design

pregno:


apps/docs/src/components/document/components/component-status-table.tsx
```diff
@@ -33,7 +33,7 @@ export function ComponentStatusPill({
       </div>
       {href ? (
         <IconButton
-          icon={<Icon id="open-in-new" size="sm" />}
+          icon={<Icon icon="open_in_new" size="sm" inline={true} />}
```
hamza14khan, line 36: Will just `inline` work?

pregno: I put it because the icon works with a `<span>` element while previously was a `<div>`.
Probably, I can still use the `<div>`, but in this way, we can put the icon inline with text.

pregno: > Will just inline work?
Sorry I just get your question 😬, I have to try

pregno: Yes, it works.
But in the button it works even without the inline

BogdanDanilescu (approved):

PR #210, opened 2024-09-12
feat: react header component

BogdanDanilescu:
### **Header implementation for React library**

Key aspects:

- Properties added ( nav links, secondary nav links )
- 3 Different views depending on the screen size ( desktop, tablet, mobile )
- Works as a Server Component ( Compatible in case of no JS )

Screenshots
_**Desktop view:**_
image
**_Tablet view_**
image
**_Mobile view_**
image


packages/react/ds/styles.css
```diff
@@ -10,3 +10,17 @@
   max-height: 0;
   opacity: 0;
 }
+
+/* Header Styles */
+
+#MenuContainer:has(#SearchTrigger:checked) ~ #SearchContainer {
```
pregno, line 16: I'm not sure about this.
You should create a class used only for the container instead of using the IDs.
Check this guide
https://tailwindcss.com/docs/adding-custom-styles#adding-component-classes

Or is there something that I'm missing?

pregno (changes requested):

packages/react/ds/assets/icons/closeIcon.tsx
hamza14khan, line 1: suggestion (non-blocking): I thought we were keeping the following format for the components?

```export const CloseIcon = () => { ... }```

File: packages/react/ds/src/header/components/headerMenu.tsx [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+          className="gi-mr-8 gi-text-md gi-flex gi-items-center gi-gap-2 gi-cursor-pointer"
+        >
+          <Text size="sm" as="span">
+            Close
+          </Text>
+          <CloseIcon />
+        </label>
+      </div>
+
+      <ul className="gi-bg-white gi-px-4 sm:gi-px-8 gi-h-screen">
```
hamza14khan, line 34: Not sure if its worth to add a TODO here to replace `ul` with the future `List` component? 

@pregno

packages/react/ds/src/header/components/headerMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+            >
+              <Text className="gi-font-bold" size="sm" as="span">
+                {link.label}
+              </Text>
+            </a>
+          </li>
+        ))}
+        {languages?.map((link, i) => (
+          <li key={i}>
+            <a
```
hamza14khan, line 49: question (non-blocking): why aren't we using `Link` component here?

packages/react/ds/src/header/header.tsx
```diff
@@ -1,76 +1,105 @@
...
-                  focus:gi-shadow-[0_-2px_var(--gieds-color-yellow-400),0_4px_var(--gieds-color-gray-900)]  
-                `}
-              >
-                <GovIrelandLogo />
-              </a>
-            ) : (
-              <GovIrelandLogo />
-            )}
+    <header
+      className={`gi-relative gi-overflow-x-clip gi-bg-emerald-800 gi-border-gold-500 gi-font-primary`}
```
hamza14khan, line 24: nitpick: We can skip `{`, `}` braces here.

hamza14khan (changes requested):
Great work on the header component!
Just few comments from me and Andrea.

packages/react/ds/src/header/components/headerMenu.tsx
pregno: We will not use the `<List>` component here.

packages/react/ds/styles.css
BogdanDanilescu: what I'm trying to do here ( styling an element based on another element's state )  can be achieved through the `peer` class 
https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-sibling-state

I've tried that approach but for some reason the peer class does not work as intended.

packages/react/ds/assets/icons/closeIcon.tsx
BogdanDanilescu: this is fixed now as I removed the CloseIcon because we are using the `<Icon />` component now

packages/react/ds/src/header/components/headerMenu.tsx
BogdanDanilescu: It's a similar approach with the links in the Footer. Because they don't have a look and feel as a normal link, I think it's better to go with this approach when it comes to components like `<Header />` or `<Footer />`

packages/react/ds/src/header/header.tsx
```diff
@@ -1,76 +1,105 @@
...
+          )}
+          <label
+            htmlFor="SearchTrigger"
+            className="focus-visible:gi-outline-offset-0 focus-visible:gi-outline-none focus-visible:gi-border focus-visible:gi-border-solid focus-visible:gi-border-yellow-400 gi-hidden xs:gi-flex gi-rounded-sm hover:gi-bg-black hover:gi-bg-opacity-20 gi-p-2 gi-items-center gi-gap-md gi-cursor-pointer focus:gi-border focus:gi-border-solid focus:gi-border-yellow-400"
+          >
+            <input className="gi-hidden" id="SearchTrigger" type="checkbox" />
+            <Text className="gi-text-2md gi-font-bold gi-text-white" as="span">
+              Search
+            </Text>
+            <Icon className="search-icon gi-text-white" icon="search" />
```
pregno, line 78: Where is defined this class `search-icon`?

packages/react/ds/src/header/header.tsx
```diff
@@ -1,76 +1,105 @@
...
+            <Icon className="gi-hidden close-icon gi-text-white" icon="close" />
```
pregno, line 79: You can put the `Icon` inside the `Text`. It will inherit the text colour.

packages/react/ds/src/header/header.tsx
```diff
@@ -1,76 +1,105 @@
...
+          >
+            <input
+              className="gi-hidden"
+              id="MobileMenuTrigger"
+              type="checkbox"
+            />
+            <Text className="gi-text-2md gi-font-bold gi-text-white" as="span">
+              Menu
+            </Text>
+            <Icon className="gi-text-white" icon="menu" />
```
pregno, line 93: Same consideration as above.

packages/react/ds/src/icon/icon.tsx
```diff
@@ -24,6 +27,7 @@ export function Icon({
   ariaHidden,
   ariaLabel,
   inline = false,
+  className = '',
```
pregno, line 30: I wonder if we need this. We probably don't need it if it's only required to change the colour.

pregno (commented):
Added a couple of comments

packages/react/ds/src/header/header.tsx
BogdanDanilescu: it's in the `styles.css` as a workaround for the `peer` modifier from tailwind since `peer` did not work as expected. if we can resolve the `peer` issue then we need to remove the `className` prop from the `Icon` component

packages/react/ds/src/header/header.tsx
BogdanDanilescu: yes, we could but we will have the same as above, since the `search-icon` class is used as workaround for the `peer`

packages/react/ds/src/header/header.tsx
BogdanDanilescu: I also had to replace the `Text` component with a regular span because we needed the font-size to stay the same accross all screens and the `Text` component doesn't give us that capabilty

packages/react/ds/src/icon/icon.tsx
BogdanDanilescu: we can get around it in this instance, maybe for other components we might need to change the colour but I can guess we can make the changes when/if we come across the problem

pregno (approved):

hamza14khan (approved):

PR #212, opened 2024-09-13
feat: add text input along with label, hint-text, error-text and form-group components.

hamza14khan:
Addition of the following components to support `TextInput`.

- `Label`.
- `HintText`.
- `ErrorText`.
- `FormGroup`

Testing:
Screenshot 2024-09-13 at 16 56 05
Screenshot 2024-09-13 at 16 56 16
Screenshot 2024-09-13 at 16 56 23


packages/react/ds/src/text-input/text-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+  ) => {
+    // Determine dynamic width class
+    const widthClass =
+      fullFluid && !halfFluid && !characterWidth
+        ? 'gi-w-full'
+        : halfFluid
+          ? 'gi-w-1/2'
+          : characterWidth
+            ? // TODO Not working. Review it with the team.
+              `gi-w-[${characterWidth}em]`
```
hamza14khan, line 37: @pregno @BogdanDanilescu 

Can you please help me with this? E.g. if `characterWidth = 4` it will result in `gi-w-[4em]` to the `div` but somehow this not reflecting.

packages/react/storybook/.storybook/main.ts
```diff
@@ -14,6 +14,22 @@ const config: StorybookConfig = {
     name: '@storybook/react-vite',
     options: {},
   },
+  async viteFinal(config, { configType }) {
```
hamza14khan, line 17: Do we need this now?

@BogdanDanilescu

packages/react/ds/tailwind.config.ts
```diff
@@ -3,7 +3,7 @@ import { Config } from 'tailwindcss';
 
 const config: Config = {
   prefix: 'gi-',
-  content: ['./src/**/*.tsx'],
+  content: ['../ds/src/**/*.tsx'],
```
hamza14khan, line 6: This as well?

@BogdanDanilescu

packages/react/ds/src/text-input/text-input.tsx
pregno: That's most probably because the classes are not generated at build time see here https://tailwindcss.com/docs/content-configuration#classes-arent-generated

Anyway, why are you using `[4em]` and not `w-16`

packages/react/storybook/.storybook/main.ts
BogdanDanilescu: I would say let's keep this because this is what enables hot reloading, so it saves a lot of time for us

packages/react/ds/tailwind.config.ts
BogdanDanilescu: no ,we can remove this change. this was for the tailwing config in the storybook folder. it tells tailwind where to look for classes in order to know which classes should be built at runtime

packages/react/ds/src/form-group/form-group.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+export type FormGroupProps = React.HTMLAttributes<HTMLDivElement> & {
+  hasError?: boolean;
+  children: React.ReactNode;
+};
+
+export const FormGroup = React.forwardRef<HTMLDivElement, FormGroupProps>(
+  ({ hasError = false, children, ...props }, ref) => {
```
BogdanDanilescu, line 9: do we need a default value here ? the `hasError` is optional and so it will be falsy when we don't declare it

packages/react/ds/src/text-input/text-input.tsx
BogdanDanilescu: I think it's because the class is not being generated since you are using `gi-w-[${characterWidth}em]` and it doesn't know which class to generate. One way to get around this is to safelist the classes. 

https://tailwindcss.com/docs/content-configuration#safelisting-classes

This ensures that whatever classes we add to the safelist object, it will be available in the application at runtime

packages/react/ds/src/form-group/form-group.tsx
hamza14khan: Good catch!

BogdanDanilescu (approved):

pregno (approved):

PR #222, opened 2024-09-18
feat: add html text input component

hamza14khan:
Addition of the following html components that was associated with `text-input`.

- `HintText`.
- `ErrorText`.
-  `Label`.

Note: I have added tests for `text-input` component but they are failing to due to imports of the above additional components. Let me know if I should remove them or debug it.

Testing:
Screenshot 2024-09-18 at 19 29 00


packages/html/storybook/.storybook/preview.tsx
```diff
@@ -40,7 +40,6 @@ export const decorators = [
 
 const preview: Preview = {
   parameters: {
-    layout: 'fullscreen',
```
hamza14khan, line 43: cc @BogdanDanilescu

packages/html/ds/src/text-input/text-input.test.ts
hamza14khan, line 1: help-wanted: I think we have discussed this earlier, but do you guys know if there is a fix for imports error? E.g. I am importing label, error-text and hint text inside `text-input.html` so that can be the reason the `render` function is throwing errors on executing tests.

cc @BogdanDanilescu @pregno

pregno (approved):

BogdanDanilescu (approved):
Great work on the component

PR #227, opened 2024-09-20
feat: html checkbox component

BogdanDanilescu:
### Global HTML checkbox component

`fieldId`
`items`
`errorMessage`
`noneOption`
`title`
`size`

[Link for SB](https://66d6cf734f628d9cfe0244c1-mmjhbubgac.chromatic.com/?path=/docs/form-checkboxes--docs)

**_SB Screenshots_**
image
image
image
image


packages/html/ds/src/checkbox/checkbox.html
```diff
@@ -0,0 +1,22 @@
...
+
+<div class="gi-flex gi-flex-col">
+  <div class="gi-flex gi-items-center">
+    <input data-element="{{ props.dataElement }}" id="{{ props.checkboxId }}" value="{{ props.value }}" class="{{ getSizeClass(props.size) | trim }} {{ getTickSize(props.size) | trim }} gi-cursor-pointer gi-appearance-none gi-border-[3px] gi-border-solid gi-border-black checked:before:gi-block checked:before:gi--rotate-45 checked:before:gi-relative checked:before:gi-border-l-[4px] checked:before:gi-border-b-[4px] checked:before:gi-border-black" type="checkbox" />
+    <label for="{{ props.checkboxId }}" class="gi-cursor-pointer gi-font-primary xs:gi-text-sm md:gi-text-md lg:gi-text-lg gi-pl-3.5 ">
+      {{ props.label }}
+    </label>
+  </div>
+  {% if props.hint %}
+    <div class="gi-mb-[-10px]">
```
BogdanDanilescu, line 14: @pregno @hamza14khan 
A small suggestion here. I've had to add the `-10px` value because the `hint-text` has a margin bottom of `10px` already integrated in the component. I think a better way to approach this is to keep the 'hint-text` and `error-text` components without margins since that will be applied from the parent component

packages/html/ds/src/checkbox/checkboxes-group.html
```diff
@@ -0,0 +1,63 @@
...
+          <div data-testid="title-hint" class="gi-mb-[-10px]">
+            {{ govieHintText({
+              "content": props.title.hint
+            })}}
+          </div>
+        {% endif %}
+      </legend>
+      <div data-element="checkbox-container" class="gi-flex gi-flex-col gi-gap-2.5">
+        {% if props.errorMessage %}
+          <div class="gi-mb-[-14px]">
```
BogdanDanilescu, line 33: Same suggestion here

packages/html/ds/src/checkbox/checkbox.html
pregno: Or having a props `no-margin` to disable them

packages/html/ds/src/checkbox/checkboxes-group.html
```diff
@@ -0,0 +1,63 @@
...
+  <div data-testid="govie-checkboxes" data-module="gieds-checkboxes" class="gi-flex">
+    {% if props.errorMessage %}
+    <div class="gi-w-5 gi-border-l-[5px] gi-border-l-red-600 gi-border-solid"></div>
+    {% endif %}
+    <fieldset>
+      <legend class="gi-mb-3.5 gi-font-primary sm:gi-text-sm md:gi-text-md lg:gi-text-lg">
+        {% if props.title.asHeading %}
+          {{ govieHeading({
+            "size": "lg",
+            "tag": "h1",
```
pregno, line 17: We should probably also pass `size` and `tag` as options for the heading.

BogdanDanilescu: good idea, I'll do that. This is what ogcio storybook has but yes, it's better to give users this flexibility

BogdanDanilescu: Fixed

packages/html/ds/src/checkbox/checkbox.html
BogdanDanilescu: @hamza14khan is it okay with you if I add the `no-margin` prop for the `hint-text` and `error-text` components

BogdanDanilescu: Fixed

packages/html/ds/src/checkbox/checkboxes-group.html
BogdanDanilescu: Fixed

packages/html/ds/src/error-text/error-text.html
```diff
@@ -8,7 +8,14 @@
       xs:gi-text-sm md:gi-text-md
     {% endif %}
   {% endmacro %}
-  <p class="{{ getSizeClass(props.size) | trim }} gi-font-bold gi-leading-5 gi-text-red-600 gi-clear-both gi-block !gi-mb-[14px] !gi-mt-0">
+
+  {% macro noMarginMacro(noMargin) %}
```
hamza14khan, line 12: Instead of having this particular prop, how about overriding or extending the classes?

@pregno @BogdanDanilescu

BogdanDanilescu: we could do that as well, but that will mean to give the user the ability to completely override our styles. not sure if we want to do that

hamza14khan: We can have class names as props but not expose it on storybook.

BogdanDanilescu: Sounds good to me . @pregno ?

pregno (approved):

packages/html/ds/src/error-text/error-text.html
BogdanDanilescu: fixed. I added the option to add additional classes if needed

PR #228, opened 2024-09-20
feat: add tabs for react

pregno:


packages/react/ds/src/tabs/tab-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <>
+      <input
+        name="tabs"
+        type="radio"
+        id={`tab-${valueSlug}`}
+        aria-labelledby={`tab-label-${valueSlug}`}
+        aria-roledescription="tab"
+        defaultChecked={checked}
+        className={`gi-absolute gi-opacity-0`}
```
hamza14khan, line 28: minor: We can just go without template strings here.

packages/react/ds/src/tabs/tab-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+          gi-relative
+          gi--mb-[1px]
+          gi-mr-1
+          gi-bg-gray-50
+          gi-text-center 
+          gi-text-md
+          gi-underline
+          gi-text-slate-300`}
+      >
+        <Link href={href} noVisited={true} noColor={true}>
```
hamza14khan, line 52: minor: We can go without `={true}` here. E.g. usage of just `noVisited` and `noColor`.

packages/react/ds/src/tabs/tab-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+        className={`
+          gi-inline-block 
+          gi-cursor-pointer
+          gi-px-5
+          gi-py-2
+          gi-border-solid 
+          gi-border-border-gray-200
+          gi-border-b-xs 
+          gi-relative
+          gi--mb-[1px]
```
hamza14khan, line 44: question: Is there a typo here? Shouldn't it be `gi-mb-[1px]`

pregno: nope, is negative

packages/react/ds/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,22 @@
...
+  ariaLabel?: string;
+  ariaLabelledby?: string;
+  children: React.ReactNode;
+}) {
+  return (
+    <div
+      role="tablist"
+      aria-label={ariaLabel}
+      aria-labelledby={ariaLabelledby}
+      className={`gi-flex gi-flex-wrap`}
```
hamza14khan, line 17: Same as above.

packages/react/ds/src/utils.ts
```diff
@@ -0,0 +1,13 @@
+export function slugify(value?: string) {
```
hamza14khan, line 1: Sorry I didn't catch why would we use slugify? 😄
Can we add few bits for this file as a documentation.

packages/react/ds/src/tabs/tab-item.tsx
hamza14khan: Good one! Thanks for information.

packages/react/ds/src/utils.ts
pregno: It's used because the `value` is used in the `id` field of the tab.
For avoid spacing or chars not valid in `id` field I opted for using that slug function.

hamza14khan: Thanks! Can we please add documentation for this file maybe with comments on top?

packages/react/ds/src/tabs/tab-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Link } from '../link/link.js';
+
+export function TabItem({
+  value,
+  href,
+  checked = false,
+  children,
+}: {
+  value: string;
+  href: string;
```
Thurfjell, line 12: Intended to be a required prop? 

Should the link become "link" colour? 
image

Does anybody possibly need to know what is active based on a click, or is this supposed to be a readonly component (makes sense). 

There's a behaviour I'm not convinced on, if you provide a href, you can click slightly off the link text and it will only change tab, not hit the anchor.

packages/react/ds/styles.css
```diff
@@ -17,29 +17,61 @@
...
+      display: block; 
+    }
+    
+    #MenuContainer:has(#MobileMenuTrigger:checked) ~ #HeaderMenuContainer {
+      transform: translateX(0);
+    }
+    #MenuContainer:has(#MobileMenuTrigger:checked) ~ #HeaderOverlayContainer {
+      display: block;
+      position: fixed;
+    }
```
Thurfjell, line 44: Sneak feature? 😂

packages/react/ds/styles.css
```diff
@@ -17,29 +17,61 @@
...
+  [name="tabs"]:nth-of-type(3):checked~[role="tabpanel"]:nth-of-type(3),
+  [name="tabs"]:nth-of-type(4):checked~[role="tabpanel"]:nth-of-type(4),
+  [name="tabs"]:nth-of-type(5):checked~[role="tabpanel"]:nth-of-type(5),
+  [name="tabs"]:nth-of-type(6):checked~[role="tabpanel"]:nth-of-type(6),
+  [name="tabs"]:nth-of-type(7):checked~[role="tabpanel"]:nth-of-type(7),
+  [name="tabs"]:nth-of-type(8):checked~[role="tabpanel"]:nth-of-type(8),
+  [name="tabs"]:nth-of-type(9):checked~[role="tabpanel"]:nth-of-type(9),
+  [name="tabs"]:nth-of-type(10):checked~[role="tabpanel"]:nth-of-type(10),
+  [name="tabs"]:nth-of-type(11):checked~[role="tabpanel"]:nth-of-type(11),
+  [name="tabs"]:nth-of-type(12):checked~[role="tabpanel"]:nth-of-type(12) {
```
Thurfjell, line 72: I guess there should be some questions asked if there's a tab field of 12 tabs? 😓

packages/react/ds/styles.css
pregno: not mine this one :D

BogdanDanilescu: That's my doing 😄 are we saying we should refactor it ?

packages/react/ds/styles.css
pregno: Yes, I still don't know how to solve this.
I will try to find a better solution, at least we know that this is working

packages/react/ds/styles.css
Thurfjell: Nothing to see here.

packages/react/ds/src/tabs/tab-item.tsx
```diff
@@ -49,9 +48,20 @@ export function TabItem({
           gi-underline
           gi-text-slate-300"
       >
-        <Link href={href} noVisited noColor noUnderline>
+        <a
```
BogdanDanilescu, line 51: I think this should be a `<button>` rather than a `<a>` . I don't see a use case where the menu item will be a link

Thurfjell: Use case whenever you need to use url as state.

Thurfjell (approved):

packages/react/ds/src/tabs/tab-item.tsx
pregno: As @Thurfjell wrote, we can use the anchor to change the page state. 
If you don't use the `href` the `<a>` is acting as a `<span>` but with focus.

BogdanDanilescu: Makes sense

BogdanDanilescu (approved):

packages/react/ds/src/tabs/tab-panel.tsx
```diff
@@ -0,0 +1,30 @@
...
+  ariaLabel?: string;
+  ariaLabelledby?: string;
+  children: React.ReactNode;
+}) {
+  const valueSlug = slugify(value);
+  return (
+    <div
+      role="tabpanel"
+      aria-labelledby={`tab-${valueSlug}`}
+      className={`
```
hamza14khan, line 18: We can just go without template strings here 👍

packages/react/ds/src/utils.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
hamza14khan, line 1: praise: Thanks for adding the documentation!

hamza14khan (approved):
Looks good, one minor comment.

PR #231, opened 2024-09-23
feat: add global html text-area component

hamza14khan:
Addition of global html text-area component.

Testing on Storybook:

Screenshot 2024-09-23 at 11 46 51


pregno (approved):

packages/html/ds/src/textarea/textarea.schema.ts
```diff
@@ -0,0 +1,33 @@
+import * as zod from 'zod';
+
+export const textAreaSchema = zod.object({
```
BogdanDanilescu, line 3: Not sure I understand the schema. I can't see the `hintText`, `errorText` or `label` prop. Are we purposely leaving them out ?

packages/html/ds/src/textarea/textarea.schema.ts
```diff
@@ -0,0 +1,33 @@
...
+  hasError: zod
```
BogdanDanilescu, line 4: I think the `hasError` can be transformed in a `withError` that accepts a `string` . That way you can integrate both the label of the error and the boolean value of it

pregno: I agree

pregno (approved):

PR #233, opened 2024-09-23
feat: implement react checkbox component

BogdanDanilescu:
### Global HTML checkbox component

`fieldId`
`items`
`errorMessage`
`title`
`size`

**_SB Screenshots_**
image
image
image


packages/react/ds/src/checkbox/checkboxes-group.tsx
hamza14khan, line 1: I think we should keep each component independent of other components. Can you please check `text-input.tsx`? It is only relying on a simple `react` import.

So, the end results could be like this:

```<FormGroup hasError>
  <Label htmlFor="text-input">
    Label
  </Label>
  <ErrorText>
    Error
  </ErrorText>
  <CheckboxGroup
    hasError
    id="text-input"
  />
</FormGroup>```

Just FYI, there is a `ErrorText` and `FormGroup` component already made. Let me know what you think.

cc @pregno

Comment (BogdanDanilescu) [reply packages/react/ds/src/checkbox/checkboxes-group.tsx:1]: I'm not sure I fully understand. Are you referring to the `CheckboxesGroup` and `Checkbox` that we should keep them independent from each other ?  Even for the use case of using just one checkbox, we would still use `CheckboxesGroup` . The `CheckboxesGroup` acts as a wrapper because for example the hint prop is used at the wrapper level and also at checkbox level

Comment (BogdanDanilescu) [reply packages/react/ds/src/checkbox/checkboxes-group.tsx:1]: Maybe there is something that I'm missing here

**Review by pregno — APPROVED** [review]

File: packages/react/ds/src/checkbox/checkbox.tsx [hunk 1]
```diff
@@ -0,0 +1,49 @@
+import { HintText } from '../hint-text/hint-text.js';
```
hamza14khan, line 1: Sorry! I mean it here actually, if its possible to use HintText in CheckboxGroup only?

packages/react/ds/src/checkbox/checkboxes-group.tsx
hamza14khan: Sorry! Added my comment on `checkbox.tsx`.

packages/react/ds/src/checkbox/checkboxes-group.tsx
```diff
@@ -0,0 +1,68 @@
...
+  items,
+  size,
+}: CheckboxesGroupType) => {
+  return (
+    <div className="gi-flex">
+      {errorMessage && (
+        <div className="gi-w-5 gi-border-l-[5px] gi-border-l-red-600 gi-border-solid"></div>
+      )}
+      <fieldset>
+        <legend className="gi-mb-3.5 gi-font-primary sm:gi-text-sm md:gi-text-md lg:gi-text-lg">
```
hamza14khan, line 38: Question: I was reviewing the screenshot you included in the description, and I'm not sure if this will be addressed in the design review, but in the local HTML Storybook, there's noticeably less margin at the bottom after the legend.

Local HTML Checkbox:
Screenshot 2024-09-24 at 10 11 08

packages/react/ds/src/checkbox/checkbox.tsx
BogdanDanilescu: we are also using it in `CheckboxGroup ` but that applies to the hint of the group. The hint within the `Checkbox` component applies just to that specific checkbox

packages/react/ds/src/checkbox/checkboxes-group.tsx
BogdanDanilescu: You mean the margin-bottom of the heading ? That comes from the `Heading` component. I think we need to review the `Heading` component to fix this

packages/react/ds/src/checkbox/checkbox.tsx
hamza14khan: Yes, then we can just use it in the `CheckboxGroup` like this:

```
          {items.map((checkbox, index) => (
            <React.Fragment>
            <Checkbox
              dataElement={`checkbox${index}`}
              size={size}
              checkboxId={`${fieldId}-${index}`}
              //hint={checkbox.hint}
              value={checkbox.value}
              label={checkbox.label}
            />
            <HintText className='gi-mb-0'>{checkbox.hint}</HintText>
            </React.Fragment>
          ))}
```

packages/react/ds/src/checkbox/checkboxes-group.tsx
hamza14khan: Ok, we can then fix it in a separate PR, but lets file it somewhere on Azure 👍

BogdanDanilescu: agreed

packages/react/ds/src/checkbox/checkbox.tsx
BogdanDanilescu: Ah yes, I got what you mean now. Yes, it's possible to do that as well, I don't what benefit would that bring us. Also the parent of the hint component has some styles. If we move that outside the `Checkbox` component the styles of the parent will not apply anymore

BogdanDanilescu: In light of the conversation we had, can I keep the `hint` prop ?

hamza14khan: Yes, looks good!

hamza14khan (approved):
One last thing would be to update the doc site. Looks good other than that 👍

PR #234, opened 2024-09-24
fix: use svg for footer logo

pregno:


BogdanDanilescu:
@pregno I added this fix in the PR I created, I also added a `height` and `width` prop to the `Logo` macro so we can control the measurments

PR #236, opened 2024-09-24
feat: add html select component

BogdanDanilescu:
### Implementation of HTML Select component

Props:
`id`
`label`
`options`
`hint`
`error`

**_Screenshots_**
image
image
image


packages/html/ds/src/select/select.schema.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const selectSchema = zod.object({
+  id: zod.string({
+    description: 'A unique ID for the select component',
+    required_error: 'A value is required',
+  }),
+  label: labelSchema.describe('Properties for the label component').optional(),
+  options: zod
+    .object({
+      default: zod
```
pregno, line 14: Can `default` be optional?

BogdanDanilescu: Fixed. The default is now the first item in the options array

packages/html/ds/src/select/select.schema.ts
```diff
@@ -0,0 +1,31 @@
...
+      value: zod.string({
+        description: 'Value for the option',
+        required_error: 'The value for the option is required',
+      }),
+    })
+    .array()
+    .describe('Array of options with labels and values'),
+  hint: hintTextSchema.describe('Properties for the hint component').optional(),
+  error: errorTextSchema
+    .describe('Properties for the erro component')
```
hamza14khan, line 27: Typo: `error` instead of `erro`

packages/html/ds/src/select/select.schema.ts
```diff
@@ -0,0 +1,31 @@
...
+        required_error: 'The label for the option is required',
+      }),
+      value: zod.string({
+        description: 'Value for the option',
+        required_error: 'The value for the option is required',
+      }),
+    })
+    .array()
+    .describe('Array of options with labels and values'),
+  hint: hintTextSchema.describe('Properties for the hint component').optional(),
```
hamza14khan, line 25: question: we already have generic descriptions for label, error and hint. Why don't just use them?

packages/html/ds/src/select/select.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+        category: 'Label',
+        type: { summary: 'Label' },
+      },
+    },
+  },
+  args: {
+    id: 'Unique-ID',
+    label: {
+      content: 'Label',
+    },
```
hamza14khan, line 72: Can you please also add default values for error and hint?

```
errorText: {
      content: '',
    },
    hintText: {
      content: '',
    },
 ```
 This way user will not have to make JSON schema manually.

File: packages/html/ds/src/select/select.stories.ts [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+        label: 'Option 3',
+        value: 'value-3',
+      },
+    ],
+  },
+};
+
+export const withoutLabel: Story = {
+  args: {
+    id: 'Unique-ID',
```
hamza14khan, line 144: question: Is it possible to use lowercase ids? E.g. `unique-id`.

hamza14khan (changes requested):
Looks good. Added few comments to sync with other form component conventions.

Also, you might wanna update:

- Doc site.
- Wagtail example.
- Vite example.

packages/html/ds/src/select/select.stories.ts
BogdanDanilescu: good idea. fixed

packages/html/ds/src/select/select.stories.ts
BogdanDanilescu: fixed

packages/html/ds/src/select/select.schema.ts
BogdanDanilescu: not sure I understand. You're saying I should omit the `describe` from line 25 ?

hamza14khan: I mean to say just use it as like this without describe. Because it already have a description?

```
  label: labelSchema.optional(),
  hint: hintTextSchema.optional(),
  error: errorTextSchema.optional(),
```

BogdanDanilescu: ah yes, I got what you mean now. That was my first instinct too but I noticed we are getting an error upon build if I don't add the description

pregno (approved):

hamza14khan (approved):

PR #237, opened 2024-09-24
chore: update `TextArea` and `TextInput` with label, error and hint props

hamza14khan:
Vite testing:
Screenshot 2024-09-24 at 17 18 41

Storybook props for textarea:
Screenshot 2024-09-24 at 17 28 01

Storybook props for textinput:
Screenshot 2024-09-24 at 17 15 13


packages/html/ds/src/text-input/text-input.schema.ts
```diff
@@ -42,6 +39,9 @@ export const textInputSchema = zod.object({
       description: 'Sets the unique ID for the input field.',
     })
     .optional(),
+  label: labelSchema.optional(),
+  hintText: hintTextSchema.optional(),
+  errorText: errorTextSchema.optional(),
 });
```
BogdanDanilescu, line 45: lines 42,43 and 44 will give an error because there is no description provided

BogdanDanilescu: I think the `hintText` and `errorText` should be renamed to `hint` and `error` because they are objects that contain not just the text of the components but also `size` and `classNames`

packages/react/ds/src/text-input/text-input.tsx
```diff
@@ -1,33 +1,40 @@
...
-  hasError?: boolean;
   prefix?: React.ReactNode;
   suffix?: React.ReactNode;
   halfFluid?: boolean;
   fullFluid?: boolean;
   characterWidth?: number;
+  label?: LabelProps;
+  hintText?: HintTextProps;
+  errorText?: ErrorTextProps;
 };
```
BogdanDanilescu, line 18: same here.  the hint and error components contain more props than just the text

packages/react/ds/src/textarea/textarea.tsx
```diff
@@ -1,38 +1,65 @@
...
     {
-      hasError,
       rows = 4, // default row count
       cols = 100, // default column count
       autoComplete = 'on', // default autoComplete behavior
+      label,
+      errorText,
+      hintText,
+      id,
       ...props
```
BogdanDanilescu, line 30: same here

packages/react/ds/src/textarea/textarea.tsx
```diff
@@ -1,38 +1,65 @@
...
     },
     ref,
   ) => {
+    const hasError = errorText && errorText.content;
```
BogdanDanilescu, line 34: I think there's something I'm missing here. You're checking to see if the `content` property exists but in the `ErrorText` component we don't use a `content` prop. I checked to see if it works and it does but not sure where it is picking up the value from

packages/html/ds/src/text-input/text-input.schema.ts
hamza14khan: Good catch!

packages/react/ds/src/text-input/text-input.tsx
hamza14khan: On it!

packages/react/ds/src/textarea/textarea.tsx
hamza14khan: Good catch! Fixed now.

BogdanDanilescu (approved):
LGTM, great stuff on the components 👍

examples/vite/src/application.tsx
```diff
@@ -44,12 +42,34 @@ export function App() {
...
-        <FormGroup>
-          <Label htmlFor="text-input">4 characters width</Label>
-          <HintText>Hint Text</HintText>
-          <TextInput characterWidth={40} id="text-input" />
-        </FormGroup>
         <PhaseBanner level="alpha">This is a pre-release version</PhaseBanner>
+        <TextInput
+          id="text-input-id"
+          error={{
+            content: "Error: Please correct this issue.",
```
pregno, line 49: Is `content` supposed to be `children`?

pregno (approved):

PR #238, opened 2024-09-25
feat: add tabs to global html

pregno:
Normal state
Screenshot 2024-09-26 at 17 27 33

Focused state
Screenshot 2024-09-26 at 17 27 41

Hover state
Screenshot 2024-09-26 at 17 27 49


packages/html/ds/src/tabs/tabs.html
BogdanDanilescu, line 1: I would separate the macros into their own files just to have readability and keep the consistency compared to other components.

packages/html/ds/scripts/build-properties.ts
```diff
@@ -66,7 +66,7 @@ function getSchemaProperties(schema) {
           description,
           type: toType(zodType),
           required,
-          values: zodEnums ? zodEnums : undefined,
+          values: zodEnums || undefined,
```
Thurfjell, line 69: Can `(value as any)._def.values` be 0, "" or null ? Otherwise why is the || needed?

Thurfjell (approved):

hamza14khan (approved):

hamza14khan:
Would be great if you can add Screenshot/GIFs for tabs.

hamza14khan (changes requested):
Looks good, can you please also update doc site and examples?

packages/html/ds/scripts/build-properties.ts
pregno: good point

hamza14khan (approved):
Thanks for taking care of the comments!

PR #240, opened 2024-09-25
feat: react select component

BogdanDanilescu:
### Global React select component
`id`
`label`
`options`
`hint`
`error`

**SB Screenshots**
image
image
image


packages/react/ds/src/select/select.stories.ts
```diff
@@ -0,0 +1,206 @@
...
+      },
+    },
+  },
+  args: {
+    id: 'unique-id',
+    label: {
+      content: 'Label',
+    },
+    hint: {
+      content: '',
```
hamza14khan, line 69: I have updated this with `children` since we dont have `content` with the react props like you suggested. Can you please check my PR, I have fixed your comment there and then maybe apply it here. Let me know what you think.

pregno (approved):

packages/react/ds/src/select/select.stories.ts
BogdanDanilescu: I could make the changes now but I would wait for all of us to have the discussion in regards to this. I've created a [ticket ](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21977)that tackles this. I think once we decide on a solution we can make this change across the board

hamza14khan (approved):

PR #246, opened 2024-09-27
docs: add layout template

pregno:


packages/html/ds/tailwind.config.ts
```diff
@@ -6,6 +6,11 @@ const config: Config = {
   content: ['./src/**/*.html', './src/**/*.ts'],
   theme: createTheme(),
   plugins: [],
+  safelist: [
+    // { pattern: /text-./ },
+    // { pattern: /border-./ },
+    // { pattern: /./ },
+  ],
```
BogdanDanilescu, line 13: do we need this commented code ?

pregno: I've updated it with react package version

BogdanDanilescu (approved):

PR #250, opened 2024-09-30
chore: refactor children prop

BogdanDanilescu:
As part of [21977](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21977) I've refactored the `children` into a `text` prop so now it accepts `string` instead of a react node.


pregno (approved):
LGTM

hamza14khan (approved):
Thanks for taking care of this! 👍

PR #251, opened 2024-09-30
fix: send formdata on submit and add search url prop

BogdanDanilescu:
### HTML Header Component

- send the search query on the search functionality within the `Header` component.
- Add `searchUrl` prop in order to properly redirect the user to the search page

### React Header Component

- send the search query on the search functionality within the `Header` component.
- Add `searchUrl` prop in order to properly redirect the user to the search page
- Refactor and add props to make the header more customisable 
- Implement the ability to add extra buttons either with or without label/icon
- Add an optional title to the Header
Screenshots
image
image
image
image

### DIFF
--- a/packages/react/ds/src/header/header.tsx
+++ b/packages/react/ds/src/header/header.tsx
@@ -4,8 +4,10 @@ import { Icon } from '../icon/icon.js';
 import HeaderMenu from './components/header-menu.js';
 import HeaderSearch from './components/header-search.js';
 import { IconId } from '../icon/icon.js';
+import { Heading } from '../heading/heading.js';
 
 export type HeaderProps = {
+  title?: string;
   logo: {
     image?: string;
     href: string;
@@ -32,7 +34,13 @@ export type HeaderProps = {
   }[];
 };
 
-export function Header({ tools, logo, languages, navLinks }: HeaderProps) {
+export function Header({
+  title,
+  tools,
+  logo,
+  languages,
+  navLinks,
+}: HeaderProps) {
   const hasDivider = tools?.items || tools?.search;
   return (
     <header
@@ -59,7 +67,7 @@ export function Header({ tools, logo, languages, navLinks }: HeaderProps) {
         id="MenuContainer"
         className="gi-h-20 gi-justify-between gi-items-center gi-flex gi-bg-emerald-800 gi-relative gi-py-3 gi-px-4 sm:gi-px-8 sm:gi-py-4"
       >
-        <div className="gi-flex">
+        <div className="gi-flex gi-items-center gi-gap-3">
           <a href={logo.href} className="xs:gi-block gi-hidden">
             {logo.image ? (
               <img
@@ -80,6 +88,12 @@ export function Header({ tools, logo, languages, navLinks }: HeaderProps) {
               <GovieLogoSmall />
             )}
           </a>
+          {/* Replace the h2 with the Heading Component after refactor */}
+          {title && (
+            <h2 className="gi-tracking-wider gi-font-bold gi-text-white gi-text-lg xs:gi-text-lg md:gi-text-xl xl:gi-text-2xl">
+              {title}
+            </h2>
+          )}
         </div>
         <div className="gi-flex gi-items-center">
           <ul className="gi-hidden sm:gi-flex gi-gap-4 gi-text-white">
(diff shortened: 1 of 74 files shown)


packages/react/ds/src/header/header.tsx
```diff
@@ -6,6 +6,7 @@ import HeaderSearch from './components/header-search.js';
 
 export type HeaderProps = {
   logoLink: string;
+  search_url: string;
```
pregno, line 9: I will set this optional or have another flag to define whether the search is active.
In some cases, the search will not be part of the application, and we will need to hide the button.

pregno (changes requested):

packages/react/ds/src/header/header.tsx
BogdanDanilescu: good idea. I will put this as optional and if the user doesn't use this prop, the search component will not render

pregno (changes requested):
I suggest to review the headers props using something like this.
```
{
  "logo": {
    "image": "...", //optional, url of the image, default is embedded svg
    "href": "/" // required, link of the logo
  },
  "tools": {
    "search": {
      "enable": true, // required, if false we hide the button
      "action": "...", // optional, form action of the search page
      "label": "Search", // optional, default hidded
      "icon": "icon-magnifier", // optional, default icon-magnifier
      "noJsHref": "..." // fallback link for when js is not available
    },
    "menu": {
      "label": "Menu", // optional, default hidded
      "icon": "icon-hamburger", // optional, default icon-hamburger
      "noJsHref": "..." // fallback link for when js is not available
    },
    "items": [ // optional
      { 
        "label": "Home", // optional
        "icon": "icon-home", // required
        "href": "/" // required
      },
      {

      }
    ]
  },
  "links": [
    {
      "label": "nav 1",
      "href": "..."
    },
    {
      "label": "nav 2",
      "href": "..."
    }
  ],
  "languages": [
    {
      "label": "English",
      "href": "..."
    },
    {
      "label": "Gaelic",
      "href": "..."
    }
  ]
}
```

packages/react/ds/src/header/header.tsx
```diff
@@ -6,6 +6,7 @@ import HeaderSearch from './components/header-search.js';
 
 export type HeaderProps = {
   logoLink: string;
+  search_url?: string;
```
hamza14khan, line 9: suggestion: If we can change search_url to camelCase convention?

hamza14khan (changes requested):
Looks good. Just one minor comment and Andrea's feedback.

packages/react/ds/src/header/header.tsx
```diff
@@ -42,13 +66,33 @@ export function Header({ logoLink, languages, navLinks }: HeaderProps) {
...
       >
-        <div className="gi-flex">
-          <a href={logoLink} className="xs:gi-block gi-hidden">
-            <GovieLogo />
+        <div className="gi-flex gi-items-center gi-gap-3">
+          <a href={logo.href} className="xs:gi-block gi-hidden">
+            {logo.image ? (
+              <img
+                className="gi-object-contain gi-h-12"
+                src="https://ds.blocks.gov.ie/_next/static/media/hero.01b46751.png"
```
pregno, line 74: You should add the `{logo.image}` here as `src`

packages/react/ds/src/header/header.tsx
```diff
@@ -42,13 +66,33 @@ export function Header({ logoLink, languages, navLinks }: HeaderProps) {
...
+              <GovieLogo />
+            )}
           </a>
-          <a href={logoLink} className="xs:gi-hidden gi-block">
-            <GovieLogoSmall />
+          <a href={logo.href} className="xs:gi-hidden gi-block">
+            {logo.image ? (
+              <img
+                className="gi-object-contain gi-h-10"
+                src="https://ds.blocks.gov.ie/_next/static/media/hero.01b46751.png"
```
pregno, line 84: Same as above

pregno (changes requested):
LGTM, just a minor fix

packages/react/ds/src/header/header.tsx
BogdanDanilescu: fixed

pregno (approved):

hamza14khan (approved):

PR #252, opened 2024-10-01
fix: improve link doc with tabs

pregno:
Link - Gov IE Design System · 9 53am · 10-01

### DIFF
--- a/apps/docs/content/3-components/2-library/22-link.mdx
+++ b/apps/docs/content/3-components/2-library/22-link.mdx
@@ -2,60 +2,193 @@
 title: Link
 description: Link HTML Component
 draft: false
-status: in-development
+status: stable
 ---
 
 # Link
 
 <ComponentStatusBlock componentId="link" />
 
-## Component
+Links are blue and underlined by default. If your link is at the end of a sentence or paragraph, make sure that the linked text does not include the full stop.
 
-<Tabs>
+<Paragraph>
+    <Link href="#">Link with blue color</Link>
+</Paragraph>
+
+<Tabs id="tab1">
     <TabList>
-        <TabItem value="tab1" checked>HTML</TabItem>
-        <TabItem value="tab2">Macro</TabItem>
-        <TabItem value="tab3">React</TabItem>
+        <TabItem value="tab11" checked>HTML</TabItem>
+        <TabItem value="tab12">Macro</TabItem>
+        <TabItem value="tab13">React</TabItem>
     </TabList>
-    <TabPanel value="tab1">
-        <Paragraph>
-        Base link with blue color: 
-        <a href="#" className="gi-link-base gi-link-blue">Base link with blue color</a>
+    <TabPanel value="tab11">        
         ```html
-        <a href="#" class="gi-link-base gi-link-blue"></a>
-        ```
-        </Paragraph>
-
-        <Paragraph>
-        Base link with inherited color: <a href="#" className="gi-link-base">Base link with inherited color</a>
-        ```html
-        <a href="#" class="gi-link-base"></a>
-        ```
-        </Paragraph>
-
-        <Paragraph>
-        Base link with no underline: <a href="#" className="gi-link-base gi-link-blue gi-link-no-underline">Base link with inherited color</a>
-        ```html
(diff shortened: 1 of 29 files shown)


apps/docs/content/3-components/2-library/22-link.mdx
hamza14khan, line 1: suggestion: Can we use 2 spacing tabs formatting? Its just that the formatting is not consistent with the other files.

packages/html/ds/src/link/link.test.ts
```diff
@@ -58,8 +58,10 @@ describe('link', () => {
     });
     const linkElement = screen.getByRole('link');
 
-    expect(linkElement.classList.contains('visited:gi-text-purple-700')).toBe(
-      true,
+    console.log(linkElement.classList);
```
hamza14khan, line 61: nitpick: Must be here mistakenly, can we please remove it? 😄

BogdanDanilescu, line 61: can we remove the console.log ?

packages/design/tailwind/css/layout.css
hamza14khan, line 1: Formatting issue, same for this file.

packages/design/tailwind/css/components.css
hamza14khan, line 1: Formatting issue, same for this file.

hamza14khan (approved):
Looks good, minor changes requested.

packages/html/ds/src/link/link.test.ts
pregno: ops, sorry

apps/docs/content/3-components/2-library/22-link.mdx
```diff
@@ -2,13 +2,197 @@
...
+    </TabList>
+    <TabPanel value="tab41">        
+        ```html
+        <a href="#" class="gi-link gi-link-no-visited">Link with no visited color</a>
+        ```
+    </TabPanel>
+    <TabPanel value="tab42">        
+        ```html
+        {{ 
+            govieLink({
```
hamza14khan, line 137: The only problem I see with this approach is that we will have to consistently check and update the docs if there are any props or updates to the component.

hamza14khan: Not sure if there's an alternative way.

apps/docs/src/components/document/common/mdx.tsx
```diff
@@ -142,4 +164,5 @@ export function Mdx({ code }: MdxProps) {
 
 function MarginBottom({ children }: { children: React.ReactNode }) {
   return <div className="mb-2xl">{children}</div>;
+  // return <>{children}</>;
```
BogdanDanilescu, line 167: do we need this line ?

apps/docs/content/3-components/2-library/22-link.mdx
BogdanDanilescu: Just thinking out loud here . What we are trying to achieve here, rendering each component with their different variants and states, doesn't storybook accomplish the same purpose ? Gives the users the ability to look at the components and the code needed to implement the component.

apps/docs/content/3-components/2-library/22-link.mdx
pregno: done

packages/design/tailwind/css/layout.css
pregno: I don't get it; what should it look like?

hamza14khan: I think this is still using 4 spaces tab formatting. It should be 2.

apps/docs/content/3-components/2-library/22-link.mdx
pregno: It's more a showcase of the components.
The idea is to have complete documentation where the Product team, design, and developer can find all the information needed.
I believe that Storybook is a tool for a more tech target.
Also, I've created explicit classes for the HTML case. There can be users who don't have React or Nunjucks. Storybook is almost useless for them.

I agree that maintaining this can be cumbersome, but I think it is a great value as a final product.

BogdanDanilescu (approved):

PR #260, opened 2024-10-02
fix: add server action support

BogdanDanilescu:
Following a conversation with Ludwig and Andrea, a issue has come up in regards to using Server Actions in the Search component of the Header. Since react/types does not support that we have come up with a temporary workaround until React catches up

### DIFF
--- a/packages/react/ds/src/header/components/header-menu.tsx
+++ b/packages/react/ds/src/header/components/header-menu.tsx
@@ -11,7 +11,8 @@ type HeaderMenuProps = {
     label: string;
   }[];
   searchProps?: {
-    action: string;
+    action?: string;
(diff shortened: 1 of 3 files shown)


packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -11,7 +11,8 @@ type HeaderMenuProps = {
     label: string;
   }[];
   searchProps?: {
-    action: string;
+    action?: string;
+    serverAction?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
pregno, line 15: Can you please add a comment explaining the workaround?

pregno (approved):

PR #264, opened 2024-10-03
feat: html button component

BogdanDanilescu:
### Implementation of the HTML Button component

Implementation of the Button component which contains 3 variants and each of the variant has 3 appearances. There is also the option the render a icon button by omitting the label prop. There is also the ability to add a disabled state whenever necessary

_Props:_
`label`
`variant`
`appearance`
`size`
`disabled`
`icon`

Screenshots
Screenshot 2024-10-03 at 15 40 17
Screenshot 2024-10-03 at 15 40 33
Screenshot 2024-10-03 at 15 40 53
Screenshot 2024-10-03 at 15 41 23
Screenshot 2024-10-03 at 15 41 36
Screenshot 2024-10-03 at 15 41 46


hamza14khan (approved):
Looks great! Just missing the documentation update.

pregno (approved):

packages/html/ds/src/button/button.html
```diff
@@ -0,0 +1,18 @@
...
+{% from 'icon/icon.html' import govieIcon %}
+
+{% macro govieButton(props) %}
+
+{% set appearance = props.appearance or 'default' %}
+{% set variant = props.variant or 'primary' %}
+{% set size = props.size or 'medium' %}
+{% set isDisabled = 'disabled' if props.disabled else 'notDisabled' %}
+
+<button data-testid="govieButton-{{ appearance }}-{{ variant }}-{{ size }}-{{ isDisabled }}" class="gi-gap-2 gi-flex gi-rounded-sm focus:gi-outline focus:gi-outline-[3px] focus:gi-outline-yellow-400 focus:gi-outline-offset-0 gi-items-center {{ getVariantAppearanceClass(props.disabled, props.variant, props.appearance) | trim }} {{getSizeClass(props.size) | trim }} {{getIconPosition(props.icon.position) | trim }} {{isButtonDisabled(props.disabled, props.variant, props.appearance) | trim }}">
```
pregno, line 11: Please, add these classes to `packages/design/tailwind/css/components.css` in a new class named, for example, `gi-button`.
Add the variants as separate classes to reuse on React and plain HTML.

pregno (changes requested):

packages/html/ds/src/button/button.html
BogdanDanilescu: Added the classes

BogdanDanilescu:
> Looks great! Just missing the documentation update.

Hi @hamza14khan I was planning the add the docs site in the React PR so that we have both libs when I add it

pregno (approved):

PR #268, opened 2024-10-06
feat: react button component

BogdanDanilescu:
### Implementation of the HTML Button component

Implementation of the Button component which contains 3 variants and each of the variant has 3 appearances. There is also the option the render a icon button by omitting the label prop. There is also the ability to add a disabled state whenever necessary

Docs have also been updated

Props:
```
label
variant
appearance
size
disabled
icon
iconEnd
```
image
image
image
image
image

### DIFF
--- a/apps/docs/content/3-components/2-library/5-button.mdx
+++ b/apps/docs/content/3-components/2-library/5-button.mdx
@@ -2,13 +2,249 @@
 title: Button
 description: Button HTML Component
 draft: false
-status: in-development
+status: stable
 ---
 
 # Button
 
 <ComponentStatusBlock componentId="button" />
 
-## When to use this component
+## Primary buttons
 
-## When not to use this component
+<Button label="Primary Button" />
+
+<Tabs id="tab1">
+  <TabList>
+    <TabItem value="tab11" checked>HTML</TabItem>
+    <TabItem value="tab12">Macro</TabItem>
+    <TabItem value="tab13">React</TabItem>
+  </TabList>
+  <TabPanel value="tab11">
+  ```html
+  <button id="button" data-module="govie-button" class="govie-button">Primary Button</button>
+  ```
+  </TabPanel>
+  <TabPanel value="tab12">
+  ```html
+   {{ govieButton({
+    "label": "Primary button"
+  }) }}
+  ```
+  </TabPanel>
+  <TabPanel value="tab13">
+  ```html
+  import { Button } from '@govie-ds/react';
+
+  <Button label="Primary Button" />
+  ```
+  </TabPanel>
+</Tabs>
+
+## Secondary buttons
+
+<Button label="Secondary Button" variant="secondary" />
+
+<Tabs id="tab2">
+  <TabList>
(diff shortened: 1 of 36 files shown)


packages/react/ds/src/button/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  Medium = 'medium',
+  Small = 'small',
+  Large = 'large',
+}
+
+export type ButtonProps = {
+  variant?: ButtonVariant;
+  appearance?: ButtonAppearance;
+  size?: ButtonSize;
+  label?: string;
```
pregno, line 25: See comment below about icons

packages/html/ds/src/button/button.html
```diff
@@ -8,11 +8,13 @@
 {% set size = props.size or 'medium' %}
 {% set isDisabled = 'disabled' if props.disabled else 'notDisabled' %}
 
-<button data-element="button-container" data-module="gieds-button" data-testid="govieButton-{{ appearance }}-{{ variant }}-{{ size }}-{{ isDisabled }}" class="gi-btn {{ getVariantAppearanceClass(props.disabled, props.variant, props.appearance) | trim }} {{getSizeClass(props.size) | trim }} {{getIconPosition(props.icon.position) | trim }} {{isButtonDisabled(props.disabled, props.variant, props.appearance) | trim }}">
+<button data-element="button-container" data-module="gieds-button" data-testid="govieButton-{{ appearance }}-{{ variant }}-{{ size }}-{{ isDisabled }}" class="gi-btn {{ getVariantAppearanceClass(props.disabled, props.variant, props.appearance) | trim }} {{getSizeClass(props.size) | trim }} {{getIconPosition(props.iconEnd) | trim }} {{isButtonDisabled(props.disabled, props.variant, props.appearance) | trim }}">
   {% if props.icon %}
-    {{ govieIcon(props.icon.props)}}
+    {{ govieIcon(props.icon)}}
+  {% endif %}
+  {% if props.label %}
```
pregno, line 15: You can apply the same logic here as I suggested in React.
If you have the property `content,` you can pass text and HTML (as we did for containers or paragraphs).
In this way, we can leave the developer to decide how to place the icon and the content.

apps/docs/content/3-components/2-library/5-button.mdx
```diff
@@ -2,13 +2,249 @@
 title: Button
 description: Button HTML Component
 draft: false
-status: in-development
+status: stable
```
hamza14khan, line 5: Shouldn't this be in-development? @pregno

hamza14khan (changes requested):
Looks good. Happy to approve it once Andrea comments are resolved.

hamza14khan (approved):
Looks good, just one comment left for @pregno

apps/docs/content/3-components/2-library/5-button.mdx
pregno: I think that is the status of the documentation.
Below there is the status of the library and is set to `alpha`

pregno (approved):

PR #269, opened 2024-10-07
fix: header improvements

pregno:
* update HTML schema to match React
* set label for tools optional
* set logo href optional
* review font for title


BogdanDanilescu (approved):
LGTM but I think there are some unit tests that are failing

hamza14khan (approved):

PR #270, opened 2024-10-07
chore: update tag types to sync with figma values

hamza14khan:
Updated tag types with the values from figma.

- default as `gray`.
- info as `blue`.
- success as `green`.
- yellow as `warning`.
- red as `error`.

Testing:
Screenshot 2024-10-07 at 12 45 21

### DIFF
--- a/packages/design/tailwind/css/components.css
+++ b/packages/design/tailwind/css/components.css
@@ -101,25 +101,26 @@
 /* Tag */
 
 .gi-tag {
-  @apply gi-rounded-md gi-border-xs gi-font-normal gi-py-0 gi-px-2 gi-inline-flex gi-justify-center gi-items-center;
+  @apply gi-rounded-sm gi-border-xs gi-font-bold gi-leading-6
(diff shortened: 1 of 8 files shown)


packages/design/tailwind/css/components.css
```diff
@@ -101,25 +101,26 @@
 /* Tag */
 
 .gi-tag {
-  @apply gi-rounded-md gi-border-xs gi-font-normal gi-py-0 gi-px-2 gi-inline-flex gi-justify-center gi-items-center;
+  @apply gi-rounded-sm gi-border-xs gi-font-bold gi-leading-6
```
BogdanDanilescu, line 104: can we fix the spacing here ? it seems to be on 2 lines

BogdanDanilescu (approved):
LGTM, left a small comment

packages/design/tailwind/css/components.css
hamza14khan: Good catch!

PR #271, opened 2024-10-08
fix: review header

pregno:
* restore default in theme
* fix header drawer

### DIFF
--- a/packages/html/ds/src/header/components/headerMenu.html
+++ b/packages/html/ds/src/header/components/headerMenu.html
@@ -1,6 +1,6 @@
-{% from 'header/components/headerSearchContainer.html' import
-HeaderSearchContainer %} {% from 'icon/icon.html' import govieIcon %} {% macro
-headerMenu(props) %}
+{% from 'header/components/headerSearchContainer.html' import HeaderSearchContainer %} 
+{% from 'icon/icon.html' import govieIcon %}
+{% macro headerMenu(props) %}
 
 <div
   data-element="menu-container"
@@ -32,7 +32,8 @@ headerMenu(props) %}
         <span class="gi-font-bold gi-text-sm">{{ link.label }}</span>
       </a>
     </li>
-    {% endfor %} {% for link in props.languages %}
+    {% endfor %} 
+    {% for link in props.languages %}
     <li>
       <a
         href="{{ link.href }}"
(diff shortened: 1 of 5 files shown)


packages/react/ds/src/header/header.tsx
```diff
@@ -185,6 +185,11 @@ export function Header({
               id="MobileMenuTrigger"
               type="checkbox"
             />
+            {tools?.menu?.label && (
+              <span className="gi-hidden sm:gi-block gi-text-2md gi-font-bold gi-text-white">
+                {tools?.menu?.label}
```
hamza14khan, line 190: Minor: If `tools?.menu?.label` is true then maybe replace it with `tools.menu.label`

packages/html/ds/src/header/components/headerMenu.html
```diff
@@ -38,11 +41,11 @@
         <span class="gi-text-sm">{{ link.label }}</span>
       </a>
     </li>
-    {% endfor %}
-    {% if props.searchUrl %}
-      <li class="xs:gi-hidden gi-mt-8">
-        {{ HeaderSearchContainer({"searchUrl": props.searchUrl, "dataElement": "search-container-mobile" , "dataTestId": "search-container-mobile"}) }}
-      </li>
+    {% endfor %} {% if props.searchUrl %}
```
hamza14khan, line 44: question: Is this formatting changes? If yes, then can we please revert it back to the original one as its more readable 😄

hamza14khan (approved):
Few non blocking comments. Looks good other than that.

packages/html/ds/src/header/components/headerMenu.html
```diff
@@ -1,35 +1,38 @@
-{% from 'header/components/headerSearchContainer.html' import HeaderSearchContainer %}
-{% from 'icon/icon.html' import govieIcon %}
-
-{% macro headerMenu(props) %}
+{% from 'header/components/headerSearchContainer.html' import
+HeaderSearchContainer %} {% from 'icon/icon.html' import govieIcon %} {% macro
+headerMenu(props) %}
```
BogdanDanilescu, line 3: same here for formatting issues

BogdanDanilescu (approved):
LGTM with a minor comment about formatting

PR #273, opened 2024-10-08
chore: add jinja config for prettier

pregno:


packages/html/ds/prettier.config.mjs
hamza14khan, line 1: minor: formatting to tab size 2 😄

BogdanDanilescu (approved):

hamza14khan (approved):
one minor comment other than that looks great!

PR #274, opened 2024-10-08
fix: ui review

pregno:


packages/html/ds/src/file-upload/file-upload.html
```diff
@@ -10,7 +10,7 @@
   {% if isError %}
     <div
       class="gi-pt-2 gi-mb-4 gi-px-4 gi-border-solid gi-border-l-lg gi-border-red-600"
-    ></div>
```
hamza14khan, line 13: Will this be an error? 🤔

packages/html/ds/src/heading/heading.schema.ts
```diff
@@ -30,7 +30,7 @@ export const headingSchema = zod.object({
     required_error: 'A tag element is required',
   }),
   text: zod.string({
-    description: 'Ttitle of the heading',
+    description: 'Title of the heading',
```
hamza14khan, line 33: praise: Good catch!

hamza14khan (approved):

packages/html/ds/src/file-upload/file-upload.html
pregno: it's closed below with
```
{% if isError %}
</div>
{% endif %}
```
I added by mistake in the previous PR

PR #277, opened 2024-10-09
fix: label spacing in inputs

pregno:


packages/html/ds/src/textarea/textarea.stories.ts
```diff
@@ -98,12 +98,6 @@ export const Default: Story = {
       content: 'Textarea Label',
       for: 'textarea-id',
     },
-    error: {
-      content: '',
-    },
-    hint: {
-      content: '',
-    },
   },
```
BogdanDanilescu, line 101: I think this was left here intentional so that the user can easily add a string in storybook. @hamza14khan is that right ?

pregno: Restored

BogdanDanilescu (approved):

PR #293, opened 2024-10-16
feat: update HTML and React Card component with Inset, Tag, Action and Subheading

hamza14khan:
There were some missing functionalities on card:

- Tag.
- Button slot.
- Subheading.
- Inset.

Testing on Doc site:
Screenshot 2024-10-16 at 17 56 06
Screenshot 2024-10-16 at 17 56 11


packages/design/tailwind/css/components.css
```diff
@@ -318,13 +318,22 @@ input[type='file' i] {
 }
 
 .gi-card-vertical {
-  @apply gi-flex-col gi-max-w-[348px] gi-min-w-[220px] gi-gap-2 gi-w-full;
+  @apply gi-flex-col gi-max-w-[300px] gi-min-w-[220px] gi-gap-2 gi-w-full;
```
hamza14khan, line 321: @pregno

packages/html/ds/src/card/card.stories.ts
```diff
@@ -70,31 +79,44 @@ export const Default: Story = {
...
+        type: { summary: 'Tag' },
+      },
+    },
+    action: {
+      control: 'object',
+      description: 'Single action button or link for the card.',
       table: {
         category: 'Actions',
-        type: { summary: 'Array<{ href: string, text: string }>' },
+        type: { summary: '{ type: "link" | Button | Link }' },
```
BogdanDanilescu, line 95: is this right ? shouldn't it be `{ summary: '{ type: | Button | Link }' }` ?

hamza14khan: I should rephrase it a bit better because action is going to have `type` property that defines link or button.

Let me know what you think of this?|

E.g.

```
action: {
  control: 'object',
  description: 'Single action for the card, either a button or link, determined by the "type" property.',
  table: {
    category: 'Actions',
    type: { summary: '{ type: "button" | "link", ...Button | Link }' },
  },
},
```

BogdanDanilescu: yeah, I think that works

BogdanDanilescu (approved):
Great work on the card, LGTM

ramonmulia (approved):

pregno (changes requested):
- inset variants are 3:
  - none: no spacing, image and content aligned
  - body: only content has 16px margin
  - full: image and content has 16px margin, between image and content it's only 16px
    - when inset is enabled, there is a gap between the image and content, so the margin-top (with vertical) or margin-left (with horizontal) increase
- image aspect-ratio has 2 variants:
  - 5:3
  - 1:1
  - check this https://tailwindcss.com/docs/aspect-ratio

ramonmulia (approved):

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
       )}
-      <div className="gi-card-content">
-        <div className="gi-card-heading">
-          <Heading as="h5" customClasses="!gi-my-0">
-            {href ? <Link href={href}>{title}</Link> : title}
-          </Heading>
+      <div className={`gi-card-content ${getInsetClass()}`}>
+        <div className="gi-card-header">
+          <div className="gi-card-heading">
+            <Heading
```
pregno, line 73: Use the heading classes, but don't use the headings tags (eg `e5`) because we don't know what will be the hierarchy of the content.

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
+              as="h5"
+              customClasses={`!gi-my-0 ${subTitle ? '!gi-mb-2' : ''}`}
+            >
+              {href ? <Link href={href}>{title}</Link> : title}
```
pregno, line 77: We should limit the title length to be maximum on two lines (check with Lexa if two lines are fine)

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
+            </Heading>
+            <Paragraph
+              size="sm"
+              className="gi-text-gray-600 !gi-font-normal !gi-my-0"
```
pregno, line 81: Why `!gi-font-normal`?

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
+            >
+              {subTitle}
```
pregno, line 83: Also this field should have a length limitation

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
+            </Paragraph>
+          </div>
+          {tag && tag.text && tag.type && (
+            <div className="gi-card-tag">
+              <Tag text={tag.text} type={tag.type} />
+            </div>
+          )}
         </div>
         <div className="gi-card-paragraph">
           <Paragraph size="sm">{content}</Paragraph>
```
pregno, line 93: Also the content should have a length limitation

packages/react/ds/src/card/card.tsx
```diff
@@ -42,24 +67,43 @@ export const Card = (props: CardProps) => {
...
+      <div className={`gi-card-content ${getInsetClass()}`}>
+        <div className="gi-card-header">
+          <div className="gi-card-heading">
+            <Heading
+              as="h5"
+              customClasses={`!gi-my-0 ${subTitle ? '!gi-mb-2' : ''}`}
+            >
+              {href ? <Link href={href}>{title}</Link> : title}
+            </Heading>
+            <Paragraph
```
pregno, line 79: Don't use the Paragraph here. You can use the same classes or have one dedicated for the card subheading

pregno (changes requested):
We should limits the characters on some fields to avoid layout disruptions.

packages/html/ds/src/card/card.schema.ts
```diff
@@ -1,16 +1,17 @@
...
-    description: 'Text for the action link',
-    required_error: 'text is required for actions',
+  linkSchema.extend({
+    type: zod.literal('link').describe('Type of action is a link'),
   }),
-});
+]);
 
 export const cardSchema = zod.object({
   type: zod.enum(['vertical', 'horizontal'], {
```
ramonmulia, line 17: I believe you should replace `zod.enum` to `zod.nativeEnum`, otherwise it will raise an error while building `html/ds`

Type 'string[]' is not assignable to type 'Record<string, string>'.
  Index signature for type 'string' is missing in type 'string[]'

packages/html/ds/src/card/card.schema.ts
```diff
@@ -31,16 +37,20 @@ export const cardSchema = zod.object({
       description: 'Image URL for the card (if applicable)',
     })
     .optional(),
+  inset: zod
+    .enum(['body', 'full', 'none'], {
```
ramonmulia, line 41: the same above.

packages/react/ds/src/card/card.tsx
```diff
@@ -1,61 +1,136 @@
...
+  const renderTitle = () => {
+    const titleContent = href ? (
+      <Link href={href}>{truncatedTitle}</Link>
+    ) : (
+      truncatedTitle
+    );
+
+    return (
+      <Heading
+        as="h5"
```
pregno, line 94: You shouldn't use an `h5` here; instead, use a `div` with the same style as an `h5`.

packages/react/ds/src/card/card.tsx
```diff
@@ -1,61 +1,136 @@
...
+
+  return (
+    <div className={cardClasses}>
+      {renderMedia()}
+      <div className={`gi-card-content gi-card-inset-${inset}`}>
+        <div className="gi-card-header">
+          <div className="gi-card-heading">
+            {renderTitle()}
+            {truncatedSubTitle && (
+              <Paragraph size="sm" className="gi-card-subheading">
```
pregno, line 117: Use a `div` not `Paragraph`, otherwise if we change something in the Paragraph the card will be affected.

packages/react/ds/src/card/card.tsx
```diff
@@ -1,61 +1,136 @@
...
 };
 
-export const Card = (props: CardProps) => {
-  const { type = 'vertical', title, img, icon, content, actions, href } = props;
-  const layoutClasses = {
-    vertical: 'gi-card gi-card-vertical',
-    horizontal: 'gi-card gi-card-horizontal',
-  };
+// Memoized truncation utility
+const useTruncatedText = (text: string | undefined, maxLength: number) => {
```
pregno, line 28: use [text overflow](https://tailwindcss.com/docs/text-overflow) instead of this function.

pregno (changes requested):
The same comments also apply for HTML version.

pregno (approved):

PR #294, opened 2024-10-17
feat: maxChar prop for textarea

ramonmulia:
### PR for adding maxChar property in textarea

- add global CSS classes
- add stories
- add tests
- add example in Wagtail, vite, nextjs
- add documentation

new prop:
`maxChar`

story:
Screenshot 2024-10-17 at 09 01 03

### DIFF
--- a/packages/html/ds/src/textarea/textarea.ts
+++ b/packages/html/ds/src/textarea/textarea.ts
@@ -1,0 +1,48 @@
+import {
+    BaseComponent,
+    BaseComponentOptions,
+    initialiseModule,
+  } from '../common/component';
+  
+  export type TextareaProps = BaseComponentOptions;
+  
+  export class Textarea extends BaseComponent<TextareaProps> {
+    calculateRemainingChars: () => void;
+    getAllTextarea: NodeListOf<HTMLTextAreaElement>;
+
+    constructor(options: TextareaProps) {
+      super(options);
+      this.getAllTextarea = document.querySelectorAll('textarea')
+       
+      this.calculateRemainingChars = () => {
+        for (const textarea of this.getAllTextarea) {
+        const {id, value, maxLength} = textarea
+        const remainingCharsContainer = document.querySelector(`div[data-remaining-chars-container=${id}]`)
+        
+        if(textarea?.maxLength > -1 && remainingCharsContainer){
+          const container = remainingCharsContainer.firstElementChild
+          if(container){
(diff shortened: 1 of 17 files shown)


BogdanDanilescu:
edge case scenario, not sure if we should tackle it but just pointing out. If the user decides on having our textarea component and a normal `textarea` html tag we will get an error in the console

Screenshot 2024-10-17 at 18 37 52

again, not sure if it applies to a real scenario because I don't see a reason why the user would use both but just pointing it out. Otherwise, LGTM. Well done @ramonmulia

BogdanDanilescu (approved):

hamza14khan (approved):

pregno (changes requested):
@ramonmulia just a minor change after @BogdanDanilescu input

packages/html/ds/src/textarea/textarea.ts
```diff
@@ -0,0 +1,53 @@
...
+  getAllTextarea: NodeListOf<HTMLTextAreaElement>;
+
+  constructor(options: TextareaProps) {
+    super(options);
+    this.getAllTextarea = document.querySelectorAll('textarea');
+
+    this.calculateRemainingChars = () => {
+      for (const textarea of this.getAllTextarea) {
+        const { id, value, maxLength } = textarea;
+        const remainingCharsContainer = document.querySelector(
```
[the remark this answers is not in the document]
ramonmulia: try catch added.

packages/html/ds/src/textarea/textarea.ts
```diff
@@ -0,0 +1,57 @@
...
+export class Textarea extends BaseComponent<TextareaProps> {
+  calculateRemainingChars: () => void;
+  getAllTextarea: NodeListOf<HTMLTextAreaElement>;
+
+  constructor(options: TextareaProps) {
+    super(options);
+    this.getAllTextarea = document.querySelectorAll('textarea');
+
+    this.calculateRemainingChars = () => {
+      try {
```
pregno, line 18: Put it inside the for-loop.
In this way, if you have two textarea, the first without the counter and the second with the counter, it will quit too soon

pregno (changes requested):

packages/html/ds/src/textarea/textarea.ts
ramonmulia: good catch! fixed.

pregno (approved):

PR #296, opened 2024-10-18
feat: add `cookie banner`

BogdanDanilescu:
PR to implement the `cookie banner` component on HTML and React

props:
`children`
`accept`
`reject`
`dismissButton`
`cookieLink`

**SB Screenshots**
Screenshot 2024-10-18 at 15 56 52
Screenshot 2024-10-18 at 15 57 06
Screenshot 2024-10-18 at 15 57 17

_The background color was not matching our tailwind pallete, awating feedback from design before making the necessary changes_


packages/react/ds/src/cookie-banner/cookie-banner.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    {cloneElement(reject.triggerButton, {
+                      onClick: handleCloseBanner,
+                    })}
+                    {cookieLink && cookieLink}
+                  </div>
+                </>
+              )}
+
+              {/* accepted container of cookie banner */}
+              {status === true && (
```
hamza14khan, line 55: suggestion: just status would be a lot cleaner?

BogdanDanilescu: the status has the following type `<null | boolean | 'dismiss'>` if I set it to just `status` it will check if it's truthy, which will be true for `true` and `dismiss`

packages/react/ds/src/cookie-banner/cookie-banner.tsx
```diff
@@ -0,0 +1,87 @@
...
+                      {cloneElement(dismissButton, {
+                        onClick: handleDismissBanner,
+                      })}
+                    </>
+                  )}
+                </>
+              )}
+
+              {/* rejected container of cookie banner */}
+              {status === false && (
```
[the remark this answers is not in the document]
BogdanDanilescu: same here . If I set it to `!status` it will be true for `false` and `null`

ramonmulia (approved):
LGTM

PR #297, opened 2024-10-18
feat: add list

ramonmulia:
### PR for adding `list` component

- add global CSS classes
- add stories
- add tests
- add example in Wagtail, vite, nextjs
- add documentation

Screenshot 2024-10-18 at 11 59 09

### DIFF
--- a/apps/docs/src/components/document/common/mdx.tsx
+++ b/apps/docs/src/components/document/common/mdx.tsx
@@ -12,6 +12,7 @@ import {
   Tag,
   RadiosGroup,
   Card,
+  Modal,
 } from '@govie-ds/react';
 import { MDXComponents } from 'mdx/types';
 import { useMDXComponent } from 'next-contentlayer/hooks';
@@ -152,6 +153,7 @@ const documentComponents: MDXComponents = {
   Footer: (props) => <Footer {...props}>{props.children}</Footer>,
   RadiosGroup: (props) => <RadiosGroup {...props} />,
   Card: (props) => <Card {...props}>{props.children}</Card>,
+  Modal: (props) => <Modal {...props}>{props.children}</Modal>,
 };
 
 export function Mdx({ code }: MdxProps) {
--- a/apps/docs/src/components/document/common/mdx.tsx
+++ b/apps/docs/src/components/document/common/mdx.tsx
@@ -13,6 +13,7 @@ import {
   RadiosGroup,
   Card,
   Modal,
+  List,
 } from '@govie-ds/react';
 import { MDXComponents } from 'mdx/types';
 import { useMDXComponent } from 'next-contentlayer/hooks';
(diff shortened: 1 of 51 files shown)


apps/docs/src/components/document/common/mdx.tsx
```diff
@@ -64,6 +64,7 @@ import { ColorPrimitives } from '@/components/document/color/color-primitives';
 import { TwoThirds, TwoThirdsOneThird } from '@/components/layouts/two-thirds';
 import { Highlight } from '@/components/typography/highlight';
 import { Link } from '@/components/typography/link';
+import { List } from '@/components/typography/list';
```
BogdanDanilescu, line 68: I think you can directly import the `List` component from `@govie-ds/react` without having to create a new component in the docs repo

packages/html/ds/src/list/list.schema.ts
```diff
@@ -0,0 +1,16 @@
...
+  Normal = 'normal',
+  Bullet = 'bullet',
+  Number = 'number',
+}
+
+const linkPropsSchema = zod.object({
+  items: zod.array(zod.union([zod.string(), linkSchema])),
+  spaced: zod.boolean().optional(),
+  type: zod.nativeEnum(TypeEnum).optional(),
+});
```
BogdanDanilescu, line 14: I think you'll get an error here when you do `pnpm build` in `packages/html/ds` because there is a description required for each of the properties

ramonmulia: where could I find the description required?

I ran the  pnpm build and I got all succeeded, should I go inside of `packages/html/ds` and run pnpm build?

BogdanDanilescu: yes, you have to go into `packages/html/ds` and there run `pnpm build`. For some reason the errors don't show when you do pnpm build in the root

apps/docs/src/components/document/common/mdx.tsx
BogdanDanilescu: I'm not sure about the others. If they have some other functionality added to them I think they can stay like this

packages/react/ds/src/list/list.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  },
+  argTypes: {
+    items: {
+      control: 'object',
+      type: { name: 'string', required: true },
+      description: 'List of texts to be displayed',
+      defaultValue: ['Item 1', 'Item 2', 'Item 3'],
+    },
+    type: {
+      options: ['normal', 'bullet', 'number'],
```
hamza14khan, line 23: minor: I see you are using enums on html storybook for options but not here, can we please make it consistent? Also, if you can also replace `normal` with enum too or replace `normal` maybe with `none`.

hamza14khan (approved):
Looks good. One minor comment.

packages/react/ds/src/list/list.stories.tsx
ramonmulia: good call!

packages/html/ds/tsconfig.json
```diff
@@ -19,7 +19,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals", "@testing-library/jest-dom"]
```
ramonmulia, line 22: Added this to not raise typescript errors in test files when building the html/ds project.

BogdanDanilescu (approved):

pregno (approved):

PR #299, opened 2024-10-23
feat: improve accessibility tests

pregno:
* Set up the SB accessibility test on CI and CLI
* Update the CI pipeline. Chromatic is blocked if accessibility has errors
* Fixed existing accessibility issues


apps/docs/src/components/document/get-started/developers-advice.tsx
```diff
@@ -1,6 +1,6 @@
 'use client';
 import { Heading } from '@govie-ds/react';
-import { Children, isValidElement, useState } from 'react';
+import React, { Children, isValidElement, useState } from 'react';
```
BogdanDanilescu, line 3: Why did we have to import `React` here ? Are we using it somewhere ?

packages/html/ds/src/button/button.stories.ts
```diff
@@ -22,6 +23,10 @@ const meta = {
 export default meta;
 type Story = StoryObj<typeof meta>;
 
+function sleep(ms: number) {
```
BogdanDanilescu, line 26: are we using this function anywhere ?

packages/html/ds/src/tabs/tabs.ts
```diff
@@ -0,0 +1,189 @@
...
+  tabElement: Element;
+  tabPanels: HTMLElement[];
+
+  constructor(tabElement: Element) {
+    this.tabElement = tabElement;
+
+    this.tabs = [];
+
+    this.tabs = [
+      ...this.tabElement.querySelectorAll('[role=tab]'),
```
BogdanDanilescu, line 22: Can we tighten the selector on this one ? something like selecting it by class and role, I'm just thinking that if there are other elements with role of `tab` on the page, it will have unexpected behaviour.

packages/html/ds/src/tabs/tabs.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        tab.addEventListener('keydown', this.onKeydown.bind(this));
+        tab.addEventListener('click', this.onClick.bind(this));
+      }
+    }
+
+    this.setSelectedTab(this.firstTab, false);
+  }
+
+  attachListeners() {
```
BogdanDanilescu, line 54: Can we keep the same naming conventions across the components ? So far we've been using `initComponent` and `destroyComponent`

apps/docs/src/components/document/get-started/developers-advice.tsx
pregno: nope, removing

packages/html/ds/src/button/button.stories.ts
pregno: nope, removing

packages/html/ds/src/tabs/tabs.ts
pregno: That `querySelectorAll` does queries within the scope of `tabElement`, so it does not reference any other tabs.

ramonmulia (approved):
LGTM

PR #300, opened 2024-10-23
feat: adding focus state

ramonmulia:
### PR for adding focus state for buttons, text input and nav components

add global CSS classes

### DIFF
--- a/packages/react/ds/src/button/button.stories.tsx
+++ b/packages/react/ds/src/button/button.stories.tsx
@@ -62,7 +62,7 @@ export const Default: Story = {
   },
 };
 
-export const WithIcon: Story = {
+export const WithIconLeft: Story = {
   args: {
     children: (
       <Fragment>
--- a/packages/react/ds/src/button/button.stories.tsx
+++ b/packages/react/ds/src/button/button.stories.tsx
@@ -62,6 +62,49 @@ export const Default: Story = {
   },
 };
 
+export const AllVariants: Story = {
+  args: {
+    children: 'Button',
+  },
+  render: () => (
+    <div className="gi-flex gi-flex-col gi-gap-4">
+      <div className="gi-flex gi-gap-4">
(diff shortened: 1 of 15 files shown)


BogdanDanilescu:
I think we still have some focus classes in the following files:

`file-upload.tsx`
`select.tsx`
`text-input.tsx`
`header-menu.tsx`
`file-upload.html`
`select.html`
`text-input.html`
`header-menu.html`

if you do a search with `focus:` in your editor you will see the components above are using generic tailwind classes instead of our custom components

packages/react/ds/src/button/button.stories.tsx
pregno, line 1: Could you add a story for each state?

pregno (commented):
added a question

packages/react/ds/src/button/button.stories.tsx
ramonmulia: I added a new story for html and react stories that covers all button variants.

BogdanDanilescu (approved):
LGTM

pregno (commented):
The focus state on the link create a layout shift, the text is moving on left by few pixels.
Please check with Lexa; on the link you should probably use box-shadow instead of border.

ramonmulia:
> The focus state on the link create a layout shift, the text is moving on left by few pixels. Please check with Lexa; on the link you should probably use box-shadow instead of border.

Good call, I replaced for box-shadow.
Focus and focus-visible states should be good now.

pregno:
We are getting closer :)
I've noticed that the new focus on inputs (checkboxes, text-input, textarea) doesn't have rounded corners.

ramonmulia:
> We are getting closer :) I've noticed that the new focus on inputs (checkboxes, text-input, textarea) doesn't have rounded corners.

This is intentional, I talked with Lexa she asked to not put them rounded.

pregno (approved):

PR #303, opened 2024-10-28
feat: chip component

ramonmulia:
### PR for adding chip component

- add global CSS classes
- react: component and storybook
- html: raw html and storybook
- docs updated

Screenshot 2024-10-29 at 06 46 58

Screenshot 2024-10-29 at 06 47 22


hamza14khan (approved):
Looks good. Can you also please add any screenshots or GIF from storybook or doc site? Thanks!

packages/design/tailwind/css/components.css
```diff
@@ -442,3 +442,9 @@ input[type='file' i] {
   @apply gi-mt-2;
 }
 /* TextArea */
+
+/* Chip */
+.gi-chip {
+  @apply gi-flex gi-items-center gi-gap-1 gi-pl-1.5 gi-pr-1 gi-py-0.5 gi-bg-gray-100 gi-w-fit gi-rounded-sm gi-cursor-default gi-text-xs hover:gi-bg-gray-200 focus:gi-bg-gray-200 focus:gi-outline focus:gi-outline-gray-950 focus:gi-outline-[2px] focus:gi-shadow-[inset_0_0_0_1px_white,0_0_0_5px_var(--gieds-color-yellow-400)];
```
BogdanDanilescu, line 448: Is this a different focus style than we have in the focus classes ? If it's not, can we use the focus classes instead ?

packages/react/ds/src/chip/chip.tsx
```diff
@@ -0,0 +1,42 @@
...
+    // should close if Enter or Space
+    if (['Enter', ' '].includes(event.key)) {
+      event.preventDefault();
+      onClose(event);
+    }
+  };
+
+  return (
+    <div
+      className={'gi-chip'}
```
BogdanDanilescu, line 28: I don't think we need the brackets here since we're just passing a string

packages/design/tailwind/css/components.css
ramonmulia: I had to use different focus styles for the Chip, so I couldn't use the classes created.

BogdanDanilescu (approved):

PR #307, opened 2024-10-30
chore: adding unit tests for React components 

ramonmulia:
### Adding unit tests for React components
- structure for unit tests in the React package.
- Includes AXE tests
- minor adjustments in some components to support testing (e.g., added `data-testid` attributes and exporting component prop types).


packages/react/ds/src/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+  it('should render hint text if provided', () => {
+    const screen = renderTextArea({
+      hint: {
+        text: 'This is a hint',
+      },
+    });
+    const hintElement = screen.getByText('This is a hint');
+
+    expect(hintElement).toBeTruthy();
+    expect(hintElement.tagName).toBe('DIV'); // Assuming hint is rendered as a <div> tag
```
ramonmulia, line 59: Changed to `.toBe('DIV');` to satisfies the test. 
The html test expect `.toBe('P');`

packages/react/ds/src/text-input/text-input.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  it('should render hint text if provided', () => {
+    const screen = renderTextInput({
+      hint: {
+        text: 'This is a hint',
+      },
+    });
+    const hintElement = screen.getByText('This is a hint');
+
+    expect(hintElement).toBeTruthy();
+    expect(hintElement.tagName).toBe('DIV'); // Assuming hint is rendered as a <div> tag
```
ramonmulia, line 70: Changed to `.toBe('DIV');` to satisfies the test. 
The html test expect `.toBe('P');`

packages/react/ds/src/label/label.tsx
```diff
@@ -8,7 +8,7 @@ export enum LabelSize {
 
 // Extend `React.LabelHTMLAttributes<HTMLLabelElement>` for correct label attributes
 export type LabelProps = React.LabelHTMLAttributes<HTMLLabelElement> & {
-  text: string;
+  text: ReactNode;
```
ramonmulia, line 11: Change this to accept things like: <strong>Bold text</strong> like macro does.

packages/react/ds/src/hint-text/hint-text.tsx
```diff
@@ -9,7 +9,7 @@ export enum HintSize {
 // Extend `React.InputHTMLAttributes<HTMLInputElement>` so that
 // the component can accept all the standard attributes and events that an `<input>` element can handle.
 export type HintTextProps = React.HTMLAttributes<HTMLInputElement> & {
-  text: string;
+  text: ReactNode;
```
ramonmulia, line 12: Change this to accept things like: `<strong>Bold text</strong>` like the macro does.

packages/react/ds/src/error-text/error-text.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    render(<ErrorText {...props} />);
+
+  it('should render error text with the correct content', () => {
+    const screen = renderErrorText({
+      text: 'This is an error message',
+      size: ErrorSize.md,
+    });
+    const errorElement = screen.getByText('This is an error message');
+    expect(errorElement).toBeTruthy();
+    expect(errorElement.tagName).toBe('DIV');
```
ramonmulia, line 15: Changed to `.toBe('DIV');` to satisfies the test. 
The html test expect `.toBe('P');`

packages/react/ds/src/textarea/textarea.test.tsx
BogdanDanilescu: instead of changing the test will it be better if we change the structure of the jsx ? As you pointed earlier, the UI will not change, it's just changing the HTML tag

packages/react/ds/src/label/label.tsx
BogdanDanilescu: Not sure if we want to do this, it will give the user the ability to add any components which will result in unexpected results

packages/react/ds/src/textarea/textarea.test.tsx
ramonmulia: yes, I will update those

hamza14khan (approved):
Looks great! Should we update the github workflow as well? @pregno

packages/react/ds/src/label/label.tsx
ramonmulia: agreed, changing back to string

pregno: yes, I was thinking the same.

pregno (approved):

BogdanDanilescu (approved):

PR #311, opened 2024-11-04
feat: use variants for heading

matteolc:
This PR adds a proposal change to the Heading component in order to use [tailwind-variants](https://www.tailwind-variants.org/docs/comparison).

Notes:
- We could introduce helper types if we get tired of typing `keyof typeof xxxVariants.variants.yyy`
- We could introduce helper consts if we want an enum like representation of the type:

```
const createEnum = <T extends string>(values: T[]): { [K in T]: K } => {
  const obj = {} as { [K in T]: K };
  for (const value of values) {
    obj[value] = value;
  }
  return obj;
};

type HeadingSize = keyof typeof headingVariants.variants.size

const HeadingSizes = createEnum(Object.keys(HeadingSize))
```


packages/react/ds/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@govie-ds/react",
-  "version": "0.20.1",
+  "version": "0.20.2",
```
pregno, line 3: Please don't update the version here. We have the release pipeline that will update it automatically.

pregno (changes requested):

packages/react/ds/package.json
matteolc: https://github.com/ogcio/govie-ds/commit/e970909744cf37e7c46af7bc794d27aaa1b7bfaf

packages/react/ds/src/heading/heading.stories.tsx
```diff
@@ -14,3 +14,16 @@ export const Default: Story = {
...
+export const WithSize: Story = {
+  render: () => (
+    <div className="flex flex-row gap-2">
+      {Object.values(HeadingSize).map((size) => (
+        <Heading key={size} size={size}>
+          Heading {size}
+        </Heading>
+      ))}
+    </div>
+  ),
```
BogdanDanilescu, line 28: in the Jinja/Nunjucks library, we have a story for each size variant. Keeping in mind that we looking to achieve consistency between the two libraries when it comes to storybook, are we going with one story that renders all sizes like the one above or do we want to have a story for each size ? WDYT ?

pregno: Can we achieve the same also with the macro?

BogdanDanilescu: I don't believe that we can. Since we have a `.ts` extension for the macros we cannot use the tsx features. Of course we can try to change the extension from `.ts` to `.tsx` but we might get some unexpected results if we do that since they are not React components

ramonmulia: the render function in Jinja should return a html string, we have an example in html/ds/button/button.stories.ts (AllVariants) story.

BogdanDanilescu (approved):

packages/react/ds/src/heading/heading.tsx
```diff
@@ -1,79 +1,51 @@
-export type HeadingAs = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
-export type HeadingSize = 'xl' | 'lg' | 'md' | 'sm' | 'xs' | '2xs';
-export type HeadingProps = {
+import { tv, type VariantProps } from 'tailwind-variants';
+import { cn } from '../cn.js';
+
+const headingVariants = tv({
```
pregno, line 4: Could you add a test with the following properties:
```
size: 'xl',
as: 'h6`
```
That should render an `h6` tag with an `xl` style.
I'm concerned about how it prioritizes the styles.

pregno (commented):
@matteolc Added a comment and there are a couple of conflicts.

packages/react/ds/src/heading/heading.tsx
matteolc: https://github.com/ogcio/govie-ds/commit/e0c0288d4009dcf5b22f3df88abd7257c69657d7

packages/react/ds/src/heading/heading.stories.tsx
matteolc: Resolving this since we removed enums in favour of non-breaking change with tailwind-variants

pregno (approved):

PR #320, opened 2024-11-07
feat: alert component

BogdanDanilescu:
### PR to implement `Alert` for HTML and React

_Props:_
`variant`
`title`
`children`
`className`
`dismissible`

**SB screenshots:**
Screenshot 2024-11-12 at 15 31 01
Screenshot 2024-11-12 at 15 31 13


packages/react/ds/src/alert/alert.tsx
```diff
@@ -0,0 +1,81 @@
...
+function Alert({ title, children, variant = 'info', className }: AlertProps) {
+  const { base, heading, container, content } = alertVariants({
+    variant,
+  });
+
+  return (
+    <div className={cn(base(), className)} role="alert">
+      <div className={cn(container())}>
+        <Icon size="lg" icon={icon({ variant })} />
+        <Heading size="sm" as="h2" customClasses={heading()}>
```
pregno, line 71: Having a heading here can possibly affect the content hierarchy.
I suggest using a div with the same class as the h2.
Be careful about the margins that heading is introducing; we probably need different spacing here.

packages/react/ds/src/alert/alert.tsx
```diff
@@ -0,0 +1,81 @@
...
+  children?: ReactNode;
+  className?: string;
+};
+
+const icon = ({ variant }: VariantProps<typeof alertVariants>) => {
+  let icon;
+  switch (variant) {
+    case 'danger':
+    case 'warning': {
+      icon = 'error';
```
pregno, line 48: use `warning` icon for warnings

pregno (changes requested):
Same comments apply to html version

packages/react/ds/src/icon/icon.tsx
```diff
@@ -16,7 +16,10 @@ export type IconId =
   | 'home'
   | 'logout'
   | 'download'
-  | 'keyboard_arrow_down';
+  | 'keyboard_arrow_down'
+  | 'info'
+  | 'check_circle'
+  | 'error';
```
pregno, line 22: Add `warning` here

packages/react/ds/src/alert/alert.tsx
```diff
@@ -0,0 +1,103 @@
...
+  defaultVariants: {
+    variant: 'info',
+  },
+});
+
+type AlertProps = {
+  variant?: VariantProps<typeof alertVariants>['variant'];
+  title: string;
+  children?: ReactNode;
+  className?: string;
```
pregno, line 40: I know that in the past we added `className` but since we decided to remove any possible customizations from the users, we should remove it.

pregno (changes requested):

examples/vite/src/application.tsx
```diff
@@ -186,8 +187,8 @@ export function App() {
           title={{
             value: "Where do you live?",
             asHeading: {
-              size: 'md',
-              as: 'h3',
+              size: "md",
```
ramonmulia, line 190: I believe our lint configs aren’t aligned, it would be helpful to ensure we’re all using the same linting configuration to avoid these small unintentional changes across the files.

packages/react/ds/src/alert/alert.tsx
ramonmulia: maybe keeping it would be helpfull to build the Toast?

packages/react/ds/src/alert/alert.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return (
+    <div className={cn(base(), className)} role="alert">
+      <Icon icon={icon({ variant })} />
+      <div className={container()}>
+        <p className={heading()}>{title}</p>
+        {children}
+      </div>
+      {dismissible && (
+        <IconButton
+          onClick={() => setIsDismissed(true)}
```
ramonmulia, line 91: maybe adding an "onClick" handler as prop for when the alert is closed it would allow the parent component to detect the closure and perform some action

BogdanDanilescu: Don't know about adding the `onClick` as a prop in this case since the action is intended just for closing the `Alert`. Not sure I see a different use case other than that. Thoughts @pregno @hamza14khan  ?

examples/vite/src/application.tsx
BogdanDanilescu: The issue here was that we didn't have the prettier config set up for the vite example, I added that now and it should apply the same rules for all of us

packages/react/ds/src/alert/alert.tsx
BogdanDanilescu: Just had a conversation with @pregno earlier and the suggestion was that we should remove the className prop across all components so that users will have limited flexibility when it comes to styling as it might affect accessibility standards. The ticket to tackle that is [here](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/23203)

ramonmulia (approved):
LGTM

packages/react/ds/src/alert/alert.tsx
pregno: Good point. 
We can have an optional callback props `onClose` (or `onDismiss`) that is triggered when the user press the `X` close button.

matteolc (approved):
LGTM

packages/react/ds/src/alert/alert.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+  },
+  defaultVariants: {
+    variant: 'info',
+  },
+});
+
+type AlertProps = {
+  variant?: VariantProps<typeof alertVariants>['variant'];
+  title: string;
```
matteolc, line 42: We could make the title an optional prop, sometime we may not need it?

image

packages/react/ds/src/alert/alert.tsx
BogdanDanilescu: Added

PR #323, opened 2024-11-11
feat: add react pagination component

hamza14khan:
Addition of:

- Pagination component. It is responsive when screen size is below or equal to 480px.
- Addition of useBreakpoint hook.
- Associated documentation update.
- Relevant tests.
- Update `Button` component with disabled attribute for accessibility.

Testing:
Screenshot 2024-11-11 at 16 38 03
Screenshot 2024-11-11 at 16 37 50


packages/react/ds/src/pagination/pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+        variant="flat"
+        size="medium"
+        appearance="dark"
+        disabled={currentPage === 1}
+        onClick={() => onPageChange(currentPage - 1)}
+      >
+        <React.Fragment key="previous-btn-pagination">
+          <Icon icon="arrow_left_alt" />
+        </React.Fragment>
+        {!isResponsive && previousLabel}
```
pregno, line 109: I suggest to use `gi-hidden sm:gi-block`

packages/react/ds/src/pagination/pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </Button>
+
+      {isResponsive ? renderPaginationLabel() : renderPaginationBtns()}
```
pregno, line 112: Here you can have both rendered and use the classes for hide/show the correct one

pregno (dismissed):
I think we can use CSS without the need of the `useBreakpoint` hooks.
Especially if we want to achieve the same results on HTML.

pregno (commented):
Screenshot 2024-11-12 at 14 48 04
This should be centered, am I right?

BogdanDanilescu (approved):

ramonmulia (approved):
LGTM, I just missed an example in next app.

pregno (approved):

PR #324, opened 2024-11-12
fix: radio feedback fixes

BogdanDanilescu:
Implementation of the following:

- Expose the `Radio` Component
- Add the following props:
1.   `checked` -> if `true` the input is checked by default
2.  `onChange` --> callback function
- Rename the prop `fieldId` to `groupId`
- Add stories for the `Radio` Component

The above applies to the Macro and React libraries

**SB screenshots**
Screenshot 2024-11-12 at 10 02 22
1d33f26)
Screenshot 2024-11-12 at 10 02 54

### DIFF
--- a/packages/react/ds/src/radio/radio.tsx
+++ b/packages/react/ds/src/radio/radio.tsx
@@ -41,12 +41,13 @@ export const Radio = ({
   label,
   value,
   hint,
-  radioId,
+  id,
   size,
   conditionalInput,
   checked,
   onChange,
 }: RadioProps) => {
+  const radioId = id ?? value;
   return (
     <div className="gi-radio-container">
       <div className="gi-radio-input-container">
@@ -66,7 +67,7 @@ export const Radio = ({
       {(hint || conditionalInput) && (
         <div className="gi-radio-conditional-divider-container">
           <div
-            className={`${addConditionalDivider(checked, conditionalInput)} ${getRadioWidth(size)}`}
+            className={`${addConditionalDivider((checked = false), conditionalInput)} ${getRadioWidth(size)}`}
           >
             <div
               className={`gi-radio-conditional-divider-border-container ${getRadioWidth(size)}`}
--- a/packages/react/ds/src/radio/radio.tsx
+++ b/packages/react/ds/src/radio/radio.tsx
@@ -1,8 +1,8 @@
 'use client';
 
 import { HintText } from '../hint-text/hint-text.js';
-import { TextInput, TextInputProps } from '../text-input/text-input.js';
-import { RadioProps, RadiosSizeEnum } from './types.js';
+import { TextInput, type TextInputProps } from '../text-input/text-input.js';
+import { type RadioProps, RadiosSizeEnum } from './types.js';
 
 const getRadioSize = (size?: RadiosSizeEnum) => {
   let sizeClass = 'gi-radio-medium';
@@ -52,6 +52,7 @@ export const Radio = ({
     <div className="gi-radio-container">
       <div className="gi-radio-input-container">
         <input
+          readOnly={checked}
           onChange={onChange}
           checked={checked}
           name={name}
@@ -73,7 +74,7 @@ export const Radio = ({
             <div
               className={`gi-radio-conditional-divider-border-container ${getRadioWidth(size)}`}
             >
-              <div className="gi-radio-conditional-divider-border"></div>
+              <div className="gi-radio-conditional-divider-border" />
             </div>
           </div>
           <div>
--- a/packages/react/ds/src/radio/radio.tsx
+++ b/packages/react/ds/src/radio/radio.tsx
@@ -59,6 +59,7 @@ export const Radio = ({
           value={value}
           className={`gi-radio-base ${getRadioSize(size)}`}
           type="radio"
+          aria-label={radioId}
         />
         <label htmlFor={radioId} className="gi-radio-label">
           {label}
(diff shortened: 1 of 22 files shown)


packages/react/ds/src/radio/radio.tsx
```diff
@@ -58,6 +59,7 @@ export const Radio = ({
           value={value}
           className={`gi-radio-base ${getRadioSize(size)}`}
           type="radio"
+          aria-label={radioId}
         />
```
matteolc, line 63: `checked` should be `defaultChecked`, else you'll have warnings all over the place when you try to change the value (about changing an uncontrolled/controlled component)

packages/react/ds/src/radio/radios-group.tsx
```diff
@@ -1,24 +1,34 @@
...
+    for (const radio of items) {
+      if (radio.checked) {
+        setValue(radio.value);
+      }
+    }
+  }, []);
+
   const onOptionChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     setValue(event.target.value);
+    onChange && onChange(event);
```
matteolc, line 31: `onChange?.(event);`

packages/react/ds/src/radio/radios-group.tsx
```diff
@@ -59,11 +69,11 @@ export const RadiosGroup = ({
                 key={`${value}-${index}`}
```
matteolc, line 69: Please use self-closing elements when there are no children:

Nope: `<div className="classname"></div>`
Yep: `<div className="classname"/>`

matteolc: If possible please also avoid using the `index` of an array as key property in an element

packages/react/ds/src/radio/radios-group.tsx
```diff
@@ -1,24 +1,34 @@
...
   const [value, setValue] = useState<null | string>(null);
 
+  useEffect(() => {
+    for (const radio of items) {
+      if (radio.checked) {
+        setValue(radio.value);
+      }
+    }
+  }, []);
+
```
matteolc, line 28: If you don't specify all of `useEffect` dependencies I'd suggest adding a comment explaining why you can safely ignore

packages/react/ds/src/radio/radio.tsx
```diff
@@ -66,7 +68,7 @@ export const Radio = ({
       {(hint || conditionalInput) && (
         <div className="gi-radio-conditional-divider-container">
           <div
-            className={`${addConditionalDivider(checked, conditionalInput)} ${getRadioWidth(size)}`}
+            className={`${addConditionalDivider(conditionalInput, checked)} ${getRadioWidth(size)}`}
           >
             <div
               className={`gi-radio-conditional-divider-border-container ${getRadioWidth(size)}`}
```
matteolc, line 74: `<div className="gi-radio-conditional-divider-border"></div>`
`<div className="gi-radio-conditional-divider-border"/>`

packages/react/ds/src/radio/radio.tsx
```diff
@@ -27,10 +27,10 @@ export const getRadioWidth = (size?: RadiosSizeEnum) => {
 };
```
matteolc, line 27: Please use `import type` or `import { type ..., }` if your imports are only used as types

packages/react/ds/src/radio/radio.tsx
matteolc: image

packages/react/ds/src/radio/radios-group.tsx
matteolc: I am not sure why we need this effect? I feel this is a bit confusing..
Could I give the group a list of radios that can have their (default) value checked (say by mistake)? What happens if I do that? Does the last one checked in the list win? Will I see it checked?

matteolc: I would expect the default value of a radio group to be controlled at radio group level, not for each radio?

packages/react/ds/src/radio/radio.tsx
BogdanDanilescu: Changing the prop name resulted in the same warning. As a solution, I tied the `readonly` attribute value to match `checked` attribute value thus ensuring that if the `Radio` has a `checked` prop it will only be a readonly

packages/react/ds/src/radio/radios-group.tsx
BogdanDanilescu: Fixed

packages/react/ds/src/radio/radios-group.tsx
BogdanDanilescu: Fixed. Jusy FYI I kept the index as a key because it is used in conjunction with `value` and the `index` won't change in this particular case since we are not mutating the array

packages/react/ds/src/radio/radios-group.tsx
BogdanDanilescu: Good point @matteolc . As an alternative, I removed the `useEffect` and also removed the `checked` property from the items array and instead added a `defaultValue` to the group level. If `defaultValue` matches any of the values inside the items array then that will be the default checked radio. Does that sound like a better solution ?

packages/react/ds/src/radio/radio.tsx
BogdanDanilescu: fixed

packages/react/ds/src/radio/radio.tsx
BogdanDanilescu: fixed.

matteolc (approved):
Nice work 👍🏽

pregno (approved):

ramonmulia (approved):

PR #328, opened 2024-11-14
feat: stack component

ramonmulia:
### Adding stack component
- adding stack component in react and html
- adding tests
- docs and storybook updated

Screenshot 2024-11-14 at 13 58 33
Screenshot 2024-11-19 at 09 57 33


packages/react/ds/src/stack/stack.tsx
BogdanDanilescu, line 1: maybe this is a good use case for us to integrate the `tailwind-variants` package ?

BogdanDanilescu:
I haven't looked closely if the `Stack` component applies to any of our React components but if we can integrate it in our existing components maybe it will be a good idea ? an example would be the links on the `Footer`

apps/docs/content/3-components/2-library/35-stack.mdx
```diff
@@ -0,0 +1,62 @@
...
+        gap={{ base: 5, md: 3 }}
+        fixedHeight="200px"
+        >
+        <div className="gi-bg-gray-300 gi-p-2">Item 1</div>
+        <div className="gi-bg-gray-300 gi-p-2">Item 2</div>
+        <div className="gi-bg-gray-300 gi-p-2">Item 3</div>
+    </Stack>;
+
+```
+</TabPanel>
```
hamza14khan, line 51: minor: formatting issue.

hamza14khan (approved):
Looks good!

packages/react/ds/src/stack/stack.tsx
ramonmulia: After diving into the tailwind-variants I realized that the component actually doesn't need variant in tv, cause it is not creating predefined groups of class styles based on multiple options, instead it is handling responsive breakpoints for props (like `direction` and `gap`) dynamically.

BogdanDanilescu (approved):

matteolc (approved):
❤️ 
I'd maybe only suggest to move type declarations to a dedicated file and then importing them in the implementation file with `import type ..`
This helps with separation of concern and also saves the compiler a few bytes 😄 
Also having a story for the component with its default props usually helps 😜

PR #332, opened 2024-11-20
[In Progress] feat: html pagination component

hamza14khan:


packages/html/ds/src/pagination/pagination.util.ts
hamza14khan, line 1: I have a similar file for react as well. Do you guys have any suggestion to share the common utils across mono repos?

packages/react/ds/src/pagination/pagination.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import React from 'react';
 import { Button } from '../button/button.js';
-import { Breakpoint, useBreakpoint } from '../hooks/use-breakpoint.js';
```
hamza14khan, line 4: Unused import.

packages/html/ds/src/pagination/pagination.util.ts
BogdanDanilescu: one way would be to expose it in the react package and then use it like so `import {...} from @govie-ds/react`

packages/html/ds/src/button/button.html
hamza14khan, line 1: @BogdanDanilescu

packages/html/ds/src/pagination/pagination.html
```diff
@@ -0,0 +1,158 @@
...
+  <div
+    class="gi-pagination"
+    data-current-page="{{ props.currentPage }}"
+    data-total-pages="{{ props.totalPages }}"
+    data-testid="govie-pagination"
+    data-module="gieds-pagination"
+  >
+    {{
+      govieButton({
+        "content": "<span data-testid=\"govie-icon\" role=\"presentation\" class=\"material-symbols-outlined gi-block gi-text-[24px]\">arrow_left_alt</span> <span class='sm:gi-block gi-hidden'>Previous</span>",
```
BogdanDanilescu, line 13: I'm not sure if this will render an error but can we use the `govieIcon` macro here ? Otherwise if we change something for this component, the changes will not reflect here as we manually add the rendered html

BogdanDanilescu:
can you add an example in the wagtail app please ?

ramonmulia (approved):
LGTM

packages/html/ds/src/pagination/pagination.html
```diff
@@ -0,0 +1,158 @@
...
+        "disabled": props.currentPage == 1,
+        "href": "#",
+        "target": "_self",
+        "testId": "govie-pagination-prev-btn",
+        "data-current-page": props.currentPage - 1
+      })
+    }}
+
+    {# Compact View #}
+    {{ goviePagintationCompactView(props) }}
```
pregno, line 26:
```suggestion
    {{ goviePaginationCompactView(props) }}
```

packages/html/ds/src/pagination/pagination.html
```diff
@@ -0,0 +1,158 @@
...
+        "target": "_self",
+        "disabled": props.currentPage == props.totalPages,
+        "testId": "govie-pagination-next-btn",
+        "data-current-page": props.currentPage + 1
+      })
+    }}
+  </div>
+{% endmacro %}
+
+{% macro goviePagintationCompactView(props) %}
```
pregno, line 50:
```suggestion
{% macro goviePaginationCompactView(props) %}
```

pregno (changes requested):
there is a typo

pregno (approved):

PR #334, opened 2024-11-21
feat: breadcrumbs component

ramonmulia:
PR to implement Breadcrumbs component

added:

- stories
- tests
- macro
- docs
- example in nextjs and wagtail

Screenshot 2024-11-25 at 13 41 41


apps/docs/content/3-components/2-library/4-breadcrumbs.mdx
```diff
@@ -9,6 +9,61 @@ status: in-development
 
 <ComponentStatusBlock componentId="breadcrumbs" />
 
+<Breadcrumbs>
+  <BreadcrumbLink href="/home">Home</BreadcrumbLink>
+  <BreadcrumbEllipses />
```
BogdanDanilescu, line 14: I think there's a typo. I think it should be `BreadCrumbEllipsis`

pregno (approved):

BogdanDanilescu:
can you add an example in the wagtail application ?

packages/react/ds/src/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,43 @@
+import { Link } from '../link/link.js';
+import type { BreadcrumbLinkProps, BreadcrumbProps } from './types.js';
+
+export const BreadcrumbEllipsis = () => (
+  <div className="gi-breadcrumb-ellipsis" aria-hidden="true">
```
hamza14khan, line 5: suggestion: Not sure if you have consult this with Lexa or not, but we have an icon for this.

```
<span data-testid="govie-icon" class="material-symbols-outlined gi-gray-700">more_horiz</span>
```

ramonmulia: hmm I didn't know that, let me check it.

packages/html/ds/src/breadcrumbs/breadcrumbs.html
```diff
@@ -0,0 +1,30 @@
+<nav aria-label="Breadcrumbs" class="gi-breadcrumbs">
```
pregno, line 1: Could you add a macro to create this structure?
I think we can accept as a param a list of items like this:
```
{
   "label": string;
   "href": string;
   "ellipsis": boolean;
}
```
So this example will be achieved using:
```
{{
   govieBreadcrumbs([
      {
         "label": "Home",
         "href": "/"
      },
      {
         "ellipsis": true
      },
      {
         "label": "Travel",
         "href": "/travel"
      },
      {
         "label": "Documentation",
         "href": "/travel/docs"
      }
   ])
}}
```

pregno (changes requested):

packages/react/ds/src/breadcrumbs/breadcrumbs.tsx
hamza14khan: I have used in my html pagination PR if you want a reference.

hamza14khan (approved):

BogdanDanilescu (approved):

pregno (approved):

PR #337, opened 2024-11-25
fix: pagination consistent width across html and react platforms

hamza14khan:
- Pagination (react and html): usage icon classes on next and previous buttons if no label is shown (e.g. under 640px - small screens).
- Extended class names on button macro.
- Documentation update.

Testing:
Screenshot 2024-11-25 at 22 40 31

### DIFF
--- a/apps/docs/dictionary.txt
+++ b/apps/docs/dictionary.txt
@@ -121,12 +121,9 @@ itemsAlignment
 itemsDistribution
 fixedHeight
 groupId
-<<<<<<< HEAD
 hideToc
 Dismissible
-=======
 gi-pagination-next-btn
 gi-btn
 gi-btn-flat-dark
 gi-btn-large
->>>>>>> a6f72ab7 (fix: dictionary update)
--- a/apps/docs/dictionary.txt
+++ b/apps/docs/dictionary.txt
@@ -121,5 +121,12 @@ itemsAlignment
 itemsDistribution
 fixedHeight
 groupId
+<<<<<<< HEAD
 hideToc
 Dismissible
+=======
+gi-pagination-next-btn
+gi-btn
+gi-btn-flat-dark
+gi-btn-large
+>>>>>>> a6f72ab7 (fix: dictionary update)
--- a/packages/design/tailwind/css/components.css
+++ b/packages/design/tailwind/css/components.css
@@ -904,10 +904,21 @@ input[type='file' i] {
 }
 
 /* End of Toast */
+
 /* Pagination */
 
 .gi-pagination {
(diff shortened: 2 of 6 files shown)


packages/react/ds/src/pagination/pagination.tsx
```diff
@@ -18,8 +18,9 @@ export const Pagination: React.FC<PaginationProps> = ({
   onPageChange,
 }) => {
   const { breakpoint, width } = useBreakpoint();
-  const isCompactView = breakpoint === Breakpoint.XS; // Custom breakpoint for compact view.
-  const showLabel = width < 639;
+  const isCompactView = breakpoint === Breakpoint.XS;
+  const isSMWidth = width < 639;
+  const showLabel = isSMWidth;
```
pregno, line 23: If `showLabel` and `isSMWidth` are the same, why having 2 const?

apps/docs/dictionary.txt
```diff
@@ -122,4 +122,8 @@ itemsDistribution
 fixedHeight
 groupId
 hideToc
-Dismissible
\ No newline at end of file
+Dismissible
+gi-pagination-next-btn
```
hamza14khan, line 126: Do you guys know why do I have to add classes in dicitonary.txt? 🤔 

@pregno @ramonmulia @BogdanDanilescu

pregno: Because we are adding the `html` in the mdx file.

hamza14khan: It never complained on other HTML files using classes?

ramonmulia (approved):

packages/design/tailwind/css/components.css
```diff
@@ -904,10 +904,21 @@ input[type='file' i] {
...
 /* End of Toast */
+
 /* Pagination */
 
 .gi-pagination {
-  @apply gi-flex xs:gi-justify-center gi-justify-between gi-items-center  gi-gap-2;
+  @apply gi-flex xs:gi-justify-center gi-justify-between gi-items-center gi-gap-2;
 }
 
+@media (max-width: 640px) {
```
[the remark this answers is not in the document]
hamza14khan: It throws me this error:

The `max-sm:gi-p-3` class does not exist. If `max-sm:gi-p-3` is a custom class, make sure it is defined within a `@layer` directive.

When using this:

```
.gi-pagination-next-btn,
.gi-pagination-prev-btn {
  @apply max-sm:gi-p-3 max-sm:gi-h-12 max-sm:gi-w-12 max-sm:gi-justify-center;
}
```

apps/docs/dictionary.txt
pregno: It's checking only mdx files

pregno (approved):

PR #339, opened 2024-11-26
feat: summary list component

ramonmulia:
PR to implement Summary List component

added:

- stories
- tests
- macro
- docs
- example in nextjs and wagtail

Screenshot 2024-11-26 at 13 45 53

Screenshot 2024-11-26 at 13 45 39


apps/docs/content/3-components/2-library/summary-list.mdx
```diff
@@ -7,24 +7,131 @@ libraries:
...
+          <SummaryListRow label="Name">
+            <SummaryListValue>John Smith</SummaryListValue>
+            <SummaryListAction href="/change-name">Change name</SummaryListAction>
+          </SummaryListRow>
+          <SummaryListRow label="Date of Birth">
+              <SummaryListValue>8 November 1982</SummaryListValue>
+              <SummaryListAction href="/change-dob">Change date of birth</SummaryListAction>
+          </SummaryListRow>
+          <SummaryListRow label="Address">
+            <SummaryListValue>
```
hamza14khan, line 107: @BogdanDanilescu @pregno 

Do you guys have any alternative ways of passing an array or object to the summary list value? I haven't seen this pattern tbh to pass JS as children 🤔

hamza14khan: Maybe this can be more convenient for the developer.

```
<SummaryListValue values={['72 Guild Street', 'London', 'SE23 6FH']} />
```

ramonmulia: good call, I'm good either way tbh, I was just trying to keep ```children``` prop following the pattern from the others summary list children components.
Let me know your toughts!

BogdanDanilescu: Shouldn't the `<SummaryListValue />` accept strings ? Is there a reason why we are passing an array of strings here ?

BogdanDanilescu: or alternatively, if we could do something like this:
```
<SummaryListValue>
  <Paragraph>72 Guild Street</Paragraph>
  <Paragraph>London</Paragraph>
  <Paragraph>SE23 6FH</Paragraph>
</SummaryListValue />
```

ramonmulia: I'm following the Local version, so there they are breaking lines ```<br/>``` so I ended up to having an array of strings.
So intead of adding a Paragraph that will include non desired styles, what about to have something like this:

```
<SummaryListValue>
   72 Guild Street
   <br/>
   London
   <br/>
   SE23 6FH
 <SummaryListValue>
 ```
 
 @BogdanDanilescu

BogdanDanilescu: Yes, this works too. But just thinking about what the other devs were saying about giving them the ability to compose / add what they need depending on their use case. Having that in mind, would it be a good idea to have a `<Slot />` in the ``<SummaryListValue>`` and so it will be up to the developer what they want to render as a value. They can either do something similar to your example with the `<br />` or they can just do a mapping of paragraphs, whatever they think it will work best in their situation

ramonmulia: agree, this will bring more flexibility.

BogdanDanilescu (approved):

PR #343, opened 2024-11-28
fix: replace push method with append

BogdanDanilescu:
- Replaced the `push` method with `append`. 
- Used a `temp` variable to mutate the array as the `append` method works in place and it would have return a `None` value


ramonmulia (approved):

hamza14khan (approved):
Great work!

Can you please write a comment that says something about `temp` variable? 😄

PR #350, opened 2024-12-02
feat: slot items for Header component (react)

ramonmulia:
Add Slot items for the Header component

- header slot feature
- tests
- examples how it could be used (storybook)
- breakpoints fixed (mobile and tablet)

Screenshot 2024-12-04 at 07 42 41

Screenshot 2024-12-04 at 07 43 17


packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -16,19 +19,97 @@ type HeaderMenuProps = {
...
+  item: { label?: string; slot: React.ReactNode };
+};
+
+const MenuListItem = ({ href = '#', label = '', bold = true }) => (
+  <a
+    aria-label={label || 'link with no label'}
+    href={href}
+    className="hover:gi-underline hover:gi-underline-offset-sm gi-block gi-py-4 gi-border-gray-100 gi-border-b-xs gi-border-solid"
+  >
+    <span className={cn('gi-text-sm', 'gi-ml-1', { 'gi-font-bold': bold })}>
```
BogdanDanilescu, line 35: do we need `{ 'gi-font-bold': bold }` ? can't we do just `gi-font-bold` instead ?

packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -16,19 +19,97 @@ type HeaderMenuProps = {
...
     icon?: IconId;
   };
+} & Pick<HeaderProps, 'tools'>;
+
+type MenuItemAccordionProps = {
+  index: number;
+  item: { label?: string; slot: React.ReactNode };
+};
+
+const MenuListItem = ({ href = '#', label = '', bold = true }) => (
```
BogdanDanilescu, line 29: can we add types to this function ?

packages/react/ds/src/header/components/header-slot.tsx
```diff
@@ -0,0 +1,78 @@
+import { useEffect } from 'react';
+import ReactDOMClient from 'react-dom/client';
+import { Icon } from '../../icon/icon.js';
+
+type SloTProps = {
```
BogdanDanilescu, line 5: the letter 'T' should be lowercase

packages/react/ds/src/header/components/header-slot.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return null;
+};
+
+const SlotItemAction = ({ item: { label, icon }, index }: SloTProps) => {
+  return (
+    <label
+      htmlFor={`ItemActionTrigger-${index}`}
+      className="gi-header-tool-item"
+    >
+      <input
```
BogdanDanilescu, line 53: is it a good idea to use the `Checkbox` component here ?

packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -16,19 +19,107 @@ type HeaderMenuProps = {
...
+    href={href}
+    className="hover:gi-underline hover:gi-underline-offset-sm gi-block gi-py-4 gi-border-gray-100 gi-border-b-xs gi-border-solid"
+  >
+    <span className={cn('gi-text-sm', 'gi-ml-1', { 'gi-font-bold': bold })}>
+      {label}
+    </span>
+  </a>
+);
+
+export const MenuItemAccordion = ({ index, item }: MenuItemAccordionProps) => {
```
pregno, line 51: Why an `accordion`?

packages/react/ds/src/header/components/header-search.tsx
```diff
@@ -14,28 +14,23 @@ function HeaderSearch({
...
-            className="gi-flex-auto"
-          />
-          <Button className="gi-ml-1 gi-flex-none">
-            Search
-            <Icon icon={icon} ariaLabel="Search" />
-          </Button>
-        </div>
-      </form>
-    </div>
+    <form action={ActionType} className="gi-max-w-md gi-mx-auto">
```
pregno, line 17: There is probably something wrong with the merge.
The current version of `main` has the correct styling.

packages/react/ds/src/stack/stack.tsx
```diff
@@ -204,6 +204,7 @@ export const Stack: React.FC<StackProps> = ({
     'gi-flex',
     'gi-overflow-auto',
     'gi-w-full',
+    'gi-p-1',
```
pregno, line 207: Why this padding?

packages/react/ds/src/header/components/header-slot.tsx
pregno: If you are using the `checkbox` why do you need to attach the listener to the buttons with `attachEventsToItemActionTriggers`?

ramonmulia: I needed to attach events on the buttons to toggle the icons, hide all slots opened (including search) and show the current one.
As we have differents custom items slots it weren't possible to have css rules for that (like SearchTrigger), so attaching the event `change` will allow the all checkboxs close and open just the one clicked.

packages/react/ds/src/header/components/header-menu.tsx
ramonmulia: Afeter some discussions with Lexa we came out with this solution, just for mobile in this case: https://www.figma.com/design/ZEqfpXhocdZs1K5YfsE5ws/Design-system?node-id=3412-1644&node-type=canvas&m=dev

this way if the user wants, the user can include the prop `keepOnMobile` and if it is true we show the slots in mobile/tablet

packages/react/ds/src/stack/stack.tsx
ramonmulia: my bad, going to remove it.

packages/react/ds/src/header/components/header-search.tsx
ramonmulia: The styles are the same, I just moved the div container to the parent component, so in that case we show the right styles only on desktop.

pregno (approved):

packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -16,19 +19,107 @@ type HeaderMenuProps = {
...
+  return (
+    <div
+      key={`Accordion-item-${index}`}
+      id={`Accordion-item-${index}`}
+      className="gi-accordion-item-container"
+      data-open="false"
+    >
+      <div
+        aria-label={item.label}
+        className="gi-accordion-item-toggle hover:gi-no-underline hover:gi-underline-offset-sm gi-block gi-py-4 gi-border-gray-100 gi-border-b-xs gi-border-solid"
```
hamza14khan, line 92: question: Can we create a class for this?

packages/react/ds/src/header/components/header-menu.tsx
```diff
@@ -16,19 +19,107 @@ type HeaderMenuProps = {
...
+      >
+        <div className="gi-full-width gi-flex gi-justify-between">
+          <span className="gi-text-sm gi-font-bold gi-ml-1">{item.label}</span>
+          <Icon icon="chevron_right" className="gi-accordion-item-icon" />
+        </div>
+      </div>
+
+      <div className="gi-accordion-item-slot gi-hidden gi-bg-gray-50 gi-p-4">
```
hamza14khan, line 100: same as above.

BogdanDanilescu (approved):

PR #352, opened 2024-12-03
fix: remove !important classes

BogdanDanilescu:
PR to remove the important classes ( `!gi...` ) with the exception of the`notyf` package styles as we need to override the native classes that come from the package.


ramonmulia (commented):
I see some spacing changes in Chromatic, are they intentional?

BogdanDanilescu:
> I see some spacing changes in Chromatic, are they intentional?

No it's not intentional, I'll have a look

BogdanDanilescu:
@ramonmulia there are no more UI changes now

ramonmulia (approved):
LGTM

packages/html/ds/src/card/card.html
```diff
@@ -61,7 +57,7 @@
               }}
             </div>
           {% else %}
-            <div class="{{ titleClasses }}">{{ props.title }}</div>
+            <div class="gi-card-title">{{ props.title }}</div>
```
hamza14khan, line 60: praise: good catch!

hamza14khan (approved):

pregno (approved):

PR #354, opened 2024-12-05
fix: tailwind typography configuration

pregno:
Tailwind typography wasn't configured in the default theme.
It worked in the documentation website because it was configured in its tailwind setup.

This means that in a new setup if the user doesn't specify the typography plugin, the `prose` utils will not work.
Now, we can expose the `prose` class with the DS CSS.


BogdanDanilescu (approved):

apps/docs/src/app/[...slug]/page.tsx
```diff
@@ -43,7 +43,7 @@ export default function DocumentPage({ params }: DocumentPageProps) {
 
   return (
     <>
-      <div className="w-full flex-auto prose max-w-none">
+      <div className="w-full flex-auto gi-prose gi-max-w-none">
```
Comment (ramonmulia) [inline apps/docs/src/app/[...slug]/page.tsx:46]: replace these classes without gi prefix to  ```gi-w-full``` and  ```gi-flex-auto```

apps/docs/src/components/document/common/mdx.tsx
```diff
@@ -197,7 +197,7 @@ const documentComponents: MDXComponents = {
   ComponentContainer: (props) => (
     <div
       {...props}
-      className={cn('my-xl stroke-gray-950 not-prose', props.className)}
+      className={cn('my-xl stroke-gray-950 gi-not-prose', props.className)}
```
ramonmulia, line 200: replace these classes without gi prefix to  ```gi-my-xl``` and  ```gi-stroke-gray-950```

ramonmulia: hmm interesting, ok

ramonmulia (approved):

PR #356, opened 2024-12-05
feat: Progress Stepper component

ramonmulia:
#### Progress Stepper component

added:

- react component
- tests
- docs
- html (macro)
- examples

Screenshot 2024-12-06 at 12 58 34

Screenshot 2024-12-06 at 12 58 46

Screenshot 2024-12-09 at 11 02 29


packages/design/tailwind/css/components.css
```diff
@@ -944,6 +944,88 @@ input[type='file' i] {
...
+}
+
+.gi-progress-stepper[data-orientation='vertical'] {
+  @apply gi-flex-col;
+}
+
+.gi-progress-stepper-step {
+  @apply gi-text-white 
+  gi-rounded-full 
+  gi-relative
```
BogdanDanilescu, line 960: not sure if this is a formatting issue but we have some classes that are being displayed line by line and some that are being displayed one after another in a single line, we should decide which formatting style we want to follow so we are all aligned

BogdanDanilescu (approved):
LGTM, Great work on this 👍

packages/design/tailwind/css/components.css
ramonmulia: good call, I believe it was something wrong with my lint plugin, going to put them in a line format.

packages/react/ds/src/progress-stepper/progress-stepper.tsx
```diff
@@ -0,0 +1,78 @@
...
+        className="gi-progress-stepper-step-container"
+        data-orientation={orientation}
+        data-current={isCurrentStep}
+        data-completed={isCompleted}
+        data-next={isNextStep}
+      >
+        <div className="gi-progress-stepper-step">
+          {isCompleted ? <Icon icon="check" /> : <div>{stepNumber}</div>}
+        </div>
+        {children}
```
pregno, line 39: Should we limit the number of chars?
Maybe let the content break on two lines if it exceeds an X number of chars and then ellipsis.

For example:
If you have a longer string like "This step is for requesting information."
we can render something like
```
This step is for
requesting info...
```

At least we don't risk to break the UI if the user put a longer text.

ramonmulia: good call, it makes sense, working on this.

pregno (approved):

packages/html/ds/src/progress-stepper/progress-stepper.test.ts
hamza14khan, line 1: Don't you think we should be adding more tests? 😄

packages/react/ds/src/progress-stepper/types.ts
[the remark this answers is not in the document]
pregno: I don't see a possible use case; I thought it was informative.
If the user can navigate backwards and forward, it's preferable to have some navigation buttons, IMHO.

What use case do you see for a callback?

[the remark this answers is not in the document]
hamza14khan: Oh ok, additional navigation buttons will help.

Just a thought if a user wants to directly navigate from 3 to 5. Example a tutorial app, where step 1 is a course video and then second step is a quiz and then the last step is a survey. The user goes directly to the last step after completing the course video, skipping the quiz.

packages/html/ds/src/progress-stepper/progress-stepper.test.ts
ramonmulia: yes for sure, I missed some.

PR #359, opened 2024-12-06
feat: table component html and macro implementation

hamza14khan:
- Add HTML and macro component for table.
- Expose associated components for documentation.
- Documentation and example jinja page update.

Testing:
Screenshot 2024-12-06 at 18 15 46

### DIFF
--- a/packages/html/ds/src/table/table.html
+++ b/packages/html/ds/src/table/table.html
@@ -1,0 +1,37 @@
+{% macro govieTable(props) %}
+  <table class="gi-table {{ props.className | default('') }}">
+    {% if props.captionText %}
+      <caption class="gi-table-caption-text gi-text-lg">{{ props.captionText }}</caption>
+    {% endif %}
+    {% if props.headers %}
+      <thead>
+        <tr>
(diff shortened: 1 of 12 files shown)


packages/react/ds/src/table/table.stories.tsx
```diff
@@ -163,6 +156,53 @@ export const Default: Story = {
     </Table>
   ),
 };
+export const BasicTable: Story = {
```
hamza14khan, line 159: Added this story to sync same stories across documentation, react and html.

packages/html/ds/src/table/table.html
```diff
@@ -0,0 +1,37 @@
+{% macro govieTable(props) %}
+  <table class="gi-table {{ props.className | default('') }}">
```
BogdanDanilescu, line 2: are we still adding the `className` as a prop for components ?

pregno: I don't see `className` in the schema, I think you can remove it

BogdanDanilescu (approved):

PR #360, opened 2024-12-09
feat: accordion component

BogdanDanilescu:
PR to implement the `Accordion` component for HTML and React.

For React we have used the pattern of composition like so:
```
    <Accordion>
      <AccordionItem>
        User content
      </AccordionItem>

      <AccordionItem >
         User content
      </AccordionItem>

      <AccordionItem >
         User content
      </AccordionItem>
    </Accordion>   
```

For HTML we have used an items array that contains the following:
```
label
content
defaultExpanded
disabled
```

_Both libraries achieve the same end result_
**P.S: The styles, tests and documentation will be covered in a later task**

_Screenshots_:
Screenshot 2024-12-09 at 15 00 30
Screenshot 2024-12-09 at 15 00 42


packages/html/ds/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,54 @@
...
+const Accordion = renderComponent<AccordionProps>(macro);
+
+const meta = {
+  component: Accordion,
+  title: 'Layout/Accordion',
+  parameters: {
+    macro,
+    docs: {
+      description: {
+        component: 'Accordion component',
```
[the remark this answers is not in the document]
BogdanDanilescu: This is just temporary, I was thinking to add the description and `argTypes` when we have a full understanding of what other props do we need to add or include any additional props to the component

packages/react/ds/src/accordion/accordion.tsx
```diff
@@ -0,0 +1,7 @@
+type Props = {
+  children: React.ReactNode;
+};
+
+export const Accordion = ({ children }: Props) => {
+  return <div className="gi-max-w-[700px]">{children}</div>;
```
pregno, line 6: Don't use this hardcoded size.
If you want to define a max width, try to evaluate `max-w-prose`

packages/react/ds/src/accordion/accordion-item.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useState } from 'react';
+import { cn } from '../cn.js';
+import { Icon } from '../icon/icon.js';
+
+type Props = {
```
pregno, line 5: Can we have a prop to let the user choose to make the icon appear on the left or right?

pregno (changes requested):

packages/react/ds/src/accordion/accordion.tsx
BogdanDanilescu: This was intended to be used as a temporary measure because as agreed previously, we will add the styles ( including a max width ) at a later iteration. But sure I can use prose for now

packages/react/ds/src/accordion/accordion-item.tsx
```diff
@@ -0,0 +1,42 @@
...
+  label,
+  disabled,
+}: Props) => {
+  const [isExpanded, setIsExpanded] = useState(defaultExpanded);
+
+  return (
+    <div
+      className={cn(
+        'gi-py-4 gi-border-b-gray-150 gi-border-b gi-border-solid',
+        disabled && 'gi-opacity-30',
```
ramonmulia, line 24: alt:

 cn(
        'gi-py-4 gi-border-b-gray-150 gi-border-b gi-border-solid',
        { 'gi-opacity-30': disabled }
    )

ramonmulia (approved):
LGTM

packages/react/ds/src/accordion/accordion-item.tsx
BogdanDanilescu: fixed

pregno (approved):

PR #368, opened 2024-12-11
feat: button and links

BogdanDanilescu:
PR to implement additional functionality to the `Link` components. We now have the ability to:
- use a `Link` that is styled as a button through the `asButton` prop which accepts all the props of the `Button` component
- use a `Link` that has the `button` tag but it is styled as a link through the `as` property

A `onClick` has also been added to the link component ( in React ) . For the HTML library the event listener will have to be added by the user since the render happens on the server

SB screenshots:
Screenshot 2024-12-11 at 16 39 36
Screenshot 2024-12-11 at 16 40 01


BogdanDanilescu:
> LGTM
> 
> Can you remove also the `'@/components/form/button'` in docs? It is used only in the homepage and we can use the new variant in there

Fixed

hamza14khan (approved):

ramonmulia (approved):

pregno (approved):

PR #369, opened 2024-12-11
feat: Progress Bar component

ramonmulia:
#### Adding progress bar component

- react component
- macro
- examples
- tests
- docs

Screenshot 2024-12-12 at 12 44 11

Screenshot 2024-12-12 at 12 44 49


apps/docs/content/3-components/2-library/progress-bar.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+      <ProgressBar label="Loading..." size="sm" />
+
+      <ProgressBar isIndeterminate />
+    ```
+
+  </TabPanel>
+</Tabs>
+
+## Usage
```
hamza14khan, line 72: I am currently following a practice where I am adding Tabs below Usage heading. Is that the expectations or Usage heading requires content?

@pregno

apps/docs/content/3-components/2-library/progress-bar.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+### Best Practices
```
hamza14khan, line 74: nitpick: If you could add the content here.

packages/html/ds/src/progress-bar/progress-bar.schema.ts
```diff
@@ -0,0 +1,38 @@
+import * as zod from 'zod';
+
+export const progressBarSchema = zod.object({
+  currentValue: zod
```
hamza14khan, line 4: suggestion: you can just use `value` and for `length` for `finalValue` instead.

What do you guys think? @pregno @BogdanDanilescu

packages/react/ds/src/progress-bar/progress-bar.tsx
```diff
@@ -0,0 +1,63 @@
...
+        className={cn('gi-progress-bar', progressBarStyles({ size }))}
+      >
+        <div
+          className={cn(progressBarStyles({ color, size }), {
+            'gi-progress-bar-indeterminate': isIndeterminate,
+          })}
+          style={isIndeterminate ? {} : { width: `${fillPercentage}%` }}
+        />
+      </div>
+      {label ? <span>{label}</span> : null}
```
hamza14khan, line 60: question: will this work instead?

```
label && <span>{label}</span>
```

hamza14khan (changes requested):
Looks good. Left couple of comments, awaiting feedback from the team. 👍

packages/react/ds/src/progress-bar/progress-bar.tsx
ramonmulia: pretty much the same, right?

packages/react/ds/src/progress-bar/progress-bar.tsx
```diff
@@ -0,0 +1,63 @@
+import { useId } from 'react';
```
BogdanDanilescu, line 1: Are we using this hook anywhere ?

ramonmulia: no, going to remove it

packages/react/ds/src/progress-bar/progress-bar.tsx
hamza14khan: Yes, in that case no need for ternary operator 👍

packages/html/ds/src/progress-bar/progress-bar.schema.ts
BogdanDanilescu: maybe we can use `value` instead of `currentValue` and `max` instead of `finalValue` ?

BogdanDanilescu:
Does the component work in a scenario where the progress bar is completed ? If so, can we add a story showing that ?

packages/html/ds/src/progress-bar/progress-bar.schema.ts
hamza14khan: Good catch! I would go for `max` and `value` too! 

Also, I can't think of a use-case where we will have `max` as 50% or anything less than 100%. Maybe a developer can use this for size of the progress bar but again `max` would be 100% either way.

apps/docs/content/3-components/2-library/progress-bar.mdx
ramonmulia: @hamza14khan do you have an example? 
All components I took a look it is above including Pagination.
Not sure if I'm missing something.

BogdanDanilescu (approved):

apps/docs/content/3-components/2-library/progress-bar.mdx
pregno: Use the tabs with the snippet code below each variant.
The `Usage` paragraph is intended for group best practices, do's and don'ts

packages/react/ds/src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+      value: 200,
+      size: 'lg',
+    });
+
+    const progressBar = screen.getByTestId('progress-bar');
+    expect(progressBar).toBeTruthy();
+
+    const progressBarInner = progressBar.firstElementChild;
+
+    console.log(progressBarInner);
```
hamza14khan, line 55: Hello 😄

hamza14khan (changes requested):
Just missing the following points:

- Removal of console log.
- Content for best practices.

hamza14khan (approved):
Looks good 👍

apps/docs/content/3-components/2-library/progress-bar.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+- Ensure progress bars are visually distinct and easy to interpret against their background for clarity.
+- Place progress bars close to the task or content they represent to maintain contextual relevance.
+- Provide additional textual feedback (e.g., percentage complete or estimated time) to improve user understanding, especially for long-duration tasks.
+
+#### When to use this component
```
pregno, line 104:
```suggestion
### When to use this component
```

apps/docs/content/3-components/2-library/progress-bar.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+- Use a determinate progress bar when the progress of a task can be measured (e.g., file upload, form completion).
+- Use an indeterminate progress bar when the duration of a task is unknown but progress should be indicated (e.g., loading state).
+
+#### When not to use this component
```
pregno, line 109:
```suggestion
### When not to use this component
```

pregno (changes requested):

PR #371, opened 2024-12-16
feat: add html and macro for tooltip

hamza14khan:
Addition of:

- Tooltip HTML and jinja macro.
- Documentation update.
- Testing.
- Example page update.
- Tooltip instance initialised.
- Root level decorator for tooltip.

Documentation site:

Screenshot 2024-12-16 at 11 13 58

### DIFF
--- a/apps/docs/content/3-components/2-library/tooltip.mdx
+++ b/apps/docs/content/3-components/2-library/tooltip.mdx
@@ -11,7 +11,7 @@ libraries:
     link: '?path=/docs/application-tooltip--docs'
   - platform: 'react'
     status: 'beta'
-    link: ?path=/docs/application-tooltip--docs'
+    link: '?path=/docs/application-tooltip--docs'
 ---
 
 # Tooltip
--- a/apps/docs/content/3-components/2-library/tooltip.mdx
+++ b/apps/docs/content/3-components/2-library/tooltip.mdx
@@ -7,7 +7,8 @@ libraries:
     link: '?path=/docs/application-tooltip--docs'
     status: 'stable'
   - platform: 'global'
-    status: 'considering'
+    status: 'beta'
+    link: '?path=/docs/application-tooltip--docs'
   - platform: 'react'
     status: 'beta'
     link: ?path=/docs/application-tooltip--docs'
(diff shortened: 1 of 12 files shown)


packages/html/storybook/.storybook/preview.tsx
```diff
@@ -22,6 +22,11 @@ const Decorator = (arguments_, parameters) => {
     }
     return classes;
   }
+
+  if (parameters?.macro?.name === 'govieTooltip') {
```
hamza14khan, line 26: thanks @BogdanDanilescu

apps/docs/content/3-components/2-library/tooltip.mdx
```diff
@@ -7,7 +7,8 @@ libraries:
     link: '?path=/docs/application-tooltip--docs'
     status: 'stable'
   - platform: 'global'
-    status: 'considering'
+    status: 'beta'
+    link: '?path=/docs/application-tooltip--docs'
   - platform: 'react'
     status: 'beta'
     link: ?path=/docs/application-tooltip--docs'
```
pregno, line 14: missing `'`

pregno: Considering the definition below of `Components/Tooltip` I think that the link should be `?path=/docs/components-tooltip--docs`

pregno (dismissed):

pregno (approved):

packages/html/ds/src/tooltip/tooltip.schema.ts
```diff
@@ -0,0 +1,39 @@
...
+      description:
+        'Defines the position of the tooltip relative to the target element.',
+    })
+    .default('top'),
+  id: zod
+    .string({
+      description: 'Sets the ID for the tooltip, used for accessibility.',
+    })
+    .optional(),
+  ariaLabel: zod
```
ramonmulia, line 26: what do you think to have something like this:

const ariaAttributes = zod.object({
  label: zod
    .string({
      description: 'Aria label for the element',
    })
    .optional(),
  describedBy: zod
    .string({
      description: 'ID of the element that describes this element',
    })
    .optional(),
});

and share among the compoents schema as we go..
 
 
 example:
 
   id: zod
    .string({
      description: 'Sets the ID for the tooltip, used for accessibility.',
    })
    .optional(),
    aria: ariaAttributes.optional(),

BogdanDanilescu (approved):
LGTM

packages/html/ds/src/tooltip/tooltip.schema.ts
hamza14khan: I think this is good. Gov UK is doing something similar. I also think this goes with all the other components as well. What do you think tackling down this in a separate task for other components as well? Happy to create a task for this as well.

ramonmulia (approved):
LGTM

PR #373, opened 2024-12-16
feat: adding slots feature in govieHeader macro - in progress

ramonmulia:
#### Adding slots feature in govieHeader macro
- in progress

### DIFF
--- a/packages/html/ds/src/header/header.stories.ts
+++ b/packages/html/ds/src/header/header.stories.ts
@@ -21,6 +21,86 @@ const meta = {
 export default meta;
 type Story = StoryObj<typeof meta>;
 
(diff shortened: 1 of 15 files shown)


packages/html/ds/src/header/header.stories.ts
```diff
@@ -97,6 +155,10 @@ export const Default: Story = {
         href: '#',
         label: 'Gaeilge',
       },
+      {
```
pregno, line 158: Please, remove `languages` from the Default story (also in React).

pregno (changes requested):

pregno (approved):

hamza14khan (approved):

PR #374, opened 2024-12-17
fix: assets and logo

pregno:


ramonmulia (approved):

packages/react/ds/src/assets/logos/gov-of-ireland.tsx
```diff
@@ -1,4 +1,4 @@
-function Logo() {
+function LogoGovOfIreland() {
```
hamza14khan, line 1: suggestion: Can this be renamed with `GovieLog`?

hamza14khan (approved):

BogdanDanilescu (approved):

packages/react/ds/src/assets/logos/gov-of-ireland.tsx
pregno: I want to keep a meaningful name, like with the file name

PR #377, opened 2024-12-18
chore: update table stories with tooltip

hamza14khan:
- Sync React props with macro prop. `label` renamed with `text`.
- Updated html, macro and react stories with tooltip usage.

Example:
Screenshot 2024-12-18 at 10 16 11


pregno (approved):

ramonmulia (approved):
LGTM

PR #378, opened 2024-12-18
feat: add external link property to header and footer

G100g:
This pull request updates the `Header` and `Footer` components to ensure that specified links now point to external URLs instead of internal routes.


BogdanDanilescu:
do we want to implement the same logic in the macros library ? to ensure we have consistency between them ?

packages/react/ds/src/primitives/anchor.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+type AnchorProps = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
+  external?: boolean;
```
ramonmulia, line 4: There is a Link component with ```external``` property, maybe adding a variant there instead of creating a new anchor component?

G100g: Hey @ramonmulia thank you for the feedback.

The `Link` already have an external prop, but it also contains the `gi-link` style class by default.
That's why I created a primitive one with no style.

G100g: Maybe I can use Anchor primitive component inside the link component to avoid duplicated code.
Leaving only the logic inside the primitive
What you think?

ramonmulia: Good catch, I like this idea

G100g: I updated the component using the primitive

ramonmulia (approved):
LGTM

pregno (approved):

hamza14khan (approved):

PR #381, opened 2024-12-20
feat: composable combo box

BogdanDanilescu:
PR to make the `Combobox` composable and decouple the form from the component.

Components added:
- `Combobox`
- `DropdownItem`
- `Form` - this component has been created in a separate `/common` folder as it can be used in other components as well, not just `Combobox`

SB screenshot
Screenshot 2024-12-20 at 12 15 29


packages/react/ds/src/combo-box/dropdown-item.tsx
```diff
@@ -68,7 +68,7 @@ export const DropdownItem = ({
   };
 
   return (
-    <div>
+    <div role="group" aria-label={`${children} dropdown`}>
```
ramonmulia, line 71: what about to expose aria attributes so the user can decide the label?

something like:
```aria-label={props?.aria?.label ||  `${children} dropdown`}```

ramonmulia: as we have a specific task to add aria attributes I think would be better to cover this on aria task.

pregno (approved):

ramonmulia (approved):

hamza14khan (approved):

PR #382, opened 2025-01-02
poc: aria attributes for button 

hamza14khan:
This PR focuses on:

- POC for adding aria attributes on button using a helper created in common.
- Associated testing and schema.

Testing:

Screenshot 2025-01-02 at 17 29 59
Screenshot 2025-01-02 at 17 30 20

### DIFF
--- a/packages/html/ds/src/button/button-schema.ts
+++ b/packages/html/ds/src/button/button-schema.ts
@@ -31,6 +31,13 @@ export enum ButtonType {
   Button = 'button',
 }
 
+ export const ariaSchema = zod.record(
+  zod.string({
+    description: 'ARIA attributes key-value pairs',
+  }),
+  { description: 'An object of ARIA attributes' }
+).optional();
+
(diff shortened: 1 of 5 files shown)


packages/html/ds/src/button/button-schema.ts
```diff
@@ -31,6 +31,13 @@ export enum ButtonType {
   Button = 'button',
 }
 
+export const ariaSchema = zod.record(
```
ramonmulia, line 34: what about to have something like this:
```
const validAriaProps = [ ----allowed aria props for buttons ----];

export const ariaSchema = zod
  .record(zod.enum(validAriaProps) ....
  ```
  
  so we allow only button aria props..

ramonmulia (changes requested):

packages/html/ds/src/button/button-schema.ts
hamza14khan: Yeah a good one. Will fix it now.

PR #384, opened 2025-01-06
test: Axe tests for all component variants

ramonmulia:
it Includes the structure to test all axe variants in react and macro library.


hamza14khan (approved):
Solid work!

PR #386, opened 2025-01-08
fix: add aria attributes for text-input, checkbox, radio, button, link and tooltip

hamza14khan:
Update schema for aria attributes for accessibility in HTML/Macro components:

- Tooltip.
- Text input.
- Checkbox.
- Radio.
- Button.
- Link.


ramonmulia (approved):
LGTM

packages/html/ds/src/button/button.html
```diff
@@ -14,6 +14,7 @@
 
   {% if props.href %}
     <a
+      {{ commonAriaAttributes(props) }}
```
pregno, line 17: is import `{% from 'common/aria-attributes.html' import commonAriaAttributes %}` missing?

hamza14khan: Oh so, the import was already made for button in the poc PR for button. This one is for the button as type link.

Reference: https://github.com/ogcio/govie-ds/pull/382

pregno (approved):

PR #388, opened 2025-01-09
fix: fixing Header component and templates

ramonmulia:
### Fixes for
- Header component reactjs and macro
- Basic Template
- Search Template


packages/html/ds/src/header/header.html
```diff
@@ -86,99 +86,105 @@
 
     <div id="HeaderContainer" class="{{ containerClassName }}">
       <div class="{{ menuContainerClassNames }}">
-        <div class="gi-header-logo">
-          {% if props.logo.href %}
-            <a href="{logo.href}" aria-label="Go to the home page">
+        <div>
+          <div class="gi-header-logo">
+            {% if props.logo.href %}
+              <a href="{logo.href}" aria-label="Go to the home page">
```
hamza14khan, line 92: question: if this is a typo? 

`{{ logo.href }}` instead of `{logo.href}`

hamza14khan (changes requested):
Looks good, just one comment.

packages/html/ds/src/header/header.html
ramonmulia: good catch!
it was there for a long time nobody saw it.
fixed!

pregno (approved):

PR #393, opened 2025-01-13
feat: Modal enhancements

ramonmulia:
### Modal enhancements
- react/docs/examples changes 
- macro updates 

all the ```right```, ```left``` and ```bottom``` styles for prop ```position``` will be addressed in drawer component pr


packages/react/ds/src/modal/modal.tsx
```diff
@@ -1,45 +1,146 @@
...
-        </div>
+        <ModalCloseButton onClick={onClose} label={closeButtonLabel} />
+        <div className="gi-flex gi-flex-col gi-h-full">{renderChildren}</div>
       </div>
     </div>
   );
 };
+
+export const ModalTitle = ({ children }: { children: string }) => (
+  <Heading as="h2">{children}</Heading>
```
pregno, line 100: I wonder if we should put the size variant of the title as a property

pregno (approved):

PR #400, opened 2025-01-17
feat: add layout page example

hamza14khan:
Addition of layout page with:

- 2/3 column layout.
- 1/2 column layout.
- Hero banner.
- Layout pattern example with HTML.

Screenshot 2025-01-17 at 16 08 12


packages/design/tailwind/css/layout.css
```diff
@@ -26,4 +26,12 @@
   @apply gi-w-full;
 }
 
+.gi-layout-container-2-column {
```
ramonmulia, line 29: I'm not completely sure, but I think you could use the Stack component for that instead of creating column classes?

https://storybook-react.design-system.blocks.gov.ie/?path=/docs/layout-stack--docs

pregno: The stack is available only on React.

pregno (approved):

PR #401, opened 2025-01-20
feat: add social network icons

pregno:


packages/html/ds/src/icon/icon.schema.ts
hamza14khan, line 1: I think the variable names are self explanatory so I don't think comments are adding any value here.

hamza14khan (changes requested):
Looks good one minor change requested

hamza14khan (approved):

packages/react/ds/src/icon/icon.tsx
```diff
@@ -31,11 +40,38 @@ export function Icon({
   ...props
 }: IconPropTypes) {
   const fontSize = meta.light.resolved.primitive.size[size].$value;
+  const display = inline ? '' : 'gi-block';
+  const strokeColor = disabled ? 'gi-stroke-gray-700' : '';
+  const iconColor = disabled ? 'gi-text-gray-700' : '';
+
+  if (icon === 'social_bluesky') {
+    return <Bluesky size={fontSize} className={cn(display, strokeColor)} />;
```
ramonmulia, line 48: alt: 
```className={cn({'gi-block': inline, ...}, strokeColor)}```

ramonmulia (approved):

PR #418, opened 2025-01-30
fix: add fixed table layout

pregno:


packages/html/ds/src/table/table.schema.ts
```diff
@@ -29,6 +29,7 @@ export const tableSchema = zod.object({
         'An array of strings representing column headers for the table.',
     })
     .optional(),
+  layout: zod.enum(['auto', 'fixed']).optional(),
```
ramonmulia, line 32: most of the components are using ```.nativeEnum```.
I'm not sure if ```.enum``` would cause issues while building or running lint directly from package/html folder

ramonmulia (approved):
LGTM.
just a comment about .enum

hamza14khan (approved):

packages/html/ds/src/table/table.schema.ts
pregno: updated

PR #433, opened 2025-02-10
chore: Removing enums

ramonmulia:


packages/react/ds/src/checkbox/types.ts
```diff
@@ -0,0 +1,23 @@
+import { InputHTMLAttributes } from 'react';
+
+export const CheckboxSizeEnum = {
+  Large: 'lg',
+  Medium: 'md',
+  Small: 'sm',
+} as const;
+
+export type CheckboxSizeEnumType =
```
pregno, line 9: I think it's better to not add additional types, you can use the same name
```suggestion
export type CheckboxSizeEnum =
```

ramonmulia: using a const object doesn’t automatically give us a type like an enum does.... I mean you can use enum as type and const at the same time.

 so we do need a separate type definition, we can reuse the same name  “name shadowing” (having a const and type with the same name), but having a separate name keeps it clearer that one is a runtime object and the other is a type..

packages/html/ds/src/link/link.schema.ts
```diff
@@ -1,10 +1,11 @@
 import * as zod from 'zod';
 import { buttonSchema } from '../button/button-schema';
+import { getEnumValues } from '../helpers';
 
-export enum LinkSize {
-  SMALL = 'sm',
-  MEDIUM = 'md',
-}
+export const LinkSize = {
+  SMALL: 'sm',
```
pregno, line 6: Check that all the "small" are named `sm` and not `SMALL` or `Small`.
This comment applies to all the enums.

pregno (changes requested):

hamza14khan (approved):
Great job! Approving it now, but can be merged after Andrea's comments are resolved. 👍

packages/html/ds/src/alert/alert-schema.ts
```diff
@@ -1,15 +1,18 @@
...
+import { getEnumValues } from '../helpers';
 
-export enum AlertVariant {
-  INFO = 'info',
-  DANGER = 'danger',
-  SUCCESS = 'success',
-  WARNING = 'warning',
-}
+export const AlertVariant = {
+  INFO: 'info',
```
pregno, line 5: I also noticed that sometimes we are using ALL_CAPS, and sometimes not.
Can we uniform this one too?

pregno (changes requested):

packages/html/ds/src/link/link.schema.ts
ramonmulia: As this PR is becoming huge I think we can have a separated PR to deal with the "enum" values.

All the "enum" keys are in uppercase so we have a pattern to follow, we can address the values changes in a separeted PR.

what do you think?

pregno (approved):

PR #434, opened 2025-02-11
chore: update card to accept custom aspect ratio

hamza14khan:
Update React & HTML `Card` component to accept custom aspect ratios and minor styling update on vertical card.

Testing:
Screenshot 2025-02-11 at 01 56 58


packages/react/ds/src/card/card.tsx
```diff
@@ -13,6 +13,7 @@ type Action =
 type ImagePropTypes = {
   src: string;
   alt?: string;
+  aspectRatio?: string;
```
pregno, line 16:
```suggestion
  aspectRatio?: '4/3' | '1/1' | string;
```

pregno (changes requested):
Let's provide the defaults as options

ramonmulia (approved):

PR #445, opened 2025-02-19
fix: update link as button classes

hamza14khan:
- Focus state classes removed on link as button because it will eventually use button focus state.

### DIFF
--- a/packages/html/ds/src/link/link.html
+++ b/packages/html/ds/src/link/link.html
@@ -61,7 +61,7 @@
     {{ getExternalAttribute(props.external) | trim }}
     data-testid="{{ props.dataTestid }}"
     class="
-      gi-link
+      {% if not props.asButton %} gi-link {% endif %}
       {{ getColorClass(props.noColor) | trim }}
       {{ getUnderlineClass(props.noUnderline) | trim }}
       {{ getVisitedClass(props.noVisited) | trim }}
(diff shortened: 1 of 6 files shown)


apps/docs/content/3-components/2-library/link.mdx
```diff
@@ -54,7 +54,7 @@ Links are blue and underlined by default. If your link is at the end of a senten
 
 <ComponentContainer>
   <Link href="#" noColor>
```
hamza14khan, line 56: thought: not sure why `p` was being added automatically, so a workaround was to add a `span` tag.

apps/docs/content/3-components/2-library/chip.mdx
```diff
@@ -25,7 +25,13 @@ It allow users to close the chip using keyboard interactions. For accessibility,
 
 <ComponentContainer>
   <Chip label="Label 1" />
+</ComponentContainer>
```
pregno, line 28: This will solve the problem in the documentation, but what if a user wants to add Chip items in an horizontal stack component?

hamza14khan: They will achieve it like this:

Code:
```
<div className="gi-gap-4 gi-flex-row gi-flex gi-w-fit">
    <Chip onClose={() => null} label="Default" />
    <Chip onClose={() => null} label="Hover" />
    <Chip onClose={() => null} label="Focus" />
</div>
```

Output:
Screenshot 2025-02-21 at 02 27 49

pregno (approved):

PR #451, opened 2025-02-21
feat: add support to react router

pregno:
Added support to:
* React Router Link
* NextLink

Link and Anchor are supporting Slot using [radix-ui/slot](https://www.radix-ui.com/primitives/docs/utilities/slot)
When you need to replace the anchor element (`a`) you can do it in this way:
```
<Link asChild>
  <button>This is a button</button>
</Link>
```
So if you want to use NextLink (or React Router Link) you can do it in this way:
```
<Link asChild>
  <NextLink href={'/destination'}>This is a NextLink</NextLink>
</Link>
```

In the same way, the link item in the header has been reviewed to support this use case.


hamza14khan (approved):

hamza14khan (approved):
Looks good 👍

PR #453, opened 2025-02-25
fix: restore half width on input

pregno:


hamza14khan (commented):
Should we not apply changes for html as well?

pregno:
> Should we not apply changes for html as well?

I will add a story for that

hamza14khan (approved):

PR #465, opened 2025-03-06
fix: update tab list checked state

hamza14khan:
- Update initial set active tab if any children got `checked` true.
- If no `checked` state is defined then set `0` index tab item to be visible.
- Added associated story for that.
- Added associated unit test.

Testing Screenshot:
Screenshot 2025-03-06 at 03 58 01


ramonmulia (approved):
LGTM

pregno (approved):

PR #467, opened 2025-03-06
chore: use semantic tokens for header colors

pregno:
* add semantic tokens for primary colors
* update header for use primary colors instead of specific color shade


ramonmulia (approved):
LGTM

hamza14khan (approved):

PR #501, opened 2025-04-03
fix: update pagination and use breakpoint hook

hamza14khan:
## Issue
React compares the server-rendered HTML (based on `xs`) with the initial client-side render (based on `xl`). They don't match! This causes a hydration mismatch error.

## Description
The issue was fixed by moving the useEffect hook to the component itself and using a variable to check if client was ever loaded and then using the default values.

## Type of Issue

- [ ] 🚀 Feature
- [X] 🐛 Bug
- [ ] 🔧 Chore

## Checklist

- [X] I have added/updated tests to cover the changes made (if applicable).

## Related Issues
N/A.

## Additional Notes
N/A.


examples/nextjs/src/app/page.tsx
```diff
@@ -185,17 +186,17 @@ export default function Home() {
             </Combobox>
           </Form>
           <h2>Checkboxes Group</h2>
-          <CheckboxGroup
+          {/* <CheckboxGroup
```
hamza14khan, line 189: Commented these components out as they were throwing errors on VSCode, can look up into this in some improvement PR.

packages/react/ds/src/pagination/pagination.tsx
```diff
@@ -1,13 +1,12 @@
 'use client';
```
pregno, line 1: Will it work on SSR with `use client` defined?

examples/nextjs/src/app/page.tsx
pregno: I fixed those already in the lint PR

hamza14khan: Perfect! I will let the lint PR merge first.

packages/react/ds/src/pagination/pagination.tsx
hamza14khan: Yes it should. Isn't that true for other components as well? They use the same approach 🤔

ramonmulia (approved):

PR #505, opened 2025-04-04
feat: Sync Link component with figma

ramonmulia:
## Description

Sync Link component with Figma

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore

## Checklist

## Related Issues

[27044](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/27044)

Screenshot 2025-04-04 at 13 31 58


pregno (approved):

packages/react/ds/src/link/link.tsx
```diff
@@ -17,62 +18,139 @@ export type LinkProps = AnchorProps & {
...
+            'gi-link-disabled': disabled && !asButton && !asChild,
+            'gi-link-light':
+              appearance === 'light' &&
+              !asButton &&
+              !asChild &&
+              !noColor &&
+              !noVisited &&
+              !disabled,
+          },
+          `gi-paragraph-${size || 'sm'}`,
```
pregno, line 129: Be careful using `gi-paragraph-{size}` because it is also setting the font weight.
For example, if you use the link in a Header you will see mixed styles like **Header with** [link](www.google.com)
Please take a look at the storybook of the Card, the title of the card used to be bold.

pregno (changes requested):

hamza14khan (approved):

packages/react/ds/src/link/link.tsx
hamza14khan: Also, one minor thing: if its worth adding `gi-paragraph-{SIZE}` to the safelist in tailwind config since it is dynamically being evaluated?

cc @pregno

packages/react/ds/src/link/link.tsx
```diff
@@ -17,62 +18,141 @@ export type LinkProps = AnchorProps & {
...
+            '!gi-inline-flex': asButton,
+            'gi-link-disabled': disabled && !asButton && !asChild,
+            'gi-link-light':
+              appearance === 'light' &&
+              !asButton &&
+              !asChild &&
+              !noColor &&
+              !noVisited &&
+              !disabled,
+            'gi-text-sm': !size || size === 'sm',
```
pregno, line 128: Define the default in the init props at line 87.
Set the `md` as the default since it is the default for the text.

pregno (changes requested):

packages/react/ds/src/link/link.tsx
ramonmulia: @pregno I was wondering about that, to set to md as default, but what if someone else is using the link expecting sm? 
Card title is an example

pregno (approved):

PR #518, opened 2025-04-09
feat: add button group component

hamza14khan:
## Description

- Add react button group component.
- Add html button group implementation.
- Associated docs update.

Screenshot 2025-04-15 at 03 46 53

## Type of Issue

- [X] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

- [X] I have included relevant attachments for user testing (e.g., screenshots of Storybook showcasing the changes).
- [X] I have added/updated tests to cover the changes made (if applicable).
- [X] I have updated the documentation site content (if applicable).

## Related Issues
N/A.

## Additional Notes
N/A.

### DIFF
--- a/apps/docs/content/3-components/2-library/button.mdx
+++ b/apps/docs/content/3-components/2-library/button.mdx
@@ -426,12 +426,12 @@ Used to display a collection of related buttons (e.g., for feedback, selection, 
 
 ### Best Practices
 
-#### ✅ When to use this component
+#### When to use this component
 
 - Use **Button** for user-triggered actions like submitting a form or confirming a choice.
 - Use **ButtonGroup** when offering a set of mutually exclusive options, such as rating selections or binary questions.
 
-#### 🚫 When not to use this component
+#### When not to use this component
 
 - Don’t use **Button** when the goal is navigation — use a **Link** component instead.
 - Don’t use **ButtonGroup** if you only need a single action — use just a **Button** in that case.
--- a/apps/docs/content/3-components/2-library/button.mdx
+++ b/apps/docs/content/3-components/2-library/button.mdx
@@ -56,13 +56,14 @@ libraries:
   </TabList>
   <TabPanel value="tab21">
     ```html
-    <button id="button" class="gi-btn gi-btn-secondary gi-btn-regular">Secondary Button</button>
+    <button id="button" class="gi-btn gi-btn-secondary gi-btn-regular">
+      Secondary Button
+    </button>
     ```
   </TabPanel>
   <TabPanel value="tab23">
-    ```jsx
+    ```jsx 
     import { Button } from '@govie-ds/react';
-    
     <Button variant="secondary">Secondary Button</Button>
     ```
   </TabPanel>
@@ -80,14 +81,15 @@ libraries:
     <TabItem value="tab33">React</TabItem>
   </TabList>
   <TabPanel value="tab31">
-  ```html
-  <button id="button" class="gi-btn gi-btn-flat gi-btn-regular">Flat Button</button>
-  ```
+    ```html
+    <button id="button" class="gi-btn gi-btn-flat gi-btn-regular">
+      Flat Button
+    </button>
+    ```
   </TabPanel>
   <TabPanel value="tab33">
-    ```jsx
+    ```jsx 
     import { Button } from '@govie-ds/react';
-    
     <Button variant="flat">Flat Button</Button>
     ```
   </TabPanel>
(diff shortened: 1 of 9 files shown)


packages/react/ds/src/button/button-group/button-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <FormField
+      className="gi-w-full"
+      label={{
+        text: label,
+        htmlFor: name,
+      }}
+      hint={hint ? { text: hint } : undefined}
+    >
+      <div className="gi-flex gi-flex-wrap gi-gap-2 gi-pt-2 gi-justify-between">
+        {options.map((option) => (
```
ramonmulia, line 47: this component uses "options" prop but I think would be better to follow the child composability pattern like some components: Select, Breadcrumbs, Tabs, ProgressStepper and so on, something like:
```
<ButtonGroup name="my-group">
  <ButtonGroupItem value="1">One</ButtonGroupItem>
  <ButtonGroupItem value="2">Two</ButtonGroupItem>
</ButtonGroup>
```
it would be much more flexible and easy to extend.
what do you think?

packages/react/ds/src/button/button-group/button-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <FormField
+      className="gi-w-full"
+      label={{
+        text: label,
+        htmlFor: name,
+      }}
+      hint={hint ? { text: hint } : undefined}
+    >
+      <div className="gi-flex gi-flex-wrap gi-gap-2 gi-pt-2 gi-justify-between">
```
ramonmulia, line 46: why not use Stack component here?

packages/react/ds/src/button/button-group/button-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {options.map((option) => (
+          <Button
```
ramonmulia, line 48: I believe would be better to use InputRadio instead of Button, what do you think?

packages/react/ds/src/button/button-group/button-group.tsx
hamza14khan: Sounds good!

packages/react/ds/src/button/button-group/button-group.tsx
hamza14khan: Stack would've been overkill for just a plain row button render and top of that as per designs for mobile we might not want to row every button instead center as they wrap.

packages/react/ds/src/button/button-group/button-group.tsx
hamza14khan: Maybe I am not following you on this?

I think there's something similar available [here](https://ds.services.gov.ie/storybook-react/?path=/story/form-radio-inputradiogroup--inline&globals=viewport:responsive), so it may become duplicate.

packages/design/tailwind/css/components/button.css
```diff
@@ -118,3 +118,13 @@
 .gi-btn-regular {
   @apply gi-h-10 gi-px-3 gi-py-2 gi-text-sm;
 }
+
+/* Button Group */
+
+.gi-btn-group {
+  @apply gi-flex gi-flex-wrap gi-gap-3 gi-justify-center sm:gi-justify-start;
```
pregno, line 125: Stretch the items so they will all have the same width.
Consider using a grid since we have to control the behaviour of items when shown on mobile.

pregno (changes requested):

packages/design/tailwind/css/components/button.css
```diff
@@ -118,3 +118,13 @@
...
+}
+
+.gi-btn-group button {
+  @apply gi-min-w-[48px] gi-justify-center;
```
ramonmulia, line 129: replace for gi-min-w-12

pregno (approved):

packages/react/ds/src/button/button-group/button-group.tsx
```diff
@@ -0,0 +1,79 @@
...
+    </Button>
+  );
+};
+
+type ButtonGroupProps = {
+  name: string;
+  size?: ButtonSize;
+  onChange?: (value: string) => void;
+  defaultValue?: string;
+  children?: React.ReactNode;
```
ramonmulia, line 58: you don't need to declare children here, use ```PropsWithChildren<>``` instead.

packages/react/ds/src/button/button-group/button-group.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+export const ButtonGroupItem: React.FC<ButtonGroupItemProps> = ({
+  value,
+  children,
+}) => {
+  const context = React.useContext(ButtonGroupContext);
+
+  if (!context) {
+    throw new Error('ButtonGroupItem must be used within a ButtonGroup');
```
ramonmulia, line 30: cool!

ramonmulia (approved):
minor comments, LGTM

apps/docs/content/3-components/2-library/button.mdx
```diff
@@ -200,17 +212,170 @@ libraries:
...
+            colleague?
+          </label>
+          <div class="gi-hint-text-md gi-hint-text gi-mb-1">
+            1 = Not likely, 10 = Extremely likely
+          </div>
+        </div>
+      </div>
+      <div class="gi-btn-group">
+        <button
+          class="gi-btn gi-btn-large gi-btn-secondary-dark"
```
pregno, line 279: Can we have a single class here? something like `gi-btn-group-item`

PR #553, opened 2025-05-16
fix: update combo box focus state

hamza14khan:
## Description

- Fixed content on focus state that was being pushed inside.
- Usage of `Button` component on React and associated button classes in HTML.
- Missing HTML documentation.

## Type of Issue

- [ ] 🚀 Feature
- [X] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

- [X] I have updated the documentation site content (if applicable).

## Related Issues
N/A.

## Additional Notes
N/A.


ramonmulia (approved):

packages/html/ds/src/helpers/combobox.ts
```diff
@@ -78,10 +79,6 @@ export const createComboBox = (arguments_: ComboBoxProps) => {
     containerOpen.append(checkBoxContainer);
     checkBoxContainer.className = 'gi-combobox-checkbox-container';
 
-    if (!item.noSearch) {
-      checkBoxContainer.className += ' gi-h-64';
```
pregno, line 82: This is needed for avoid layout shifts when the results are few. Is there any reason why you removed this?

hamza14khan: Yes, I tested it out with fewer data and it seems like it is taking redundant space. Also if you have no data then it will show a blank container of 64h. I tested out mobile as well and it seems like its of no use. Can you please specify what layout shifts I have not explored? Maybe I missed something 👍

pregno: Check here, you see that the container is reducing while typing and there is also a missing padding for the empty state.

https://github.com/user-attachments/assets/7c0022a3-ce03-4eb5-999f-208eded1cbd3

packages/react/ds/src/combo-box/dropdown-item.tsx
```diff
@@ -123,11 +126,7 @@ export const DropdownItem = ({
           </div>
         )}
 
-        <div
```
ramonmulia, line 126: I believe you should keep this.

pregno (approved):

PR #577, opened 2025-06-04
chore: update menu input type to button for screen reader

hamza14khan:
## Description
This should fix the screen reader issue that says "Tickbox" or "Button, group" using Voice Over or other tools.

## Type of Issue

- [ ] 🚀 Feature
- [X] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
N/A.

## Additional Notes
We should also test other components similarly.

### DIFF
--- a/packages/react/ds/src/header/components/header-slot.tsx
+++ b/packages/react/ds/src/header/components/header-slot.tsx
@@ -47,7 +47,7 @@ const DrawerTrigger = ({
           className="gi-block gi-w-0 gi-absolute gi-h-0"
           id={`ItemActionDrawerTrigger-${index}`}
           data-index={index}
-          type="checkbox"
+          type="button"
         />
         {label && <span className="label">{label}</span>}
         {icon && (
@@ -103,7 +103,7 @@ export const SlotItemAction = ({ item, index }: HeaderSlotProps) => {
         data-index={index}
         aria-expanded="false"
         aria-controls={`SlotContainer-${index + 1}`}
-        type="checkbox"
+        type="button"
       />
       {item.label && <span className="label">{item.label}</span>}
       {item.icon && (


packages/react/ds/src/header/header.tsx
```diff
@@ -208,7 +208,6 @@ export function Header({
               const { label, showItemMode = 'desktop-only' } = item;
               return (
                 <div
-                  aria-label={label}
```
pregno, line 211: I think it's better to add an `aria-label` prop (optional)

packages/react/ds/src/header/components/header-slot.tsx
```diff
@@ -58,6 +58,7 @@ const DrawerTrigger = ({
             })}
             icon={icon}
             id={`ItemIconDrawerActionTrigger-${index}`}
+            ariaHidden={!!label}
```
pregno, line 61:
```suggestion
            ariaHidden={true}
```
The value is the icon name and might differ from the label's meaning.

packages/react/ds/src/header/components/header-slot.tsx
```diff
@@ -47,7 +47,7 @@ const DrawerTrigger = ({
           className="gi-block gi-w-0 gi-absolute gi-h-0"
           id={`ItemActionDrawerTrigger-${index}`}
           data-index={index}
-          type="checkbox"
+          type="button"
         />
         {label && <span className="label">{label}</span>}
```
pregno, line 52: we can consider to use `ariaHidden` here if the `aria-label` prop has a value

pregno (changes requested):

pregno (approved):

ramonmulia (approved):

PR #907, opened 2026-03-31
feat(AB#35985): Adding Paragraph mitosis component 

ramonmulia:
## Description

Adding Paragraph mitosis component 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues

[35985](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/35985)


Ademsk1 (commented):
I am personally not a fan of the `getSize` and `getWhitespace` functions when we can allow tailwind-variants to handle most of it - and am not a fan of the objects we've created to use it. It feels a little redundant when we can just pass the size straight in to tv. Other than that the new work looks good.

Ademsk1 (commented):
do we want to also export GiParagraph? Or have it purely used internally, and have the original <Paragraph/> as the wrapper?

packages/react/src/paragraph/paragraph.tsx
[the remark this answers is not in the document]
ramonmulia: good call!

DmitryB-NF (changes requested):

DmitryB-NF (changes requested):
The implementation looks good! We basically have the entire Typography covered by Mitosis now 🎉 
Could you check the `styles` discussion to make sure we have a consistent API across all the components?

DmitryB-NF:
@ramonmulia The PR should be properly rebased to drop excessive commits. Let me know if you need help to resolve this

DmitryB-NF (dismissed):
@ramonmulia, could you resolve conflicts to merge this?

packages/angular/src/stories/Paragraph.stories.ts
```diff
@@ -0,0 +1,113 @@
...
+  AllParagraphSizes as allSizes,
+  AllWhitespaces as allWhitespaces,
+  AllAlignments as allAlignments,
+} from '../atoms/storybook/Paragraph.meta';
+
+const meta = {
+  ...paragraphMeta,
+  argTypes: {
+    ...paragraphMeta.argTypes,
+  },
```
DmitryB-NF: I feel like this is excessive to copy argTypes
```suggestion
  ...paragraphMeta,
```

packages/react/src/paragraph/paragraph.tsx
```diff
@@ -1,107 +1,61 @@
...
-      }
-    }
-  })();
-  if (As === 'span') {
+  if (As === 'p') {
     return (
-      <Text
+      <GiParagraph
         size={size}
+        align={align}
```
DmitryB-NF: @ramonmulia @Ademsk1 it seems like before the change, `align` has been applied to `Text` also and now it is not. Does it make sense to add at least an additional `className` to the internal `Text` to make sure the existing Paragraph is not broken?
It could be done as a separate PR to fix this with a comment

packages/react/src/paragraph/paragraph.tsx
```diff
@@ -1,107 +1,61 @@
...
-      case 'break-spaces': {
-        return 'gi-whitespace-break-spaces';
-      }
-      default: {
-        return 'gi-whitespace-normal';
-      }
-    }
-  })();
-  if (As === 'span') {
+  if (As === 'p') {
```
DmitryB-NF: @ramonmulia could you slightly update it to use `As === 'span'` to make this an exceptional case with a comment? If in the JS case someone sends an invalid value or null, we should render `<p>`, not `<span>`

cc: @Ademsk1

packages/react/src/paragraph/paragraph.tsx
ramonmulia: I'm not sure why "align" was removed from the Text component. I'm happy to fix it here if needed.

DmitryB-NF: I think `align` may not be required for the `Text` component. I think we can fix it in the Paragraph only and discuss it in a future improvement

@Ademsk1 could you check if we need the `align` for the `Text`?

ramonmulia: `align` is working correctly in atom Paragraph, there is also a snapshot test for coverage.

DmitryB-NF: But it will not work with the existing paragraph like
```
<Paragraph as="span" align="center">...
```
This will stop working after the lib upgrade

DmitryB-NF: @Ademsk1 I do not think we need to stop this PR from merging. Could you please fix the alignment issue separately within the `Text` component and its support here?

DmitryB-NF (approved):

packages/react/src/paragraph/paragraph.tsx
Ademsk1: Ah. In principle, `align` has only a visual effect on elements who's width is greater than the content, so it can move horizontally when aligned. However, the span element only takes up the width of its content. 
Before, when using `<Paragraph as="span" align="center"/>` would do nothing. Check out https://ds.services.gov.ie/storybook-react/?path=/docs/typography-paragraph--docs

However. I recognise now that a user **could** define the span size, and therefore align it I guess?

DmitryB-NF: @Ademsk1, that's a fair point, and we may clearly remove the default behaviour when the size is not set. I also see that there is no `width` prop to define the actual size by default, so the client will have to pass classes/styles, and `align` is not really useful.

The main issue is that this is a breaking change, even if it does not make sense for the isolated `Text` component. I'm fine proceeding without fixing it, but we need to ensure it won't cause issues for lib users.

PR #916, opened 2026-04-08
chore(AB#37102): strict ESLint rule

DmitryB-NF:
## Description

Fix ESLint errors.
Reformat all examples with the up-to-date ESLint and Prettier configs.

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

AB#37102


packages/react/src/select/select-next.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import {
   Children,
-  FC,
```
ramonmulia, line 4: why did only this file change?
I noticed that if we run `pnpm lint --fix` in react side, it changes a lot of files.

packages/react/eslint.config.js
```diff
@@ -14,4 +14,9 @@ export default defineConfig([
       'storybook/no-uninstalled-addons': 'off',
     },
   },
+  {
+    rules: {
+      'no-console': 'warn', // TODO: ESLINT remove this after error fixes
+    },
```
Ademsk1, line 20: remove? Or is this an in-progress thing ?

packages/react/src/select/select-next.tsx
Ademsk1: Also, you committed - so it must've passed the `husky`... Is the `husky` still working?

DmitryB-NF: @ramonmulia This file is changed to fix errors but not warnings. It still has a lot of warnings and can be fixed, but this is not a part of the PR

packages/react/eslint.config.js
DmitryB-NF: There are 20+ errors in multiple components in React, and fixing all existing errors is out of scope for the existing PR. So, I added a TODO to fix all the `console.logs` within the refactoring and Mitosis migration

ramonmulia (dismissed):

PR #917, opened 2026-04-08
fix(AB#37034): add align to legacy paragraph for as="span"

Ademsk1:
## Description

Adds in align to Text. Adds in a story to exemplify this. 

## Type of Issue

- [ ] 🚀 Feature
- [x] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#37034](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37034)


ramonmulia (dismissed):

DmitryB-NF (dismissed):
The original intent of this change was to ensure the existing React Paragraph component works without changes after the migration to Atoms. As mentioned in https://github.com/ogcio/govie-ds/pull/907#discussion_r3050509226, we only need to change one component to ensure it applies the proper styles and looks the same.

Overall, as it was mentioned, we do not need the `align` prop for `Text` because the `span` tag is adjusted to the content width and align has no effect

Ademsk1:
> The original intent of this change was to ensure the existing React Paragraph component works without changes after the migration to Atoms. As mentioned in [#907 (comment)](https://github.com/ogcio/govie-ds/pull/907#discussion_r3050509226), we only need to change one component to ensure it applies the proper styles and looks the same.
> 
> Overall, as it was mentioned, we do not need the `align` prop for `Text` because the `span` tag is adjusted to the content width and align has no effect



> The original intent of this change was to ensure the existing React Paragraph component works without changes after the migration to Atoms. As mentioned in [#907 (comment)](https://github.com/ogcio/govie-ds/pull/907#discussion_r3050509226), we only need to change one component to ensure it applies the proper styles and looks the same.
> 
> Overall, as it was mentioned, we do not need the `align` prop for `Text` because the `span` tag is adjusted to the content width and align has no effect

I've renamed the PR, and made the changes in `Paragraph` such that in the very rare case that a user has passes in `display: block` and set a width and used the align, an align **can** work.

packages/angular/src/atoms/storybook/Text.meta.ts
```diff
@@ -9,7 +9,8 @@ export const textMeta = {
     id: 'text-default',
     dataTestId: 'text-default',
     size: Size.MD,
-    whitespace: Whitespace.NORMAL
+    whitespace: Whitespace.NORMAL,
+    align: Align.START
```
Ademsk1: will remove

packages/angular/src/stories/Text.stories.ts
```diff
@@ -15,8 +15,8 @@ type Story = StoryObj<Text & { content: string }>;
 
 export const Default: Story = {
   ...TextStoryMeta.Default,
-  render: (args: any) => ({
-    props: { ...args, content: loremIpsum },
+  render: (arguments_: any) => ({
+    props: { ...arguments_, content: loremIpsum },
```
Ademsk1: adjusted this because of spelling only

packages/core/atoms/utilities.ts
```diff
@@ -11,3 +11,19 @@ export const Whitespace = {
...
 } as const;
+
+export const Align = {
+  START: 'start',
+  CENTER: 'center',
+  END: 'end',
+  JUSTIFY: 'justify',
+} as const;
+
+export const getSize = (x: (typeof Size)[keyof typeof Size] = Size.MD) =>
```
Ademsk1, line 22: These utils will be helpful in the future - so exporting them from here

packages/angular/src/stories/Text.stories.ts
DmitryB-NF: Name it `props` to fix ESLint and highlight the actual purpose

cc: @ramonmulia

packages/core/atoms/storybook/Text.meta.ts
```diff
@@ -90,9 +90,9 @@ export const Default = {
   args: textMeta.args,
   play: async ({ canvasElement }: StoryContext<Renderer>) => {
     const canvas = within(canvasElement as HTMLElement);
-    const el = canvas.getByTestId('text-default');
-    expect(el).toBeInTheDocument();
-    expect(el.tagName).toBe('SPAN');
+    const element = canvas.getByTestId('text-default');
```
DmitryB-NF: 💡 The main idea of this ESLint rule is to use more reasonable names. You can name it `text` or `textNode ` to make it clear that this is the text element to be checked

packages/react/src/paragraph/paragraph.tsx
```diff
@@ -30,11 +32,12 @@ export function Paragraph({
   ariaHidden,
 }: ParagraphProps) {
   if (As === 'span') {
+    const alignStyle = getAlignStyle(align);
```
DmitryB-NF: variable can be inlined

packages/react/src/paragraph/paragraph.tsx
```diff
@@ -59,3 +62,15 @@ export function Paragraph({
     </GiParagraph>
   );
 }
+
+const getAlignStyle = (align: (typeof Align)[keyof typeof Align]) => {
```
DmitryB-NF: ⚠️ tailwind-variants for consistency?

pnpm-lock.yaml
```diff
@@ -500,6 +500,18 @@ importers:
         specifier: ^5.9.3
         version: 5.9.3
 
+  packages/angular/dist:
```
DmitryB-NF, line 500: This seems unrelated to the PR, especially if modules are not changed. Seems like a pnpm inconsistency

DmitryB-NF (changes requested):
Extraction of helpers is a good improvement!

packages/angular/src/stories/Text.stories.ts
```diff
@@ -35,8 +35,8 @@ export const Default: Story = {
 
 export const AllTextSizes: Story = {
   ...TextStoryMeta.AllTextSizes,
-  render: (args: any) => ({
-    props: { ...args },
+  render: (arguments_: any) => ({
+    props: { ...arguments_ },
```
DmitryB-NF:
```suggestion
  render: (props: Text) => ({
    props,
```

packages/angular/src/stories/Text.stories.ts
```diff
@@ -35,8 +35,8 @@ export const Default: Story = {
 
 export const AllTextSizes: Story = {
   ...TextStoryMeta.AllTextSizes,
-  render: (args: any) => ({
-    props: { ...args },
+  render: (props: any) => ({
+    props: { ...props },
```
DmitryB-NF: Just in case, we do not need to copy the object

```suggestion
  render: (props: any) => ({
    props,
```

packages/react/src/atoms/Paragraph.tsx
```diff
@@ -19,45 +19,23 @@ export type Props = {
...
-  START: 'start',
-  CENTER: 'center',
-  END: 'end',
-  JUSTIFY: 'justify',
-} as const;
+import { getSize, getAlign, getWhitespace, Align, Size, Whitespace } from './utilities';
+import { textVariants } from './Text';
 export const paragraphStyles = tv({
-  base: ['gi-font-primary', 'gi-max-w-prose'],
+  extend: textVariants,
```
DmitryB-NF, line 25: This looks much better now 🚀 

cc: @ramonmulia it could be nice feature to use later on

DmitryB-NF (dismissed):
Look much cleaner now!

packages/react/src/atoms/Paragraph.tsx
ramonmulia: yeap, it is a good idea, it will simplify our styles

DmitryB-NF (approved):
Well done!
TV extend is a really good feature to have

PR #918, opened 2026-04-08
chore(AB#36665): Adding eslint-react-hooks plugin

ramonmulia:
## Description

Adding eslint-react-hooks plugin + fixes

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[36665](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36665)

### DIFF
--- a/packages/react/eslint.config.js
+++ b/packages/react/eslint.config.js
@@ -10,7 +10,7 @@ export default defineConfig([
   ...storybook.configs['flat/recommended'],
   ...baseConfig,
   {
-    files: ['**/*.tsx'],
+    files: ['src/**/*.{js,jsx,ts,tsx}'],
     plugins: {
       'react-hooks': reactHooks,
     },
@@ -29,5 +29,5 @@ export default defineConfig([
     rules: {
       'no-console': 'warn', // TODO: ESLINT remove this after error fixes
     },
-  }
+  },
 ]);
--- a/packages/react/eslint.config.js
+++ b/packages/react/eslint.config.js
@@ -1,6 +1,7 @@
 import { defineConfig } from 'eslint/config';
 import baseConfig from '../../eslint.config.mjs';
 import storybook from 'eslint-plugin-storybook';
+import reactHooks from 'eslint-plugin-react-hooks';
 
 export default defineConfig([
   {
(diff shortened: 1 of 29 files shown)


packages/react/src/autocomplete/autocomplete.stories.tsx
```diff
@@ -322,29 +322,31 @@ export const WithLoading = () => {
...
+        </Autocomplete>
+      </FormField>
+      <Label>Watched value: {topicValue}</Label>
+    </div>
+  );
+};
+
+export const WithReactHookForm: StoryObj = {
+  tags: ['skip-playwright'],
+  render: () => <WithReactHookFormRender />,
```
DmitryB-NF: ⚠️ to fix it in the suggested Storybook way, it is better to use a named function like
```
render: function Render () {
   useForm()
}
```
This way, it will have no warnings with minimal changes

ramonmulia: much better! thanks for the suggestion.

packages/react/eslint.config.js
```diff
@@ -1,13 +1,24 @@
...
+import reactHooks from 'eslint-plugin-react-hooks';
 
 export default defineConfig([
   {
     ignores: ['src/atoms/**/*'],
   },
   ...storybook.configs['flat/recommended'],
   ...baseConfig,
+  {
+    files: ['**/*.tsx'],
```
DmitryB-NF: ⚠️ custom hooks could be in `ts` files, so it is better to include all js/ts files

DmitryB-NF (changes requested):
Fixes look good but we need small config adjustment

packages/react/eslint.config.js
```diff
@@ -1,13 +1,24 @@
...
+import reactHooks from 'eslint-plugin-react-hooks';
 
 export default defineConfig([
   {
     ignores: ['src/atoms/**/*'],
   },
   ...storybook.configs['flat/recommended'],
   ...baseConfig,
+  {
+    files: ['**/*.{tsx,js,ts}'],
```
DmitryB-NF: `JSX` for consistency, even if it is not used? I think we can use `ts,tsx` only or add `js,jsx` to be fully consistent.
Just in case, we may need to use it like `src/**/*` to narrow down the scope and speed up the processing

DmitryB-NF (approved):
Well done!

PR #919, opened 2026-04-09
feat(AB#36657): Adding Button.meta file + updating Button stories 

ramonmulia:
## Description

Adding Button.meta file + updating Button stories 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues
[36657](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36657)


packages/core/atoms/storybook/Button.meta.ts
```diff
@@ -0,0 +1,259 @@
...
+    },
+    variant: {
+      control: { type: 'select' },
+      options: Object.values(Variant),
+      description: 'Visual style variant of the button.',
+      table: {
+        type: {
+          summary: Object.values(Variant)
+            .map((v) => `"${v}"`)
+            .join(' | '),
```
DmitryB-NF: It seems like creating helper for similar values extrapalation would be nice across all Atom storybooks

packages/core/atoms/storybook/Button.meta.ts
```diff
@@ -0,0 +1,259 @@
...
+    onKeyDown: {
+      control: false,
+      description: 'Keydown event handler.',
+      table: { type: { summary: '(event: any) => void' } },
+    },
+    onKeyUp: {
+      control: false,
+      description: 'Keyup event handler.',
+      table: { type: { summary: '(event: any) => void' } },
+    },
```
DmitryB-NF: ⚠️ For `Angular`, there is a default set of props like `(click)`, `(focus)`, where `onClick` is only an internal one. I think we need this declaration for React only to make sure in Angular the native handlers are used

packages/react/src/stories/Button.stories.tsx
```diff
@@ -13,563 +22,156 @@ const meta = {
...
-      <Button dataTestId="govie-button-disabled" disabled>
-        Button
-      </Button>
-      <Button dataTestId="govie-button-icon">
-        <Icon icon="thumb_up" />
-      </Button>
+    <div className="gi-flex gi-flex-col gi-gap-4 gi-items-start">
+      {Object.values(Variant).map((variant) =>
+        Object.values(Appearance).map((appearance) =>
+          appearance === Appearance.LIGHT ? (
```
DmitryB-NF: In React, huge ternary operators could decrease readability. It is better to use it like 
```
{appearance === Appearance.LIGHT && (<>)}

{appearance === Appearance.DARK && (<>)}
```

packages/react/src/stories/Button.stories.tsx
```diff
@@ -13,563 +22,156 @@ const meta = {
...
+            <div
```
DmitryB-NF: Does it make sense to include a disabled prop to test whether it is properly processed as well?

packages/angular/src/stories/Button.stories.ts
```diff
@@ -1,60 +1,149 @@
...
 };
 
-export const Disabled: Story = {
+export const InteractionStates: StoryObj = {
+  ...buttonInteractionStates,
   render: () => ({
-    template: `<gi-button variant="primary" appearance="default" size="medium" [disabled]="true">Click me</gi-button>`,
+    moduleMetadata: { imports: [Button] },
+    template: `
+      <div class="gi-flex gi-flex-col gi-gap-4 gi-items-start">
```
DmitryB-NF, line 147: Check disabled?

DmitryB-NF (changes requested):
The Storybook looks good!
We need to make sure Angular uses native props and cover disable state for testing purposes (to make sure it is not broken)

packages/core/atoms/storybook/Button.meta.ts
ramonmulia: good call! I was planning to do something like that in IconButton, I'm going to update here instead.

Ademsk1 (dismissed):
LGTM!

packages/core/atoms/storybook/storybook-utilities.ts
```diff
@@ -0,0 +1,35 @@
+import type { InputType } from 'storybook/internal/types';
```
DmitryB-NF: 💡 You can name the file `utilities` or even `utils` because it is clear that it is related to Storybook from the export path `atoms/storybook/utils`

packages/core/atoms/storybook/storybook-utilities.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Converts an `as const` enum object into a Storybook argType with a select
+ * control, options list, and a formatted type summary for the docs table.
+ */
+export function enumArgumentType<const T extends Readonly<Record<string, string>>>(
+  enumObject: T,
+  description: string,
+  defaultValue?: T[keyof T],
```
DmitryB-NF: To make it expandable, I would suggest adding the second parameter as an `options` map with all extensions to use it like 
```
enumType<Size>(Size, {
  description: 'Optional String', // can be populated externally with `...`
  default: 'optional',
  table: { extraOption: 'optional' }
})
```
This way, by using `_.defaultsDeep`, you can make it fully expandable, and all parameters have clear names

packages/core/atoms/storybook/storybook-utilities.ts
```diff
@@ -0,0 +1,35 @@
+import type { InputType } from 'storybook/internal/types';
+
+/**
+ * Converts an `as const` enum object into a Storybook argType with a select
+ * control, options list, and a formatted type summary for the docs table.
+ */
+export function enumArgumentType<const T extends Readonly<Record<string, string>>>(
```
DmitryB-NF: `enumType` could also work

packages/core/atoms/storybook/Text.meta.ts
```diff
@@ -17,32 +18,8 @@ export const textMeta = {
...
-      table: {
-        type: {
-          summary: Object.values(Whitespace)
-            .map((v) => `"${v}"`)
-            .join(' | '),
-        },
-        defaultValue: { summary: 'normal' },
-      },
-    },
+    size: enumArgumentType(Size, 'Font size of the Text', 'md'),
```
DmitryB-NF: ⚠️ Better to use `Size.MD` as a default value

packages/core/atoms/storybook/Heading.meta.ts
```diff
@@ -10,16 +11,7 @@ export const headingMeta = {
...
-      options: ['', ...Object.values(Size)],
-      description: 'Font size of the heading. Defaults to the size mapped to the heading level.',
-      table: {
-        type: {
-          summary: Object.values(Size)
-            .map((v) => `"${v}"`)
-            .join(' | '),
-        },
-      },
+      ...enumArgumentType(Size, 'Font size of the heading. Defaults to the size mapped to the heading level.'),
```
DmitryB-NF: Spread is excessive 
```suggestion
      size: enumArgumentType(Size, 'Font size of the heading. Defaults to the size mapped to the heading level.'),
```

packages/core/atoms/storybook/Paragraph.meta.ts
```diff
@@ -15,51 +16,9 @@ export const paragraphMeta = {
...
-          summary: Object.values(Whitespace)
-            .map((v) => `"${v}"`)
-            .join(' | '),
-        },
-        defaultValue: {
-          summary: 'normal',
-        },
-      },
-    },
+    size: enumArgumentType(Size, 'Font size of the Paragraph.', 'md'),
```
DmitryB-NF: ⚠️ `Size.MD`

packages/react/src/stories/Button.stories.tsx
```diff
@@ -13,563 +51,128 @@ const meta = {
...
-    disabled: {
-      control: 'boolean',
-      description: 'Specify if the button is disabled',
-      type: 'boolean',
-    },
-  },
+  ...buttonDefault,
   args: {
+    ...buttonDefault.args,
     children: 'Button',
```
DmitryB-NF: `children` prop is already in `args` https://github.com/ogcio/govie-ds/pull/919/changes#diff-2dbc4ef132ff44f183f4245731befa502750c34b7eff76fe967d9c876b18f0fcR11
So, `args: { ... }` can be removed

packages/react/src/stories/Button.stories.tsx
```diff
@@ -13,563 +51,128 @@ const meta = {
...
-    variant: 'flat',
-    appearance: 'dark',
-    size: 'medium',
-    disabled: true,
-  },
+export const AllVariants: Story = {
+  ...buttonAllVariants,
+  render: () => (
+    <div className="gi-flex gi-flex-col gi-gap-2 gi-items-start">
+      {Object.values(Variant).map((variant) => (
```
DmitryB-NF: Just in case, you can use `lodash` here to simplify the code

```suggestion
      _.map(Variant, (variant) => (
```

packages/react/src/stories/Button.stories.tsx
```diff
@@ -13,563 +51,128 @@ const meta = {
...
-      <Button dataTestId="govie-button-disabled" disabled>
-        Button
-      </Button>
-      <Button dataTestId="govie-button-icon">
-        <Icon icon="thumb_up" />
-      </Button>
+    <div className="gi-flex gi-flex-col gi-gap-4 gi-items-start">
+      {Object.values(Variant).map((variant) =>
+        Object.values(Appearance).map((appearance) => (
+          <div
```
DmitryB-NF, line 166: Disabled buttons with the interaction?

DmitryB-NF (changes requested):
Usage of the `enum` simplifies the code really much 🚀 
We need to clean up the code a bit, and we are ready to go!

**Note:** I have unresolved earlier comments about the disabled button interaction that were not updated in the code. Could you look at them too?

packages/react/src/stories/Button.stories.tsx
ramonmulia: I don't think it's necessary, it would duplicate coverage that already exists in the Disabled story. InteractionStates is specifically for :hover and :focus pseudo-states, which disabled doesn't have.

DmitryB-NF: The point is to make sure that all combinations of Appearance, Variant, and Disable work consistently, and we have tests for it. We can cover all the cases in the DIsabled section if it seems better. 
Btw, `:focus` and `:hover` ideally should be tested too, even if passing disabled prop to make sure the button is not interactive, this is also a part of the test coverage

ramonmulia: oh, I see your point.

packages/core/package.json
```diff
@@ -16,6 +16,7 @@
     "@builder.io/mitosis-cli": "^0.12.1",
     "@prettier/sync": "^0.6.1",
     "chokidar-cli": "^3.0.0",
+    "lodash": "^4.17.23",
```
DmitryB-NF, line 20: @ramonmulia, we may need to include `@types/lodash` to avoid TS warnings

packages/react/src/button/helpers.ts
```diff
@@ -1,5 +1,24 @@
...
+  sm: 'sm',
+  md: 'md',
+  lg: 'lg',
+};
+
+/**
+ * Normalises the wrapper's wide ButtonSize union into the atom's narrow
+ * size vocabulary expected by `<GiButton>`. Defaults to `'md'` when omitted.
+ */
+export const normalizeButtonSize = (size: ButtonSize = 'md'): AtomButtonSize =>
```
DmitryB-NF: 💡 Does it make sense to name it `normalizeSize` and move it out of the Button? This way, we can clean up any other components as well and make it more generic. 
Btw, `Size` map can be accessible directly from Atoms as it is also a generit type across all components

packages/react/src/card/card-legacy.tsx
```diff
@@ -60,9 +61,11 @@ export const CardLegacy = ({
         </Link>
       );
     }
+    const { size, ...buttonProps } = action;
     return (
       <Button
-        {...action}
+        {...buttonProps}
+        size={normalizeButtonSize(size)}
```
DmitryB-NF: Just in case, you can override the prop to avoid copying the object
```suggestion
        {...action}
        size={normalizeSize(action.size)}
```
This will override the prop anyway with the correct value

packages/react/src/modal/modal.tsx
```diff
@@ -43,21 +45,33 @@ const VARIANT_ORDER: Record<
   primary: 2,
 };
 
+const SIZE_TO_ICON_BUTTON_SIZE: Record<ButtonSize, IconButtonProps['size']> = {
```
DmitryB-NF, line 49: Just in case, I would suggest adding `TODO` comments to make sure it is removed when the `IconButton` is converted

DmitryB-NF (dismissed):
Standardisation of sizes looks incredibly good!
I propose a couple of small improvements, optionally to add

packages/core/atoms/Button.lite.tsx
```diff
@@ -346,7 +348,7 @@ export const styles = tv({
   defaultVariants: {
     variant: 'primary',
     appearance: 'default',
-    size: 'medium',
+    size: 'md',
```
DmitryB-NF, line 351: `Size.MD` would be preferable everwhere instaead of strings

DmitryB-NF (approved):

packages/core/atoms/Button.lite.tsx
DmitryB-NF: @ramonmulia, could you fix this in the next iteration, so as not to block the PR?

packages/core/atoms/Button.lite.tsx
```diff
@@ -355,4 +357,4 @@ const getVariant = (x: Props['variant'] = Variant.PRIMARY) =>
   Object.values(Variant).includes(x) ? x : Variant.PRIMARY;
 const getAppearance = (x: Props['appearance']) =>
   x === Appearance.LIGHT || x === Appearance.DARK ? x : Appearance.DEFAULT;
-const getSize = (x: Props['size']) => (x === Size.SMALL || x === Size.LARGE ? x : Size.MEDIUM);
```
DmitryB-NF, line 358: Btw, this would also work as before as a proper subset

PR #922, opened 2026-04-13
chore(AB#36572): react dep audit

Ademsk1:
## Description

Outcome of audit:
1. Remove unused dependencies
2. Relocate deps that belong in devdeps
3. Remove storybook & clsx from dist out file 
4. Move generic deps to root (e.g. storybook plugins) 

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#36572](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36572)


Ademsk1:
@ramonmulia @DmitryB-NF  I'd like your advice on how to proceed with `i18n`. To me, it looks like we use `i18next` as a devDep and `i18next-react` as a regular dependency. However if you check out `packages/react/src/i18n` you can see that we export the dev dependency as a real one to our dist! This creates some out files to have i18next logic. I'd recommend we have both of these as regular dependencies. 

How would you like to proceed in this instance? Does i18next even make sense to export? Should we deprecate this export and add it as a direct dependency?

packages/react/src/accordion/accordion-item.tsx
```diff
@@ -1,6 +1,7 @@
 'use client';
-import clsx from 'clsx';
```
DmitryB-NF, line 2: ⚠️ This change does not seem related to dependency updates. Is this intentional?

Overall, `cn` is an excessive function, and it is better to use `clsx` directly to avoid additional array generation during runtime.

**Note:** that `docs` uses `twMerge` to clean up Tailwind class names and merge them properly. We may need to revise this approach, but so far, usage of `clsx` is preferable

packages/react/src/browser-support/browser-support.tsx
```diff
@@ -1,7 +1,8 @@
 'use client';
 import Bowser from 'bowser';
-import clsx from 'clsx';
```
DmitryB-NF, line 3: `clsx` is preferable https://github.com/ogcio/govie-ds/pull/922/changes#r3072947365

packages/react/package.json
```diff
@@ -60,26 +60,17 @@
...
     "@ogcio/design-system-prettier-config": "workspace:*",
     "@ogcio/design-system-tailwind": "workspace:*",
     "@ogcio/theme-doete": "workspace:*",
     "@ogcio/theme-govie": "workspace:*",
-    "@storybook/addon-a11y": "^10.3.3",
-    "@storybook/addon-coverage": "^3.0.1",
-    "@storybook/addon-docs": "^10.3.3",
-    "@storybook/addon-links": "^10.3.3",
-    "@storybook/addon-themes": "^10.3.3",
-    "@storybook/addon-vitest": "^10.3.3",
```
DmitryB-NF, line 77: ⚠️ Storybook setup uses all of those dependencies in `packages/react/.storybook/main.ts`. Storybook will not work properly without the ddirect dependency

package.json
```diff
@@ -66,6 +66,12 @@
     "@types/react-dom": "^19.0.4",
     "@types/tinycolor2": "^1.4.6",
     "@vitejs/plugin-react": "^4.6.0",
+    "@storybook/addon-a11y": "^10.3.3",
```
DmitryB-NF: ⚠️ Storybook addons should be a direct dependency of each package to make sure all packages are sufficient. This is basically the opposite of the task where we need to move all direct dependencies to each package

DmitryB-NF (dismissed):
@Ademsk1 Checking the moving of addons, we basically need the opposite to make sure all packages include all used dependencies (excluding Prettier, ESLint, and any global validators). Let me know if you need help to validate the task

⚠️ The `lodash` package is still not listed as a direct dependency, as requested in the ticket. I think we need more granular analysis

DmitryB-NF:
@Ademsk1, according to `react-i18next` and `i18n`, we need to make sure all dependencies are added correctly as `dependencies` if they are directly used in the code. This should be part of the task.

DmitryB-NF:
@Ademsk1 could you also make sure that the PR follows the naming convention the same as commits? I would also suggest changing it to `chore` because it should not change the behviour of the lib

packages/react/src/browser-support/browser-support.tsx
Ademsk1: clsx was being bundled into an out-file, despite its clear direct dependency - which looked a little clunky. By converting to using directly cn(), we remove that file. cn() is used abundantly within the react package, so seemed like there were no downsides and we remove a useless dist file.

packages/react/src/accordion/accordion-item.tsx
Ademsk1: See comment below

packages/react/src/browser-support/browser-support.tsx
DmitryB-NF: An additional `dist` file is not a problem for the build, and expected bundle practice (JS chunks will be loaded on demand). I would not say we need to optimise files and bundle chunks when code readability and performance are the primary goals. 

While the main goal of the PR is to clean up dependencies, I would rather move those changes into a separate PR/discussion to consistently clean up this.
Overall, it would be nice to remove `cn` entirely as redundant and use `clsx` directly.

Ademsk1:
Dependency list updated to reflect all the runtime required dependencies we use. 

Tested the changes locally by running `npm pack` on the newer build and installing it in a dummy website. The other dependencies will be installed on the users node_modules from now on. 

Size of dist has gone down now from 2.6MB to 1MB

packages/react/vite.config.ts
```diff
@@ -34,7 +36,8 @@ export default defineConfig({
           id === 'react' ||
           id === 'react-dom' ||
           id.startsWith('react/') ||
-          id.startsWith('react-dom/')
+          id.startsWith('react-dom/') ||
+          Object.keys(package_.dependencies).includes(id)
```
DmitryB-NF: ⚠️ This solution looks excessive. Having all dependencies properly in the `package.json` allows Vite to resolve the bundle and libs properly. I do not think we need such an aggressive approach

DmitryB-NF (dismissed):

packages/react/package.json
```diff
@@ -51,46 +51,58 @@
...
+    "tailwind-variants": "^0.2.1"
   },
   "devDependencies": {
     "@chromatic-com/storybook": "^5.1.1",
+    "@faker-js/faker": "^9.9.0",
     "@ogcio/design-system-prettier-config": "workspace:*",
     "@ogcio/design-system-tailwind": "workspace:*",
     "@ogcio/theme-doete": "workspace:*",
     "@ogcio/theme-govie": "workspace:*",
+    "@playwright/test": "^1.58.2",
```
DmitryB-NF: ⚠️ it should be only in the root folder as it is related to local tests and global Playwright config

packages/react/package.json
```diff
@@ -51,46 +51,58 @@
...
-    "@testing-library/user-event": "^14.5.2",
-    "@types/css-modules": "^1.0.5",
+    "@vitejs/plugin-react": "^4.6.0",
     "autoprefixer": "^10.4.19",
     "axe-core": "^4.10.0",
     "eslint-plugin-react-hooks": "^7.0.1",
     "eslint-plugin-storybook": "^10.2.19",
     "glob": "^11.0.1",
-    "i18next": "^25.6.2",
+    "playwright": "^1.58.2",
```
DmitryB-NF: The package should be not here and will be removed in https://github.com/ogcio/govie-ds/pull/927

packages/react/vite.config.ts
```diff
@@ -6,7 +6,6 @@ import preserveDirectives from 'rollup-preserve-directives';
 import { defineConfig } from 'vite';
 import dts from 'vite-plugin-dts';
 import { libInjectCss } from 'vite-plugin-lib-inject-css';
-
```
DmitryB-NF, line 9: nl?

DmitryB-NF (dismissed):
Changes look good! 
All dependencies are now in place, and we can migrate React to Vite 7/8 with up-to-date support 🎉 

Could you resolve a minor change and resolve merge conflicts?

DmitryB-NF (approved):

DmitryB-NF (approved):

PR #923, opened 2026-04-13
feat(AB#36573): Adding IconButton atom component

ramonmulia:
## Description

Adding IconButton atom component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[36573](dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36573)


packages/core/atoms/IconButton.lite.tsx
```diff
@@ -0,0 +1,337 @@
+import { tv } from 'tailwind-variants';
+import { useMetadata } from '@builder.io/mitosis';
+import { Variant, Appearance, ButtonSize } from './utilities';
+
+export const Size = {
+  ...ButtonSize,
+  EXTRA_LARGE: 'extraLarge',
+} as const;
```
Ademsk1: Why not re-use the Size variable in utilities?

packages/core/atoms/IconButton.lite.tsx
```diff
@@ -0,0 +1,337 @@
...
+    size: 'medium',
+    disabled: false,
+  },
+});
+
+const getVariant = (x: Props['variant'] = Variant.PRIMARY) =>
+  Object.values(Variant).includes(x) ? x : Variant.PRIMARY;
+const getAppearance = (x: Props['appearance']) =>
+  x === Appearance.LIGHT || x === Appearance.DARK ? x : Appearance.DEFAULT;
+const getSize = (x: Props['size'] = Size.MEDIUM) => (Object.values(Size).includes(x) ? x : Size.MEDIUM);
```
Ademsk1: Can now use getSize from utilities - exported. If you're happy with using Size from utilities that is..

ramonmulia: good call! but i believe we cant just use `Size` from utilities directly because if someone adds 2xl to the shared `Size` later, it would automatically leak into components that use it as their prop type.. so separating those subsets like ButtonSize and IconButtonSize, each component declares exactly what it supports.
Adding a new size to Size requires an explicit opt in per component.. so ts catches invalid sizes at compile time instead of silently ignoring them at runtime, right now IconButtonSize happens to match Size but that wont always be the case.

packages/core/atoms/IconButton.lite.tsx
```diff
@@ -0,0 +1,101 @@
...
+  tabIndex?: number;
+  dataTestId?: string;
+  ref?: any;
+};
+
+useMetadata({ angular: { selector: 'gi-icon-button' } });
+
+export default function IconButton(props: Props) {
+  return (
+    <button
```
DmitryB-NF, line 49: 💡 These look exactly the same as the props for the Button. Does it make sense to use `Button` here with some style overrides? We may not need exactly the same Atom with slightly different styles.

Could you check if we can reuse the `Button` here?

ramonmulia: good call, I looked into this, and we already have the base styles that are shared via 
```tv({ extend: buttonBaseStyles })```, so both components reuse the same visual base.
                                           
I also noticed that in angular wrapping ```Button``` inside ```IconButton``` Mitosis compiles it as a raw ```<button>``` in the angular template instead of resolving it to ```<gi-button>```. The generated output has [variant], [appearance], [size] bindings on a native ```<button>```, which don't exist.. so the Angular output is completely non-functional.        

Also IconButton uses fixed dimensions like : gi-h-10 gi-w-10 and Button uses padding based sizing liek: gi-px-3 gi-py-2 I think overriding from outside would be fragile.

Comment (DmitryB-NF) [reply packages/core/atoms/IconButton.lite.tsx:49]: For Angular, I think this is a problem with the import naming. When you import something like Button, Mitosis can use it as Button instead of GIButton, so maybe this is a matter of imports. It would be nice to try it to make sure we can avoid this limitation in the future.

For styles, I think override can work, but if this brings more complex styling and indirect overrides, new Atom actually makes sense. It would be nice to reuse the code as much as possible when the functionality has no trade-offs

File: apps/docs/content/3-components/2-library/icon-button/react.mdx [hunk 4]
```diff
@@ -110,178 +134,22 @@ An Icon Button is a button that uses an icon instead of text to represent an act
...
-</ComponentPreview>
-
-### Flat Light
-
-<ComponentPreview>
-  <StorybookFrame componentId="icon-button-react" story="flat-light" />
-</ComponentPreview>
-
-### Flat Light Hover
+### All Variants
```
DmitryB-NF, line 137: Does it make sense to name it "Variants"? Based on the documentation, it seems cear those are _all variants_

packages/react/src/stories/IconButton.stories.tsx
```diff
@@ -0,0 +1,225 @@
...
+  Default as defaultStory,
+  InteractionStates as interactionStates,
+  Disabled as disabledStory,
+  AllVariants as allVariants,
+  AllAppearances as allAppearances,
+  AllSizes as allSizes,
+} from '../atoms/storybook/IconButton.meta';
+import { Variant, Appearance } from '../atoms/utilities';
+import IconButton, {
+  IconButtonSize as Size,
```
DmitryB-NF, line 15: I am wondering if we can use `/utilities` `Size` as a `BaseSize` when importing, and have the `IconButtonSize` as `Size` to make it clear when importing. 

Basically, I would propose to use `Props['size']` as a typу and the `Size` (`BaseSize` from utilities as a set of values). We need to have `Enum` + `Type`, where the paradigm to avoid Enums in favour of constants makes this incredibly weird and confusing.

We may merge it as it is, but later on we should use only `/utilities` `Size` for values (as consts) and Props['size'] as a base for typings to make the code clean and efficient. `Size`, `ButtonSize`, `IconButtonSize` is too much

cc: @Ademsk1 @ramonmulia

packages/react/src/stories/IconButton.stories.tsx
```diff
@@ -0,0 +1,225 @@
...
+export default meta;
+
+type Story = StoryObj<typeof IconButton>;
+
+const iconSizeForButton: Record<string, number> = {
+  [Size.SM]: 16,
+  [Size.MD]: 24,
+  [Size.LG]: 24,
+  [Size.XL]: 32,
+};
```
DmitryB-NF, line 70: It seems like a good util map, also like `IconSize` to make it an actual benefit to map sizes into pixels

packages/react/src/table/table-data.tsx
```diff
@@ -108,9 +103,9 @@ export const TableExpandIcon = ({
   }, []);
 
   const sizeMap = {
-    sm: 'medium',
-    md: 'large',
-    lg: 'extraLarge',
+    sm: 'md',
```
DmitryB-NF, line 106: it would be really nice to use `Size` already but it could be later imporvement

DmitryB-NF (dismissed):
The implementation looks good 💪 ! 
I added a couple improvements but we may need to do that in the next iterrations

apps/docs/content/3-components/2-library/icon-button/react.mdx
ramonmulia: yes, I think that makes sense. I'm going to include these changes in future

packages/react/src/stories/IconButton.stories.tsx
ramonmulia: agreed

packages/react/src/table/table-data.tsx
ramonmulia: good call!

packages/core/atoms/IconButton.lite.tsx
ramonmulia: That’s correct, when using the proper GiButton, it renders as expected, the issue is that the Button styles are still affecting IconButton globally. Since Button applies padding (gi-px-3 gi-py-2), that ends up impacting the icon child as well.
I think it makes sense to create a follow up improvement task for this, that way, we can focus on fixing Button first and then reuse it properly in IconButton.

DmitryB-NF (approved):

PR #928, opened 2026-04-16
feat(AB#36951): Migrate container to mitosis

Ademsk1:
## Description

Migrate Container component over to Mitosis. Adds in story, and clsx for angular. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#26951](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36951)


packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ExtraLarge: 'xl',
+} as const;
+
+export const ContainerMaxWidthEnum = {
+  Small: 'sm',
+  Medium: 'md',
+  Large: 'lg',
+  ExtraLarge: 'xl',
+  Full: 'full',
+} as const;
```
DmitryB-NF: ⚠️ we should follow the same `Size` paradigm (`/core/atoms/utilities.ts`) for all components to be consistent. We can add `full` or `max` as an option, but all other sizes should be standardised

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export const ContainerGutterSizeEnum = {
```
DmitryB-NF: ⚠️ `Size` is preferable for consistency

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Container component when you need a centralised, consistent layout wrapper for content on your webpage.
+ * @param props - ContainerProps
+ * @param props.children - The content to be rendered inside the container.
+ * @param props.id - The id of the container.
+ * @param props.insetTop - The inset top size of the container. Default is `md`
+ * @param props.insetBottom - The inset bottom size of the container. Default is `md`
+ * @param props.className - The class name of the container.
+ * @param props.fullWidth - Whether the container should be full width. Default is `false`
+ * @param props.maxWidth - The max width of the container: `sm`, `md`, `lg`, `xl`, or `full`.
```
DmitryB-NF: ⚠️ Does it make sense to use only one prop `maxWidth` to make the container full-width only (default condition)? It seems the current implementation uses pretty much the same, where `gi-layout-container`,  `gi-layout-container-full-width`, and even `gi-layout-container-inset` have the full-width prop (**basically all conditions**)

Overall, Atom Container may not follow the existing React implementation, but it should be usable for new implementations. Most of the current usages assign class names directly without any props

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
+import { useMetadata } from '@builder.io/mitosis';
+import { tv } from 'tailwind-variants';
+import { getContainerLayoutType } from './utilities';
+
+export const ContainerInsetSizeEnum = {
```
DmitryB-NF: ⚠️ It is better to use simple names like `InsetSize`, `MaxWidth` to make it more usable in the code. It is still clear that this is related to the `Container` when importing it

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+  base: '',
+  variants: {
+    layout: {
+      standard: 'gi-w-full gi-max-w-[100vw] gi-px-4 md:gi-container md:gi-mx-auto',
+      fullWidth: 'gi-w-full gi-max-w-[100vw] gi-px-6 lg:gi-px-8 2xl:gi-max-w-screen-2xl 2xl:gi-mx-auto',
+      /** `.gi-layout-container-inset` — `hasInset` */
+      inset: 'gi-w-full gi-max-w-[100vw] gi-container gi-mx-auto',
+    },
+    gutterSize: {
+      [ContainerGutterSizeEnum.None]: '!gi-px-0',
```
DmitryB-NF: ⚠️ Is it possible to avoid `important` here? Otherwise, it will not be able to properly override it

packages/core/atoms/utilities.ts
```diff
@@ -27,3 +27,13 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
 
 export const getAlign = (x: (typeof Align)[keyof typeof Align] = Align.START) =>
   Object.values(Align).includes(x) ? x : Align.START;
+
+export function getContainerLayoutType(props: any): 'inset' | 'fullWidth' | 'standard' {
```
DmitryB-NF: ⚠️ It should not be a general util as it is used only for the `Container` and should be encapsulated there

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+ */
+export default function Container(props: Props) {
+  return (
+    <div
+      data-testid="govie-container"
+      className={styles({
+        layout: getContainerLayoutType(props),
+        maxWidth: props.maxWidth || ContainerMaxWidthEnum.Full,
+        gutterSize: props.gutterSize || ContainerGutterSizeEnum.Small,
+        ...(props.insetTop != null && { insetTop: props.insetTop }),
```
DmitryB-NF: You can always pass a value here like `insetTop: props.insetTop`. TV will pick only the correct value there with the full match (`===`)

packages/core/atoms/utilities.ts
```diff
@@ -27,3 +27,13 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+  if (props.insetTop || props.insetBottom) {
+    return 'inset';
+  }
+  if (props.fullWidth) {
+    return 'fullWidth';
+  }
```
DmitryB-NF: ❓ Do we really need this logic in a new Atom Container? I think we can make all props independent to make sure all styles can be assigned without intersections. It is always possible to use the "default" state of insets to apply only required classes instead of reassigning them later.

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,119 @@
...
+ * @param props.maxWidth - The max width of the container: `sm`, `md`, `lg`, `xl`, or `full`.
+ * @param props.gutterSize - The gutter size of the container. Default is `sm` (`gi-px-4`).
+ * @returns Container component
+ */
+export default function Container(props: Props) {
+  return (
+    <div
+      data-testid="govie-container"
+      className={styles({
+        layout: getContainerLayoutType(props),
```
DmitryB-NF: ⚠️ ultimately, we need to get rid of this approach to use a TV `compoundVariants` instead of manual classes. See `packages/core/atoms/Button.lite.tsx:146` as an example

DmitryB-NF (dismissed):
I think we need to revise the styles implementation to simplify/clean them and make them follow the TV practices.

The main goal is to create a new Atom Component that is flexible and extendable for future use. It may not cover all React implementation where we can use own wrapper for compatibility.

packages/core/atoms/Container.lite.tsx
Ademsk1: removed in the new implementation. defaults to max width of the screen.

packages/core/atoms/utilities.ts
Ademsk1: Good point.

packages/core/atoms/utilities.ts
Ademsk1: Removed

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+useMetadata({ angular: { selector: 'gi-container' } });
+
+/**
+ * Container component when you need a centralised, consistent layout wrapper for content on your webpage.
+ * @param props.id - The id of the container.
+ * @param props.children - The content to be rendered inside the container.
+ * @param props.className - The class name of the container.
+ * @param props.inset - Boolean to apply default inset padding of `md` to the container. Default is `false`.
```
DmitryB-NF: `md` is quite misleading here as there are progressive margins.
Also, the comment is removed during the generation, so the Storybok could be the primary source of data and better to have only one declaration to avoid mismatches

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,68 @@
...
+      {props.children}
+    </div>
+  );
+}
+
+export const styles = tv({
+  base: 'gi-w-full gi-max-w-[100vw] gi-container md:gi-mx-auto',
+  variants: {
+    inset: {
+      true: 'gi-container gi-mx-auto gi-py-4 md:gi-py-6 lg:gi-py-8',
```
DmitryB-NF: `gi-container` is not required as in in in the base already

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,68 @@
...
+      lg: 'gi-max-w-lg',
+      xl: 'gi-max-w-xl',
+      '2xl': 'gi-max-w-2xl',
+      full: 'gi-max-w-full',
+    },
+  },
+  defaultVariants: {
+    inset: false,
+    gutters: true,
+    maxWidth: 'full',
```
DmitryB-NF: ⚠️ no string values, emun only

packages/core/atoms/utilities.ts
```diff
@@ -19,6 +19,8 @@ export const Align = {
   JUSTIFY: 'justify',
 } as const;
 
+export const MaxWidth = { ...Size, '2xl': '2xl', full: 'full' } as const;
```
DmitryB-NF: ⚠️ should not be part of utils but encapsulated in the container

packages/core/atoms/utilities.ts
```diff
@@ -27,3 +29,12 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
 
 export const getAlign = (x: (typeof Align)[keyof typeof Align] = Align.START) =>
   Object.values(Align).includes(x) ? x : Align.START;
+
+export function getValidProp<T extends Record<string, string>>(
```
DmitryB-NF: 💡 nice function! I would call it `clamp` or similar to a short name and reuse it for all other utils

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,162 @@
+import type { StoryContext, Renderer } from 'storybook/internal/types';
+import { createElement } from 'react';
+import { within, expect } from 'storybook/test';
+import { MaxWidth } from '../utilities';
+
+/** Class fragment applied for each `maxWidth` variant (for interaction tests). */
+export function maxWidthClassToken(value: (typeof MaxWidth)[keyof typeof MaxWidth]): string {
```
DmitryB-NF: Can it be inlined into a single usage?

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,162 @@
...
+      const containerElement = canvas.getByTestId('govie-container');
+      const paragraphElement = canvas.getByText('Safe content');
+      expect(containerElement).toBeInTheDocument();
+      expect(paragraphElement).toBeInTheDocument();
+      expect(paragraphElement.innerHTML).toContain('Safe content');
+    });
+  },
+};
+
+export const TestHandleEmptyContentGracefully = {
```
DmitryB-NF: Better not to name it as Test because it will be visible to users. You can name it "Edge Case" or safety and add a tag to hide this by default. So, the Storybook will be clean and include all requeired tests

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,162 @@
...
+      expect(paragraphElement).toBeInTheDocument();
+      expect(paragraphElement.tagName).toBe('P');
+    });
+  },
+};
+
+export const TestSafelyRenderHTMLContent = {
+  tags: ['skip-playwright'],
+  args: {
+    children: createElement('p', null, createElement('script', null, "alert('XSS')"), 'Safe content'),
```
DmitryB-NF: ⚠️ This does not seem like a proper test for the Container. React escapes content by default and does not run embedded scripts.

packages/react/src/index.ts
```diff
@@ -61,7 +61,10 @@ export {
   type ComboboxProps,
   type DropdownItemProps,
 } from './combo-box/types.js';
-export { Container } from './container/container.js';
+export {
+  default as Container,
+  type Props as ContainerProps,
+} from './atoms/Container.js';
```
DmitryB-NF: ⚠️ This will basically remove the access to the actual container, which is a **breaking change** (the same as removing it entirely). We need to provide both or better to adopt wrapper the same way as all other components

DmitryB-NF (commented):
I think we need to revise the styles implementation to simplify/clean them and make them follow the TV practices.

The main goal is to create a new Atom Component that is flexible and extendable for future use. It may not cover all React implementations where we can use our own wrapper for compatibility.

packages/react/src/index.ts
Ademsk1: Good call

packages/react/src/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,143 @@
...
+    inset: {
+      control: 'boolean',
+      description: 'When `true`, applies default vertical inset padding (responsive `py` scale). Default is `false`.'
+    },
+    gutters: {
+      control: 'boolean',
+      description: 'When `true`, applies horizontal gutter padding (responsive `px` scale). Default is `true`.'
+    },
+    maxWidth: {
+      control: {
```
ramonmulia: Use the `enumType` helper from atoms/storybook/utilities here. See `Button.meta` for a reference example.

packages/react/src/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,143 @@
...
+  tags: ['autodocs'] as string[],
+  title: 'Layout/Container',
+  args: {
+    children: 'Paragraph',
+    gutters: true,
+    inset: false,
+    maxWidth: MaxWidth.full,
+    dataTestId: 'govie-container'
+  },
+  argTypes: {
```
ramonmulia, line 15: Add `table.type.summary` to all argTypes. This helps Angular identify the correct types for every argType, keeping them consistent between React and Angular.  See `Button.meta` for a reference example.

packages/angular/src/atoms/utilities.ts
```diff
@@ -16,6 +16,7 @@ export const Align = {
   END: 'end',
   JUSTIFY: 'justify'
 } as const;
-export const getSize = (x: (typeof Size)[keyof typeof Size] = Size.MD) => Object.values(Size).includes(x) ? x : Size.MD;
-export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] = Whitespace.NORMAL) => Object.values(Whitespace).includes(x) ? x : Whitespace.NORMAL;
-export const getAlign = (x: (typeof Align)[keyof typeof Align] = Align.START) => Object.values(Align).includes(x) ? x : Align.START
\ No newline at end of file
+export function clamp<T extends Record<string, string>>(value: string | undefined, options: T, defaultValue: T[keyof T]): T[keyof T] {
```
ramonmulia, line 32: great!

packages/react/src/container/container.tsx
```diff
@@ -12,10 +14,18 @@ export type ContainerInsetSizeType =
...
   id?: string;
+  /** @deprecated Use the `inset` prop on the layout container instead. */
   insetTop?: ContainerInsetSizeType;
+  /** @deprecated Use the `inset` prop on the layout container instead. */
   insetBottom?: ContainerInsetSizeType;
   className?: string;
+  /** @deprecated Use `maxWidth` (for example `'full'`) instead. */
   fullWidth?: boolean;
+  // Props for the new container
+  inset?: boolean;
```
ramonmulia: I think it would be better to import AtomContainerProps and extend that type instead of redefining the props here.

for example:
`type ContainerProps = GiContainerProps & { ... }`

This would help keep the types aligned and avoid duplicating them.

packages/react/src/container/container.tsx
```diff
@@ -24,22 +34,36 @@ export function Container({
...
 }: ContainerProps) {
-  const hasInset = insetTop || insetBottom;
+  const legacyProps = insetTop || insetBottom;
+  if (legacyProps) {
+    return (
+      <div
+        data-testid="govie-container"
+        data-inset-top={insetTop}
+        data-inset-bottom={insetBottom}
+        className={cn(className, 'gi-layout-container-inset')}
```
ramonmulia: It looks like `gi-layout-container-inset` comes from layout.css. Should we replace it with tailwind classes instead?

packages/react/src/stories/Container.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const HandlesEmptyContentGracefully: Story = {
+  ...handlesEmptyContentGracefully,
+};
+
+export const AllMaxWidths: Story = {
+  ...allMaxWidths,
+  render: () => (
+    <div className="gi-flex gi-flex-col gi-gap-8">
+      {Object.values(MaxWidth).map((maxWidth) => (
```
ramonmulia: use lodash `map` here, check react/stories/Button.stories

packages/react/src/container/container.tsx
Ademsk1: Because it's attached to the data-inset-bottom and data-inset-top, I decided not to change this here. Given the tailwind migration is to do with compaitibility with mitosis components, and we've already taken care of it, I decided to leave this here for now.

Ademsk1: cc @DmitryB-NF

DmitryB-NF: @Ademsk1 It would be nice to migrate this to TV variants. This will be a quick fix here because we will still need to inline this anyway in one of the next tickets.
Overall, the goal is to get rid of predefined classes altogether to partially resolve issues with `important` and make sure we can easily migrate to TV4

packages/core/atoms/Container.lite.tsx
DmitryB-NF: It would be nice to remove the entire comment to avoid mismatches with Storybook

packages/react/src/container/container.tsx
DmitryB-NF: As a preferable solution is to use GiContainer all the time but add extra styles when legacy props are passed to remove them easily later and inline all styles right away

packages/react/src/cookie-banner/cookie-banner.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import React from 'react';
 import type { ButtonProps } from '../button/types.js';
-import { Container } from '../container/container.js';
+import GiContainer from '../atoms/Container.js';
```
DmitryB-NF: Let's name it `Container`, as it should be. We should not mix both components together, so Container is a preferable name

apps/docs/content/3-components/2-library/container/react.mdx
```diff
@@ -11,36 +11,46 @@ component:
       fields:
         - name: 'children'
           ofType: 'node'
-          description: 'A React node representing the content to be rendered inside the container.'
-          required: true
+          description: 'Content rendered inside the container.'
+          required: false
```
DmitryB-NF: ⚠️ should it be still true as the content should be in the container?

examples/nextjs/src/app/page.tsx
```diff
@@ -18,7 +18,7 @@ import {
   Card,
   Chip,
   Combobox,
-  Container,
+  GiContainer,
```
DmitryB-NF: ⚠️ Should be Container as a single API for React lib

packages/core/atoms/Container.lite.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default function Container(props: Props) {
+  return (
+    <div
+      id={props.id}
+      data-testid={props.dataTestId}
+      class={styles({
+        inset: props.inset ?? false,
+        gutters: props.gutters ?? true,
+        maxWidth: clamp(props.maxWidth, MaxWidth, MaxWidth.screen),
```
DmitryB-NF: ⚠️ `MaxWidth.screen` is not mentioned in the Storybook as a list of variants, and `full` is set as the default in Docs. I am not sure we need `screen` at all because container should avod being bigger that its container

DmitryB-NF: Just in case, does it make sense to set it to 2XL to ensure the screen doesn't grow infinitely on larger screens? It is basically the main reason to use a container to make sure it has limited width that does not allow the content to flex too much

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const Default = {
+  args: containerMeta.args,
+  play: async ({ canvasElement, step }: StoryContext<Renderer>) => {
+    const canvas = within(canvasElement as HTMLElement);
+
+    await step('should apply the correct container classes', async () => {
+      const containerElement = canvas.getByTestId('govie-container');
+      expect(containerElement).toBeInTheDocument();
+      expect(containerElement.className).toContain('gi-w-full');
```
DmitryB-NF: ⚠️ We should avoid such tests because styling can be changed with the same behaviour. Tests should be stable

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,142 @@
...
+  },
+  play: async ({ canvasElement, step }: StoryContext<Renderer>) => {
+    const canvas = within(canvasElement as HTMLElement);
+
+    await step('should correctly handle and render indented HTML content', async () => {
+      const containerElement = canvas.getByTestId('govie-container');
+      const paragraphElement = canvas.getByText('Indented content');
+      expect(containerElement).toBeInTheDocument();
+      expect(paragraphElement).toBeInTheDocument();
+      expect(paragraphElement.tagName).toBe('P');
```
DmitryB-NF: better check by `TestId`

packages/core/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,142 @@
...
+  play: async ({ canvasElement, step }: StoryContext<Renderer>) => {
+    const canvas = within(canvasElement as HTMLElement);
+
+    await step('renders one container per max width', async () => {
+      const elements = canvas.getAllByTestId('govie-container');
+      const widths = Object.values(MaxWidth);
+      expect(elements).toHaveLength(widths.length);
+      for (const maxWidth of widths) {
+        const token = maxWidth === 'screen' ? 'gi-max-w-[100vw]' : `gi-max-w-${maxWidth}`;
+        const element = elements.find((node) => node.className.includes(token));
```
DmitryB-NF: ⚠️ no class name testing. It should check behaviour, not implementation

packages/react/src/container/container.stories.tsx
```diff
@@ -3,12 +3,13 @@ import { within, expect } from 'storybook/test';
 import { Container, ContainerInsetSizeEnum } from './container.js';
 
 const meta = {
-  title: 'Layout/Container',
+  title: 'Layout/Container/Deprecated',
```
DmitryB-NF: ⚠️ There should be a single public `Container` component, and a single Storybook. Could you check how Button storybook is implemented to have a combined approach without breaking changes.

packages/react/src/container/container.tsx
```diff
@@ -10,36 +30,50 @@ export const ContainerInsetSizeEnum = {
...
+  inset,
+  gutters,
+  maxWidth,
+  dataTestId,
 }: ContainerProps) {
-  const hasInset = insetTop || insetBottom;
+  const legacyProps = insetTop || insetBottom;
+  if (legacyProps) {
+    return (
+      <div
```
DmitryB-NF: Use Container with style overrides?

packages/react/src/footer/footer.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { renderToStaticMarkup } from 'react-dom/server';
 import {
   LogoGoldGreen as GovieLogoHarpWithText,
   LogoHarpWhite as GovieLogoHarp,
 } from '../atoms/icons/logos';
 import { cn } from '../cn.js';
-import { LogoProps } from '../common/types.js';
-import { Container } from '../container/container.js';
+import type { LogoProps } from '../common/types.js';
+import GiContainer from '../atoms/Container.js';
```
DmitryB-NF: It should be `Container`. We should use a single naming approach

packages/react/src/phase-banner/phase-banner.tsx
```diff
@@ -41,11 +41,17 @@ export function PhaseBanner({
   );
 
   if (wrap === 'container') {
-    return <Container>{bannerContent}</Container>;
+    return (
+      <GiContainer dataTestId="govie-container">{bannerContent}</GiContainer>
```
DmitryB-NF: ⚠️ Is this a default `dataTestId`? I am wondering if we actually need the default value, and also one in the other components

packages/react/tests/visual.spec.ts-snapshots/application-toast--all-positions-iphone-linux.png
DmitryB-NF, line 1: ⚠️ flaky test result

examples/nextjs/src/app/page.tsx
DmitryB-NF: ⚠️ Btw, there is no `GiContainer` exported from the lib

DmitryB-NF (dismissed):

apps/docs/content/3-components/2-library/container/react.mdx
Ademsk1: I don't think so - the <Container/> element being rendered is not dependent on children being within it.  We also state in props that children is an optional prop.

examples/nextjs/src/app/page.tsx
Ademsk1: Gonna fix this, thank you!

packages/core/atoms/Container.lite.tsx
Ademsk1: The reason for adding the `screen` as default is because the header would break in the storybook for mobile view - effectively. This was also in the original implementation. Defaulting to 2xl would break for mobile view for the storybook as well...

Happy to chat with you about this issue tomorrow in-depth. 

On the storybook side, I'll address that.

packages/react/src/phase-banner/phase-banner.tsx
Ademsk1: This was originally the default dataTestId for the <Container/> component. By adding this dataTestId here, we are not making any changes to consumers code inadvertently (e.g. if they test to find govie-container)

apps/docs/content/3-components/2-library/container/react.mdx
DmitryB-NF: Storybook can reflect not only the code state but the concept as well. Most React components can work without children (only a few use mapping or nested functions). But the `Container` is not _supposed_ to be used as a component without children. The main idea is to wrap the tree of components, and it should be reflected in the doc.
Even if the component is generally safe and can process invalid parameters/edge cases, it does not mean the edge case is supported by design.

examples/nextjs/src/app/page.tsx
DmitryB-NF: ⚠️ still not fixed in all files

packages/core/atoms/Container.lite.tsx
DmitryB-NF: I don't think this holds up — max-width only caps the upper bound, it can't force an element wider than its parent. On a 375px viewport, max-w-2xl (672px) and max-w-screen (100vw ≈ 375px) resolve to the same rendered width — the element still takes whatever the parent allows, which is ≤ viewport. Max-width ≥ viewport is effectively a no-op.

The reason not to use the `screen` is that 100vw is actively harmful on desktop: 100vw includes the vertical scrollbar width (~15–17px on Windows/Linux with classic scrollbars), so a max-w-[100vw] element becomes wider than `<body>` and introduces a horizontal scroll. macOS and mobile use overlay scrollbars, which is why the bug is invisible in most of our dev environments but shows up for real users.

If the full is not really useful for the real case, we can introduce the `default` value, which is responsive to the screen, e.g.
```
default: `gi-max-w-full xl:max-w-screen-2xl`
```
to make sure it is always full on mobile but never more than the default breakdown on larger screens (wherу the `maxWidth` actually makes sense).

I recommend investigating how it behaves on different screens to make the prop useful.

packages/react/src/container/container.tsx
```diff
@@ -1,5 +1,25 @@
-import { cn } from '../cn.js';
+import AtomicContainer, {
+  type Props as AtomicContainerProps,
+} from '../atoms/Container.js';
+import { tv } from 'tailwind-variants';
 
+const styles = tv({
```
DmitryB-NF: Better to place all private functions after public ones for better readability

examples/vite/src/application.tsx
```diff
@@ -3,12 +3,12 @@ import {
   Button,
   Card,
   Combobox,
-  Container,
   CookieBanner,
   DropdownItem,
   Footer,
   Form,
   FormFieldWithTag,
+  GiContainer,
```
DmitryB-NF: ⚠️ no `GiContainer` export

packages/angular/src/stories/Container.stories.ts
```diff
@@ -0,0 +1,136 @@
...
+} from '../atoms/storybook/Container.meta';
+
+const meta = {
+  ...containerMeta,
+  title: 'Layout/Container',
+};
+
+export default meta;
+
+const renderWithProjectedText = (arguments_: Record<string, unknown>) => ({
```
DmitryB-NF: `props` instead of `arguments_` for clarity?

packages/react/src/container/container.tsx
```diff
@@ -1,5 +1,25 @@
-import { cn } from '../cn.js';
+import AtomicContainer, {
```
DmitryB-NF: I would suggest sticking to the name convention as `GiContainer` to make sure all wrappers follow the same pattern

Ademsk1 (commented):
There are 4 images that are changed in this PR now. There wasn't any responsive padding beforehand for smaller screens - and so the padding would stay at p6 - which is not correct from the design.

packages/react/src/phase-banner/phase-banner.tsx
DmitryB-NF: This is actually a hidden issue in the existing code. `data-testId` is supposed to be unique as an ID of the element for testing. Having the same test ID for all containers in the same tree will fire a browser warning and is generally not recommended.
Since `Containers` can be used multiple times, it is better to remove this and the default value so the user can control it. Even if there is a default value right now, this more a fix while internal attributes are not customer facing feature and it is not a breaking change

packages/react/src/atoms/storybook/Container.meta.ts
```diff
@@ -0,0 +1,129 @@
...
+    children: 'Paragraph',
+    inset: true,
+    dataTestId: 'govie-container'
+  }
+};
+export const GuttersOnAndOff = {};
+export const RendersIndentedHTMLContent = {
+  tags: ['skip-playwright'],
+  args: {
+    children: createElement('p', {
```
DmitryB-NF: ⚠️ I suppose this approach actually leads to string checks in Angular. In the original idea, the Core Storybook is supposed to be used for the general documentation structure, but it was not about adding complex examples. If this does not work foк Angular, it is better explicitly reimplement this twice

packages/react/tests/visual.spec.ts-snapshots/layout-container--all-max-widths-chromium-linux.png
DmitryB-NF, line 1: ❓ Is this the existing Container behaviour? I am not sure contaner should be kind of half of the screen for the XL and 2XL sizes

packages/react/tests/visual.spec.ts-snapshots/layout-header-deprecated--full-width-iphone-linux.png
DmitryB-NF, line 1: ❓ Is this expected to have a different view after the migration? It looks like the container has invalid width

DmitryB-NF (dismissed):
Overall, the implementation seems solid, but there is a questionable approach with `maxWidth` that does not align with the existing view according to the screenshot changes.
Let me know if you need help with the implementation

packages/core/atoms/Container.lite.tsx
DmitryB-NF: @Ademsk1 I investigated the approach and I would suggest following wtih this set of styles. We basically already have the max-width support provided by the `gi-container` class based on the Tailwind theme, so we just need to extend it

```
export const styles = tv({
  base: 'gi-container gi-mx-auto',
  variants: {
    inset: {
      true: 'gi-py-4 md:gi-py-6 lg:gi-py-8',
    },
    gutters: {
      false: 'gi-px-0',   // gi-container override
    },
    maxWidth: {
      default: '', // defaults to gi-container
      sm:    'gi-max-w-screen-sm',      // 640px
      md:    'gi-max-w-screen-md',      // 768px
      lg:    'gi-max-w-screen-lg',      // 1024px
      xl:    'gi-max-w-screen-xl',      // 1280px
      '2xl': 'gi-max-w-screen-2xl',     // 1536px
      full:  'gi-max-w-full',           // parent 100% 
    },
  },
  defaultVariants: {
    inset: false,
    gutters: true,
    maxWidth: 'default',
  },
});
```
This should cover all the cases and work as expected without any additional validations

packages/angular/src/stories/Container.stories.ts
```diff
@@ -0,0 +1,136 @@
...
+  props: {
+    ...arguments_,
+    content: typeof arguments_['children'] === 'string' ? arguments_['children'] : '',
```
DmitryB-NF: Do we really need this check?

packages/react/tests/visual.spec.ts-snapshots/layout-header-deprecated--full-width-iphone-linux.png
Ademsk1: Ah, yes I believe it is, and it does standup to the stylistic practices we want.
gi-container-full-width was applied to this component. This class overrided responsive features, meaning that the padding that used to be here was px-6, rather than px-4.

packages/react/src/atoms/storybook/Container.meta.ts
Ademsk1: removed story

packages/angular/src/stories/Container.stories.ts
Ademsk1: removed

apps/docs/content/3-components/2-library/container/design.mdx
```diff
@@ -19,93 +19,31 @@ Use the Container to center page content and keep readable line lengths across b
   </Container>
 </ComponentPreview>
 
-## Anatomy & Classes
+## Layout
 
-- `gi-layout-container`: Default container with Tailwind’s container behaviour from the `md` breakpoint.
```
DmitryB-NF, line 24: ⚠️ We need this to be available in HTML still because this has not been changed. 
This should be invalidated as a Design part

apps/docs/content/3-components/2-library/container/react.mdx
```diff
@@ -11,36 +11,46 @@ component:
...
+          description: 'When true, applies responsive vertical padding (`py`). Default is false.'
+          required: false
+        - name: 'gutters'
+          ofType: 'boolean'
+          description: 'When true, applies responsive horizontal padding (`px`). Default is true.'
+          required: false
+        - name: 'maxWidth'
           ofType: 'string'
-          description: 'Defines the bottom padding of the container. Options are none, md, lg, and xl.'
+          description: 'Maximum width token: `sm`, `md`, `lg`, `xl`, `2xl`, or `full`. Default is `full`.'
```
DmitryB-NF: ⚠️ Should it be the default to flexible (which is the 2xl from `gi-container`)?

examples/vite/src/application.tsx
```diff
@@ -3,12 +3,12 @@ import {
   Button,
   Card,
   Combobox,
-  Container,
```
DmitryB-NF, line 6: Unexpected refactoring?

packages/react/src/phase-banner/phase-banner.tsx
```diff
@@ -45,7 +45,7 @@ export function PhaseBanner({
   }
 
   if (wrap === 'container-full-width') {
-    return <Container fullWidth>{bannerContent}</Container>;
+    return <Container maxWidth="full">{bannerContent}</Container>;
```
DmitryB-NF, line 48: ❓ Just in case, do we really need this to be fully available width, even bigger than the default value by default? I am not sure that `maxWidth=full` is equal to `fullWidth`. Can we check this?

pnpm-lock.yaml
DmitryB-NF, line 1: ⚠️ I think this change is not expected espicially having no dependencies change

packages/react/src/stories/Container.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <p>Try scaling the viewport to see the responsive gutters in action.</p>
+      <div className="gi-flex gi-flex-col gi-gap-2">
+        <span className="gi-font-bold gi-font-primary">gutters: true</span>
+        <Container gutters dataTestId="govie-container">
+          Sample content with horizontal gutters.
+        </Container>
+      </div>
+      <div className="gi-flex gi-flex-col gi-gap-2">
+        <span className="gi-font-bold gi-font-primary">gutters: false</span>
+        <Container gutters={false} dataTestId="govie-container">
```
DmitryB-NF: having 2 the same `dataTestId` is an antipattern

packages/react/src/stories/Container.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+          Sample content without horizontal gutters.
+        </Container>
+      </div>
+    </div>
+  ),
+};
+
+export const HandlesEmptyContentGracefully: Story = {
+  ...handlesEmptyContentGracefully,
+  tags: ['skip-playwright'],
```
DmitryB-NF: Should it be in the core `meta`?

packages/react/src/stories/Container.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const AllMaxWidths: Story = {
+  ...allMaxWidths,
+  render: () => (
+    <div className="gi-flex gi-flex-col gi-gap-8">
+      {map(MaxWidth, (maxWidth) => (
+        <div key={maxWidth} className="gi-flex gi-flex-col gi-gap-2">
+          <span className="gi-font-bold gi-font-primary">{maxWidth}</span>
+          <Container
+            maxWidth={maxWidth}
+            dataTestId="govie-container"
```
DmitryB-NF: Does it make sense not to assign `dataTestId` if it is not used?

packages/react/src/stories/Container.stories.tsx
```diff
@@ -0,0 +1,73 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import Container, { MaxWidth } from '../atoms/Container';
```
DmitryB-NF: ⚠️ It is better to test the public/client-facing Container (not atom) to make sure nothing is missing

packages/react/tests/visual.spec.ts-snapshots/layout-header-deprecated--full-width-iphone-linux.png
DmitryB-NF: Does it make sense to align it with the Design? I think we may may need to update the Header if the current behaviour is based on the design (apply styles to Header rather than on Container)

packages/react/src/stories/Container.stories.tsx
Ademsk1: I'm just removing it now. I don't think it's a sensible story.

packages/react/src/stories/Container.stories.tsx
Ademsk1: removed

packages/react/src/phase-banner/phase-banner.tsx
Ademsk1: TLDR: Yes we need this

`fullWidth` applied this style:
```
.gi-layout-container-full-width {
  @apply gi-w-full gi-max-w-[100vw] gi-px-6 lg:gi-px-8 2xl:gi-max-w-screen-2xl 2xl:gi-mx-auto;
}
```

Examining the `phase-banner` I noticed there was a missing story - that `container-full-width` was not included. I've added it now and it clearly shows the difference.

the `default` increments the max width size discretely, at each breakpoint. `max-w-full` however, allows the container to fluidly change its width with the viewports width. 

Manually checking this, I confirmed that `fullWidth` and `maxWidth="full" **do** give the same visual effects of the containers max size. Without it we don't

apps/docs/content/3-components/2-library/container/react.mdx
Ademsk1: After reviewing the designs, you're right it should be something like that. Current deployed implementation is unfortunately not doing that as well...

apps/docs/content/3-components/2-library/container/design.mdx
Ademsk1: Added back and labelled HTML only

packages/react/tests/visual.spec.ts-snapshots/layout-header-deprecated--full-width-iphone-linux.png
Ademsk1: This is aligned with the design of the Header as well. I've added a change in to match `md` sizes which wasn't getting caught before...let me know if you want to remove this.

packages/react/src/header/header-next/header-next.tsx
```diff
@@ -68,7 +68,7 @@ export const HeaderNext = forwardRef<HTMLElement, HeaderNextProps>(
           data-variant={variant}
           {...rest}
         >
-          <Container className="gi-order-2" fullWidth={fullWidth}>
+          <Container className="gi-order-2 gi-min-w-0 md:gi-px-8">
```
Ademsk1: `gi-min-w-0` is added to remove the inherited `gi-min-w-content` that comes from `grid` items. With `gi-min-w-content` truncation of the title text does not work properly. Fortunately we added a test for the header on this a couple weeks back to catch it. 

`md:gi-px-8` is actually not right and will be removed - checked with design.

packages/react/src/header/header-next/header-next.tsx
```diff
@@ -68,7 +68,7 @@ export const HeaderNext = forwardRef<HTMLElement, HeaderNextProps>(
           data-variant={variant}
           {...rest}
         >
-          <Container className="gi-order-2" fullWidth={fullWidth}>
+          <Container className="gi-order-2 gi-min-w-0">
```
Ademsk1, line 71: min-w-0 added as the container is a grid item, and they inherit the `min-w-content`. This prevented truncation from working when the title was very long (glad we added that test!). This sets the min-w-0 so that the containers width is not overflowing its own parent.

packages/react/src/container/container.tsx
```diff
@@ -10,36 +13,55 @@ export const ContainerInsetSizeEnum = {
...
+      inset={inset}
     >
       {children}
-    </div>
+    </GiContainer>
   );
 }
+
+const styles = tv({
+  base: 'gi-w-full gi-max-w-[100vw] gi-container gi-mx-auto',
```
DmitryB-NF: ⚠️ I suppose we need to remove base from here while `gi-max-w-[100vw]` brings issues, and all other styles should be properly assigned in the `GiContainer`. Basically, the main idea of having the wrapper is to process insets on top oа updated GiContainer

packages/react/src/phase-banner/phase-banner.stories.tsx
```diff
@@ -84,10 +84,19 @@ export const WrappedInContainer: Story = {
...
+    await expect(phaseBanner).toBe(containerElement?.firstElementChild);
+  },
+};
+
+export const WrappedInContainerFullWidth: Story = {
+  args: {
+    ...Default.args,
+    wrap: 'container-full-width',
   },
 };
```
DmitryB-NF, line 102: ⚠️ I am wondering if we need this test here. It does not seem to have any difference in view with `WrappedInContainer`. 
It looks like the wrap value `container-full-width` may not be relevant anymore and it is better to add TODO to remove this.

DmitryB-NF (dismissed):
The implementation of the Core Container looks good! Also, side fixes for the Header have finally made the behaviour consistent.
We just need to fix the `Container` wrapper to finally get rid of `gi-max-w-[100vw]` to make sure it follows the proper styling from `GiContainer`

DmitryB-NF:
@Ademsk1 I noticed that the old container is still used in some of the internal components. It would be nice to fix it

Screenshot 2026-04-24 at 12 57 13

packages/react/src/phase-banner/phase-banner.stories.tsx
Ademsk1: There is actually a visual difference here. Try changing the viewport between WrappedInContainerFullWidth and WrappedInContainer.

DmitryB-NF (approved):

PR #930, opened 2026-04-20
fix(AB#37355): fixing dataTestId on button react wrapper

ramonmulia:
## Description

 fixing dataTestId on button react wrapper

## Type of Issue

- [ ] 🚀 Feature
- [x] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[37355](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37355)

### DIFF
--- a/packages/react/src/button/button.tsx
+++ b/packages/react/src/button/button.tsx
@@ -24,6 +24,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       value,
       tabIndex,
       dataTestid,
+      dataTestId,
       'aria-label': ariaLabel,
       'aria-labelledby': ariaLabelledBy,
       'aria-describedby': ariaDescribedBy,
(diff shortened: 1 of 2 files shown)


packages/react/src/button/button.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render, screen } from '@testing-library/react';
+import { Button } from './button';
+
+describe('Button dataTestId forwarding', () => {
```
DmitryB-NF: Does it make sense to add one more `Play` test to existing button stories to include all consistent data across multiple frameworks?

ramonmulia: It does make sense, but our Storybook is currently using the Atom Button directly instead of the wrapper. If we want to cover this in the play test, we’d need to update Storybook to use the wrapper rather than the Atom Button itself.

I’m fine with that approach, but I’m not completely sure it’s the direction we want to go. 

what do you think?

DmitryB-NF: Hmm, it would be nice to make sure the React Storybook uses the React Legacy Button so we can test the proper component and see Client-side behaviour rather than internal functionality. We still can test _both_ components there, but the actual React Button should be tested there at first

DmitryB-NF (approved):

PR #932, opened 2026-04-22
feat(AB#37384): Adding stack mitosis component

ramonmulia:
## Description

Adding stack mitosis component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[37384](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37384)


packages/core/atoms/utilities.ts
```diff
@@ -39,3 +41,100 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+
+const breakpointKeys: BreakpointKey[] = _.values(Breakpoint);
+const baseBreakpoint: BreakpointKey = Breakpoint.XS;
+
+/**
+ * Generates a Tailwind class string from a responsive value.
+ * The `xs` breakpoint is treated as the base (unprefixed), matching Tailwind's
+ * mobile-first approach where unprefixed classes apply from zero up.
+ */
+export const resolveResponsive = <T extends string | number>(
```
DmitryB-NF: I think we can make it more usable as a parser of breakpoints, like
```
const BP = ['xs', 'sm', 'md', 'lg', 'xl', '2xl'] as const;

function resolveResponsive<T>(
  value: T | Partial<Record<typeof BP[number], T>>,
  toClass: (val: T, bp?: typeof BP[number]) => string,
): string {
  if (typeof value !== 'object') return toClass(value);

  return BP
    .filter((bp) => bp in value)
    .map((bp) => toClass(value[bp], bp))
    .join(' ');
}

// Usage:
resolveResponsive({ xs: 4, md: 6 }, (n, bp) => `${bp}:gi-gap-${n}`)
```

This way, we can use it for all components with full flexibility

packages/core/atoms/Stack.lite.tsx
```diff
@@ -0,0 +1,89 @@
...
+  ariaLabel?: string;
+  ariaLabelledBy?: string;
+  className?: string;
+  id?: string;
+  styles?: Record<string, string>;
+  children?: any;
+  dataTestId?: string;
+};
+
+export const stackVariants = tv({
```
DmitryB-NF: Better make it private and move after the component. Business logic first

packages/react/src/stack/types.ts
```diff
@@ -1,44 +1,39 @@
...
+    | 'start'
+    | 'center'
+    | 'end'
+    | 'between'
+    | 'around'
+    | 'evenly';
+  /** @deprecated Use `className` with a height utility instead. */
   fixedHeight?: string;
-  className?: string;
+  /** @deprecated Dividers are not supported in the cross-framework atom. This prop is React-only. */
```
DmitryB-NF: ⚠️ This message about Atoms does not really make sense for the React lib. It i better to note that this will be removed in future releases

packages/angular/package.json
```diff
@@ -26,6 +26,8 @@
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "~21.1.0",
+    "@types/lodash": "^4.17.24",
+    "lodash": "^4.17.23",
```
DmitryB-NF: ⚠️ Lodash should be a dependency for the correct work

packages/react/src/atoms/Stack.tsx
```diff
@@ -0,0 +1,113 @@
...
+  base: ['gi-flex', 'gi-w-full'],
+  variants: {
+    direction: {
+      row: 'gi-flex-row',
+      column: 'gi-flex-col',
+    },
+    smDirection: {
+      row: 'sm:gi-flex-row',
+      column: 'sm:gi-flex-col',
+    },
```
DmitryB-NF: ⚠️ `gi-flex-row` is basically included in all directions and can be simplified, as well as `gi-flex-col`. See the comment about the responsive sizes to make it more usefull

packages/core/atoms/Stack.lite.tsx
```diff
@@ -0,0 +1,89 @@
...
+  base: ['gi-flex', 'gi-w-full'],
+  variants: {
+    direction: { row: 'gi-flex-row', column: 'gi-flex-col' },
+    smDirection: { row: 'sm:gi-flex-row', column: 'sm:gi-flex-col' },
+    mdDirection: { row: 'md:gi-flex-row', column: 'md:gi-flex-col' },
+    lgDirection: { row: 'lg:gi-flex-row', column: 'lg:gi-flex-col' },
+    xlDirection: { row: 'xl:gi-flex-row', column: 'xl:gi-flex-col' },
+    '2xlDirection': { row: '2xl:gi-flex-row', column: '2xl:gi-flex-col' },
```
DmitryB-NF: ⚠️ I think we can use more generic helpers to simplify TV and simply generate the pure TV classes instead of parsing it twice (Responsive -f-> list of strings -tv-> list of classes)

packages/core/atoms/utilities.ts
```diff
@@ -39,3 +41,100 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+} as const;
+
+export const Breakpoint = {
+  XS: 'xs',
+  SM: 'sm',
+  MD: 'md',
+  LG: 'lg',
+  XL: 'xl',
+  '2XL': '2xl',
+} as const;
```
DmitryB-NF: 💡 It seems like it would be nice to move into a separate file like `types/constants` to have all generic enums separately to split too large utilities in the future

DmitryB-NF (dismissed):
New component looks really good and useful!
We may need to clean up the Responsive approach to make it more future-proof

packages/core/atoms/Stack.lite.tsx
```diff
@@ -0,0 +1,85 @@
+import { useMetadata } from '@builder.io/mitosis';
+import { tv } from 'tailwind-variants';
+import { Direction, AlignItems, Justify } from './constants';
+import type { ResponsiveValue } from './constants';
```
Ademsk1: Shared import location?

packages/core/atoms/utilities.ts
```diff
@@ -1,35 +1,6 @@
...
-  SECONDARY: 'secondary',
-  FLAT: 'flat',
-} as const;
-
-export const Appearance = {
-  DEFAULT: 'default',
-  DARK: 'dark',
-  LIGHT: 'light',
-} as const;
+import _ from 'lodash';
```
Ademsk1, line 1: Can we use `lodash` instead? `_` feels a little wrong to me..

packages/core/atoms/utilities.ts
```diff
@@ -39,3 +10,33 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
 
 export const getAlign = (x: (typeof Align)[keyof typeof Align] = Align.START) =>
   Object.values(Align).includes(x) ? x : Align.START;
+
+export const getAlignItems = (x: (typeof AlignItems)[keyof typeof AlignItems] = AlignItems.START) =>
+  Object.values(AlignItems).includes(x) ? x : AlignItems.START;
```
Ademsk1: lodash here?

packages/core/atoms/utilities.ts
```diff
@@ -39,3 +10,33 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+
+export const getJustify = (x: (typeof Justify)[keyof typeof Justify] = Justify.START) =>
+  Object.values(Justify).includes(x) ? x : Justify.START;
```
Ademsk1: lodash here?

packages/core/atoms/Stack.lite.tsx
```diff
@@ -0,0 +1,85 @@
...
+      aria-label={props.role ? props.ariaLabel : undefined}
+      aria-labelledby={props.role ? props.ariaLabelledBy : undefined}
+      style={props.styles}
+      class={stackVariants({
+        align: getAlignItems(props.align),
+        justify: getJustify(props.justify),
+        wrap: props.wrap ?? false,
+        class: [
+          resolveResponsive(props.direction ?? Direction.COLUMN, directionToClass),
+          resolveResponsive(props.gap ?? 0, gapToClass),
```
Ademsk1: why put resolveResponsive here? Why not give it its own variants key/value entry in the `stackVariants`?

packages/core/atoms/utilities.ts
```diff
@@ -39,3 +10,33 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+): string => {
+  if (!_.isPlainObject(value)) {
+    return toClass(value as T);
+  }
+
+  const responsive = value as Partial<Record<BreakpointKey, T>>;
+  return breakpointKeys
+    .filter((bp) => bp in responsive)
+    .map((bp) => toClass(responsive[bp] as T, bp === baseBreakpoint ? undefined : bp))
+    .join(' ');
```
Ademsk1, line 45: use the `reduce`?

packages/core/atoms/utilities.ts
```diff
@@ -39,3 +10,33 @@ export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] =
...
+const breakpointKeys: BreakpointKey[] = _.values(Breakpoint);
+const baseBreakpoint: BreakpointKey = Breakpoint.XS;
+
+/**
+ * Generates a Tailwind class string from a responsive value.
+ * The `xs` breakpoint passes `undefined` as `bp` to the callback,
+ * matching Tailwind's mobile-first approach where unprefixed classes
+ * apply from zero up.
+ */
+export const resolveResponsive = <T>(
```
Ademsk1, line 33: This function is great. It deserves a dev team meeting to demo it properly.

Ademsk1 (changes requested):

packages/core/atoms/Stack.lite.tsx
ramonmulia: yeap, it is shared, used by any component that needs layout responsive types.

packages/core/atoms/utilities.ts
ramonmulia: `import _ from 'lodash'` is the universally recognized convention for lodash, the `_ `alias is literally lodash's identity.

Renaming it to import lodash from 'lodash' would actually look unfamiliar to most developers and goes against the established convention. https://lodash.com/docs

packages/core/atoms/Stack.lite.tsx
ramonmulia: This is the same approach Dmitry said was duplication before.
It also doesn’t work well for gap. gap is a number, so you can’t list all values (gi-gap-0 ... gi-gap-96) as variants.
Instead, resolveResponsive solves this. It takes the user input and a small mapper function, and builds the final class:
```
directionToClass('row', 'md') -> "md:gi-flex-row"
gapToClass(6, 'lg') -> "lg:gi-gap-6"
```
Then we pass this to tv class option, this option is made for dynamic classes that don’t fit static variants.
Static props like align, justify, and wrap are still good as TV variants because they are single values, not responsive.

packages/core/atoms/utilities.ts
ramonmulia: Totally. I also added an example in the function description, hope it helps!

Ademsk1 (commented):
One last thing to consider

We've now merged the <Container> component. This comes with the `clamp` utility I built. Would you consider using it here instead of `getGap` or `getJustify` ?

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,76 @@
...
-export const getWhitespace = (x: (typeof Whitespace)[keyof typeof Whitespace] = Whitespace.NORMAL) =>
-  Object.values(Whitespace).includes(x) ? x : Whitespace.NORMAL;
-
-export const getAlign = (x: (typeof Align)[keyof typeof Align] = Align.START) =>
-  Object.values(Align).includes(x) ? x : Align.START;
+import _ from 'lodash';
+import { Align, AlignItems, Breakpoint, Justify, Size, Whitespace } from './constants';
+import type { BreakpointKey, ResponsiveValue } from './constants';
 
+export const getSize = (x: (typeof Size)[keyof typeof Size] | undefined) => clamp(x, Size, Size.MD);
```
ramonmulia, line 5: I'm keeping those utilities classes for now cause calling `clamp(props.size, Size, Size.MD)` in a .lite file exposes Size in the angular template, where it resolves to `this.Size`, a property that does not exist on the generated class, causing a TS2551 error.
The getter patterns like `getSize`, keeps the `as const` reference inside the function body, never in JSX, only the function itself appears in the template and gets promoted to a class property.

When the as const object is declared locally in the same .lite.tsx file (like Container in MaxWidth), Mitosis does promote it, so we don't see this issue.

apps/docs/content/3-components/2-library/stack/react.mdx
```diff
@@ -10,62 +10,100 @@ component:
...
+          description: '@deprecated — Use ariaLabel instead. Maps to aria-label.'
+          required: false
+        - name: 'aria-labelledby'
+          ofType: 'string'
+          description: '@deprecated — Use ariaLabelledBy instead. Maps to aria-labelledby.'
+          required: false
+        - name: 'data-testid'
+          ofType: 'string'
+          description: '@deprecated — Use dataTestId instead. Maps to data-testid.'
+          required: false
```
DmitryB-NF: ❓ Does it make sense to remove the deprecated prop from the Docs? Overall, it will be visible for all developers but not really required in the representation web page

packages/angular/ng-package.json
```diff
@@ -4,5 +4,6 @@
     "entryFile": "src/index.ts"
   },
   "dest": "dist",
-  "assets": ["./styles.css", "./fonts.css"]
+  "assets": ["./styles.css", "./fonts.css"],
+  "allowedNonPeerDependencies": ["lodash"]
```
DmitryB-NF: ⚠️ this should be removed in favour of `peerDependencies`. You need to merge `main` into the branch

packages/core/atoms/storybook/Stack.meta.ts
```diff
@@ -0,0 +1,237 @@
...
+      expect(element).toBeInTheDocument();
+      expect(element.tagName).toBe('DIV');
+    });
+
+    await step('exposes aria-label when role is set', async () => {
+      const element = canvas.getByTestId('stack-responsive-test');
+      expect(element).toHaveAttribute('role', 'region');
+      expect(element).toHaveAttribute('aria-label', 'layout region');
+    });
+  },
```
DmitryB-NF: Does it make sense to test in the Default? It does not seem related to `Responsive` representation?

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,76 @@
...
+ * otherwise falls back to the provided default.
+ *
+ * Example:
+ * const Size = { SM: 'sm', MD: 'md', LG: 'lg' } as const;
+ *
+ * clamp('sm', Size, Size.MD);
+ * //"sm"
+ *
+ * clamp('invalid', Size, Size.MD);
+ * //"md"
```
DmitryB-NF: JSDoc is good, but the example is kind of excessive for 2-lines function

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,76 @@
...
+ *
+ */
 export function clamp<T extends Record<string, string>>(
   value: string | undefined,
   options: T,
   defaultValue: T[keyof T],
 ): T[keyof T] {
-  const allowed = Object.values(options) as T[keyof T][];
+  const allowed = _.values(options) as T[keyof T][];
   return value !== undefined && allowed.includes(value as T[keyof T]) ? (value as T[keyof T]) : defaultValue;
```
DmitryB-NF: If using lodash, it could be simplified like `_.includes(allowed, value) ? value : defaultValue`. Lodash is typesafe, so there is no need for extra casts and checks

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,76 @@
...
+  toClass: (val: T, bp?: BreakpointKey) => string,
+): string => {
+  if (!_.isPlainObject(value)) {
+    return toClass(value as T);
+  }
+
+  const responsive = value as Partial<Record<BreakpointKey, T>>;
+  return _.reduce(
+    _.values(Breakpoint),
+    (result, bp) => {
```
DmitryB-NF: Overall, reduce is not really readable and is suggested to be avoided by the latest ESLint setup. Better to use a map and join
```
_.map(responsive, (value, bp) => Breakpoint[bp] && toClass(value, bp)).filter().join(' ')
```

packages/react/src/utils/normalize-breakpoints.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  const object = value as Record<string, unknown>;
+
+  if (_.has(object, 'base')) {
+    if (!_.has(object, 'xs')) {
+      return { ..._.omit(object, 'base'), xs: object.base } as T;
+    }
+    return _.omit(object, 'base') as T;
+  }
```
DmitryB-NF: ⚠️ This could be a violation of the existing logic in the Footer stories `packages/react/src/footer/footer.stories.tsx:191` where `direction={{ base: 'column', xs: 'column', md: 'row' }}` will work but differently. 
Basically, according to `packages/design/tokens/tokens/light/primitive/screen.json`, xs is a real breakpoint which is used in the Design system for some reason. Removing it slightly in `resolveResponsive` does not seem like a good idea. It is not aligned with the TV theme and is a breaking change for the existing Stack API

I would suggest adding the `base` to Breakpoints and resolving this separately, probably by removing XS from the theme

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,76 @@
...
+ *
+ * resolveResponsive(4, toGap);
+ * //"gi-gap-4"
+ *
+ * resolveResponsive({ xs: 2, md: 4, xl: 8 }, toGap);
+ * //"gi-gap-2 md:gi-gap-4 xl:gi-gap-8"
+ */
+export const resolveResponsive = <T>(
+  value: ResponsiveValue<T>,
+  toClass: (val: T, bp?: BreakpointKey) => string,
```
DmitryB-NF: To avoid constant checking for `bp ? {bp}:{class} : class`, you can change the api a bit to have it like
`(val: T, prefix: string, bp?: BreakpointKey) => string` to simply do `resolveResponsive({ xs: 1, sm: 2 }, (val, prefix) => prefix + f(value))` significantly simplifying the code

packages/react/src/stack/stack.tsx
```diff
@@ -1,257 +1,82 @@
...
-            )
-          : child}
-
-        {hasDivider && index < childrenComponent.length - 1 && (
-          <Divider direction={direction} />
-        )}
-      </Fragment>
-    ));
-  };
+  const normalizedDirection = normalizeLegacyBreakpoints(direction);
```
DmitryB-NF: ⚠️ we need to use `Base` or something to represent `0-480`

packages/react/src/stack/stack.tsx
```diff
@@ -1,257 +1,82 @@
...
+      direction={normalizedDirection}
+      gap={normalizeLegacyBreakpoints(gap)}
+      align={align ?? itemsAlignment}
+      justify={justify ?? itemsDistribution}
+      wrap={wrap}
+      role={role}
+      ariaLabel={ariaLabel ?? nativeAriaLabel}
+      ariaLabelledBy={ariaLabelledBy ?? nativeAriaLabelledBy}
+      className={className}
+      styles={fixedHeight ? { height: fixedHeight } : undefined}
```
DmitryB-NF: ❓ Can we do that with `className` to still support the default `style` attribute?

packages/react/src/stack/types.ts
```diff
@@ -1,44 +1,39 @@
...
+  /** @deprecated Use `xs` instead. */
+  base?: SimpleDirection;
+};
+
+type LegacyResponsiveGap = ResponsiveValue<number> & {
+  /** @deprecated Use `xs` instead. */
+  base?: number;
+};
+
+export type StackProps = Omit<StackAtomProps, 'direction' | 'gap'> & {
```
DmitryB-NF: Overall, the file is relatively small and will be even smaller with `base` as a breakpoint. Does it make sense to merge it with the stack wrapper?

DmitryB-NF (changes requested):

packages/react/src/utils/normalize-breakpoints.ts
ramonmulia: good call, at first, I followed tailwind’s default idea where xs does not exist and the base (no prefix) covers 0px+. but as you mentione according with in our design system, xs is defined as 480px in the tokens and is already used I kept `base` breakpoint.
I'm really not sure why it is set this way..

packages/react/src/stack/stack.tsx
ramonmulia: hmm not really, tailwind removes unused classes during build by scanning for fixed class names, so a dynamic class like `gi-h-[${value}]` is not a fixed string... so it will not be included in the final css.

since `fixedHeight` can receive any value (like "200px"), using inline style is the only reliable solution here to not break api.

apps/docs/content/3-components/2-library/stack/react.mdx
ramonmulia: yeah, I’ve been in favor of removing it from the start, but I kept it because of this discussion: [907](https://github.com/ogcio/govie-ds/pull/907/changes#diff-9ccda63c1dfa0bf1c8c3d0ecdbda31bb2d37b84930f436ab2c884f69ecf4f48b)
happy to remove it for Stack.

DmitryB-NF: The key point here is to avoid removing existing public docs, but also avoiding adding deprecated ones.
We definitely need to avoid breaking changes in the documented API, but we also do not need to add default React/HTML props and mark them as deprecated. If there were no documented props but silent support, we do not need to add them post-factum with a deprecation message.

packages/core/atoms/Stack.lite.tsx
```diff
@@ -0,0 +1,82 @@
...
+    >
+      {props.children}
+    </div>
+  );
+}
+
+const directionToClass = (direction: string, prefix: string): string =>
+  direction === 'row' ? `${prefix}gi-flex-row` : `${prefix}gi-flex-col`;
+
+const gapToClass = (gap: number, prefix: string): string => `${prefix}gi-gap-${gap}`;
```
DmitryB-NF: The function is now really simple and can be inlined
```
resolveResponsive(props.gap ?? 0, (n, prefix) => `${prefix}gi-gap-${n}`)
```

packages/core/atoms/utilities.ts
```diff
@@ -1,50 +1,46 @@
...
+  value: ResponsiveValue<T>,
+  toClass: (value: T, prefix: string) => string,
+): string => {
+  if (!_.isPlainObject(value)) {
+    return toClass(value as T, '');
+  }
+
+  const responsive = value as Partial<Record<BreakpointKey, T>>;
+  return _.values(Breakpoint)
+    .filter((bp) => bp in responsive)
```
DmitryB-NF, line 43: Just in case, lodash can iterate over objects, so it can be simplified like `_.filter(Breakpoint, (bp) => bp in responsive)`. Also, I would suggest adding filtering the reverse way because it requires slightly fewer iterations (size of `responsive` <= `Breakpoint` always)
```
_.fitler(responsive, bp => Breakpoint[bp])
  .map(...)
```

packages/react/src/stack/stack.tsx
```diff
@@ -1,257 +1,99 @@
...
-  return (
-    <div
-      className={`${cn(dividerClasses)} gi-bg-gray-400`}
-      aria-hidden="true"
-    />
-  );
-};
+export type StackProps = Omit<StackAtomProps, 'direction' | 'gap'> & {
+  direction?: ResponsiveValue<'row' | 'column'>;
+  gap?: ResponsiveValue<number>;
```
DmitryB-NF: This seems exactly the same as in `StackAtomProps`, so Omit is not requried

packages/react/src/stack/stack.tsx
DmitryB-NF: ⚠️ Make sense. But we still need to merge `styles` than because it is in the list of props in the documentation

DmitryB-NF (changes requested):
The implementation looks really nice and stable now!
We need to fix the `styles`/`style` prop, and we are good to go!

DmitryB-NF (approved):

PR #933, opened 2026-04-27
chore(AB#37519): bump React package deps and address pnpm audit

DmitryB-NF:
## Description

Bump lodash and realign its declarations across packages (root → consumers). 
Bump eslint stack and downgrade new unicorn rules to warn so lint stays clean. 
Bump Storybook stack to 10.3.5 and leave a TODO on the addon-themes regression.

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#37519](https://dev.azure.com/OGCIO-Digital-Services/2aff229e-6625-4ad7-8f51-ff850dd878ec/_workitems/edit/37519)


packages/angular/package.json
```diff
@@ -19,7 +19,8 @@
   },
   "peerDependencies": {
     "@angular/common": "^21.1.0",
-    "@angular/core": "^21.1.0"
+    "@angular/core": "^21.1.0",
+    "lodash": "^4.0.0"
```
ramonmulia, line 23: hmm, I think this should be a regular dependency, not a peer dependency.
As a peer dependency, every Angular consumer would need to install lodash manually, I believe It’s better to keep it under dependencies and include this config:
https://github.com/ogcio/govie-ds/blob/90a61599055a8af1dc0de808947c99decba1fea8/packages/angular/ng-package.json#L8

Ademsk1 (commented):
`lodash` is still 4.17 in app/docs - could be good to update there as well?

packages/angular/package.json
DmitryB-NF: There are 2 ways of implementing this, and the suggested Angular way is to add this to `peerDependencies` to make sure it is installed on the client side. So, the community way is implemented https://github.com/ng-packagr/ng-packagr/blob/main/docs/dependencies.md

DmitryB-NF:
@Ademsk1, we may update lodash in all other places, but it is out of scope for the PR. Overall, we need to do a similar audit for all packages eventually

ramonmulia (dismissed):

PR #937, opened 2026-04-27
fix(AB#37363): add custom Logo width/height to the Footer

Ademsk1:
## Description

Adds in user-specified width/height. Defaults to `gi-h-16` retaining aspect ratio for no visual regressions.

Outside the scope of this bug, the`getLogo` function can be massively improved. Firstly, `svgMobileString` is generated, yet it will never be used. Despite it being used as the `src` value in the image, the `<source>` overwrites it. This was probably implemented to support browsers that didn't support `<picture>`. Now all of the main ones do [(see here)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/picture#browser_compatibility).

I do not think this is the right place to make this change. But I believe it should be put forward. 

A further question - why use `<picture>` / `` at all? Why not just render the svg directly instead of generating it as an image? Perhaps we can consider these in the future / when migrating to mitosis. 

 

## Type of Issue

- [ ] 🚀 Feature
- [x] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#37363](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37363)


packages/react/src/footer/footer.stories.tsx
```diff
@@ -25,7 +25,8 @@ const meta: Meta<typeof Footer> = {
     },
     logo: {
       control: 'object',
-      description: 'Optional custom logo component',
+      description:
+        'Optional custom logo component. Either width or height must be specified to override the default size, with width taking precedence to retain aspect ratio.',
```
DmitryB-NF: I am not sure if this is required to add, as width/height are optional and may not be provided

packages/react/src/footer/footer.tsx
```diff
@@ -29,12 +29,14 @@ function getLogo({ logo }: FooterProps) {
     renderToStaticMarkup(<GovieLogoHarpWithText />),
   );
   const svgDataUriDesktop = `data:image/svg+xml;base64,${svgDesktopString}`;
-
+  const isLogoSizeDefined = logo?.width || logo?.height;
```
DmitryB-NF: Const could be inlined

packages/react/src/footer/footer.tsx
```diff
@@ -29,12 +29,14 @@ function getLogo({ logo }: FooterProps) {
...
   return (
     <picture>
       <source srcSet={logo?.imageLarge || svgDataUriDesktop} />
       <img
-        className="gi-h-16"
+        width={logo?.width}
+        height={logo?.height}
+        className={cn({ 'gi-h-16': !isLogoSizeDefined })}
```
DmitryB-NF: ⚠️ Can we have a Storybook test to cover this? Passing width and height could change the view when using custom images (instead of relative calculation), so we need to make sure this works

DmitryB-NF (dismissed):
The fix looks good, but we need a Storybook demo/test of this feature

packages/react/src/footer/footer.tsx
Ademsk1: I actually prefer this way - it's clarity is better when doing `!isLogoSizeDefined`. 
using `(!logo?.width && !logo.height)` is less clear.

DmitryB-NF: Ok, we can use the variable, but when reading `(!logo?.width && !logo.height)`, it seems clear that this is the check for `no width and height`

packages/react/src/footer/footer.stories.tsx
```diff
@@ -6,6 +6,13 @@ import { SectionBreak } from '../section-break/section-break.js';
 import { Stack } from '../stack/stack.js';
 import { Footer } from './footer.js';
 
+const customLogoSvg = `<svg xmlns="http://www.w3.org/2000/svg" width="220" height="56" viewBox="0 0 220 56" role="img" aria-label="Custom logo">
```
DmitryB-NF: It is better to use existing `generateSvgPlaceholderDataUrl` for testing to be consistent

packages/react/src/footer/footer.stories.tsx
```diff
@@ -6,6 +6,13 @@ import { SectionBreak } from '../section-break/section-break.js';
...
+  <rect width="200" height="56" rx="8" fill="#0B4C7A"/>
+  <circle cx="28" cy="28" r="12" fill="#FFFFFF"/>
+  <path d="M58 18h140v6H58zM58 32h110v6H58z" fill="#FFFFFF" opacity="0.9"/>
+</svg>`;
+const customLogoDataUri = `data:image/svg+xml,${encodeURIComponent(customLogoSvg)}`;
```
DmitryB-NF: It is better to move closer to the usage while this is a local cont

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +502,42 @@ export const TestAllSlots: Story = {
...
+
+export const CustomLogo: Story = {
+  args: {
+    logo: {
+      imageSmall: customLogoDataUri,
+      imageLarge: customLogoDataUri,
+      alt: 'Custom footer logo',
+      href: '/custom-logo',
+      // width greater than svg width to show svg size can be overridden
+      width: 220,
```
DmitryB-NF: Does it make sense both props are provided?

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +502,42 @@ export const TestAllSlots: Story = {
     });
   },
 };
+
+export const CustomLogo: Story = {
```
DmitryB-NF, line 492: ⚠️ Remove from playwright?

DmitryB-NF (dismissed):

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +496,44 @@ export const TestAllSlots: Story = {
...
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+
+    await step('should apply logo width/height when provided', async () => {
+      const img = canvas.getByAltText('Custom footer logo');
+
+      // When width/height are undefined, React omits the attribute entirely.
+      // We assert width is present here because this story sets it by default.
+      expect(img).toHaveAttribute('width', '220');
```
ramonmulia: should we also include a test for `height`?

Ademsk1: It's kind of redundant because the SVG (and therefore the image) will always retain aspect ratio - and base the height off of the width in this way. width is the deciding factor here.

DmitryB-NF: I would rather add a test for `height` because is passing a specific height=100, user can expect it to be set properly

Ademsk1: Now testing for variants of height and width props

packages/react/src/utils/atoms-utilities.test.ts
```diff
@@ -24,7 +24,8 @@ describe('clamp', () => {
   });
 });
 
-const toClass = (value: string, prefix: string): string => `${prefix}gi-flex-${value}`;
```
ramonmulia: thanks for the fix!
It seems like the auto formatting failed to trigger when I pushed that stacked PR.

DmitryB-NF: @Ademsk1 I moved the fix to `main` to make sure it is built properly

packages/react/src/footer/footer.tsx
```diff
@@ -28,13 +28,18 @@ function getLogo({ logo }: FooterProps) {
...
     <picture>
       <source srcSet={logo?.imageLarge || svgDataUriDesktop} />
       <img
-        className="gi-h-16"
+        width={logo?.width}
+        height={logo?.height}
+        style={{
+          width: logo?.width,
+          height: isLogoSizeDefined ? logo?.height : 64,
+        }}
```
DmitryB-NF: I would suggest adding defaults for our image, as it was initially to make sure we can resolve warnings and have a consistent design
```
// default logo size ...
const LOGO = { width: 181, height: 64 }
...
width={logo?.width ?? LOGO.width}
height={logo?.height ?? LOGO.height}
className="gi-h-16 gi-w-auto"
```

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +514,122 @@ export const TestAllSlots: Story = {
     });
   },
 };
+
+export const LogoDimensions: Story = {
```
DmitryB-NF: Let's name it as CustomLogo or similar to pass different dimentions to see it is properly rendered in the reserved box with a proper aspect ration

packages/react/src/footer/footer.tsx
```diff
@@ -22,19 +22,22 @@ export type FooterProps = {
   dataTestid?: string;
 } & React.HTMLAttributes<HTMLDivElement>;
 
+const LOGO = { width: 181, height: 64 };
```
DmitryB-NF: It is better to add a comment that this is the default logo size

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +514,72 @@ export const TestAllSlots: Story = {
     });
   },
 };
+
+export const CustomLogo: Story = {
+  render: (arguments_) => {
```
DmitryB-NF: Better name it `props` following a proper new convention

packages/react/src/common/types.ts
```diff
@@ -1,6 +1,8 @@
 export type LogoProps = {
   imageSmall?: string;
   imageLarge?: string;
+  height?: number;
```
ramonmulia: it looks `LogoProps` is also imported by the `Header` types, but the Header `getLogo` doesn't consume          width/height, it uses fixed sizing..
I think consumers might see these new props in the type and could assume they work on the Header too.

packages/react/src/footer/footer.stories.tsx
```diff
@@ -391,6 +392,24 @@ export const MinimalFooter: Story = {
...
     ),
   },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    await step(
+      'should render the default logo with default width and height',
+      async () => {
+        const img = canvas.getByAltText('Gov.ie Logo');
+        const computedStyle = getComputedStyle(img);
+        const width = Math.round(
```
ramonmulia: it could be simplified by using `getRoundedDimensions`  here too.

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +514,72 @@ export const TestAllSlots: Story = {
...
+  tags: ['skip-playwright'],
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+
+    await step(
+      'should reserve intrinsic ratio and enforce footer logo height',
+      async () => {
+        const img = canvas
+          .getByTestId('footer-custom-logo')
+          .querySelector('img');
```
ramonmulia: I think `canvas.getByAltText('Gov.ie Logo')` would work, and that's the pattern already used in `MinimalFooter` play test

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +514,72 @@ export const TestAllSlots: Story = {
...
+        expect(getRoundedDimensions(img as Element)).toMatchObject({
+          width: Math.round((220 / 80) * 64),
+          height: 64,
+        });
+      },
+    );
+  },
+};
+
+// utility function to get rounded dimensions of an element
```
ramonmulia: hmm I think  this comment is redundant, the function name already says what it does.

packages/react/src/common/types.ts
Ademsk1: Took it out, and added it in to the `FooterProps`
```ts
export type FooterProps = {
  primarySlot?: ReactNode;
  secondarySlot?: ReactNode;
  utilitySlot?: ReactNode;
  logo?: LogoProps & { width?: number; height?: number };
  className?: string;
  dataTestid?: string;
} & React.HTMLAttributes<HTMLDivElement>;
```

packages/react/src/footer/footer.stories.tsx
```diff
@@ -391,6 +392,19 @@ export const MinimalFooter: Story = {
       </Stack>
     ),
   },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    await step(
+      'should render the default logo with default width and height',
+      async () => {
+        const img = canvas.getByAltText('Gov.ie Logo');
+        const computedStyle = getComputedStyle(img);
```
DmitryB-NF: unused variable?

packages/react/src/footer/footer.stories.tsx
```diff
@@ -391,6 +392,19 @@ export const MinimalFooter: Story = {
...
+        const { width, height } = getRoundedDimensions(img);
```
DmitryB-NF: ⚠️ it is better to check that props are provided because the actual size will be validated via Playwright

packages/react/src/footer/footer.stories.tsx
```diff
@@ -495,3 +509,68 @@ export const TestAllSlots: Story = {
...
+      async () => {
+        const img = canvas.getByAltText('Gov.ie Logo');
+        expect(img).toHaveAttribute('width', '220');
+        expect(img).toHaveAttribute('height', '80');
+
+        expect(getRoundedDimensions(img as Element)).toMatchObject({
+          width: Math.round((220 / 80) * 64),
+          height: 64,
+        });
+      },
```
DmitryB-NF, line 511: ⚠️ Having playwright tests in place, I would suggest checking only props to be a clear Unit test. Testing browser rendering of Tailwind styles could be flaky and difficult to support

DmitryB-NF (changes requested):

DmitryB-NF (dismissed):

DmitryB-NF (approved):

PR #939, opened 2026-04-28
feat(AB#37432): Adding Grid mitosis component 

ramonmulia:
## Description

Adding Grid mitosis component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues
[37432](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37432)


packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,151 @@
...
+  base: 4,
+  xs: 4,
+  sm: 6,
+  md: 8,
+  lg: 12,
+  xl: 12,
+  '2xl': 12,
+};
+
+// Class maps must be static strings (no template literals) so Tailwind JIT can scan them.
```
DmitryB-NF: ⚠️ Does it make sense to add this to the safelist or directly to the Tailwind theme? It looks like a workaround to populate styles to use them as classes. It would be nice to split this to significantly simplify the code

packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,151 @@
...
+  return Object.fromEntries(
+    (_.keys(responsive) as BreakpointKey[]).map((bp) => [key(bp), transform(responsive[bp] as number)]),
+  );
+};
+
+const styles = tv({
+  base: 'gi-min-w-0',
+  variants: {
+    container: {
+      true: 'gi-flex gi-flex-wrap gi-mx-4 md:gi-mx-6 lg:gi-mx-8',
```
DmitryB-NF: ⚠️ I do not think we need hard margins for the Grid. Grid is supposed to be a flex-box wrapper where margins are the user's input. `Container` component should be used if we need a specific layout organisation

packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,151 @@
...
+      false: [
+        'gi-grow-0 gi-shrink-0',
+        // Grid item basis formula: span / cols * 100% - gap * (cols - span) / cols, capped at 100%.
+        // Must be a plain string so tailwind can scan and generate the CSS rule.
+        '[--grid-basis:min(calc(var(--grid-span)_/_var(--grid-parent-columns)_*_100%_-_var(--grid-parent-gap)_*_(var(--grid-parent-columns)_-_var(--grid-span))_/_var(--grid-parent-columns)),100%)]',
+        'gi-basis-[var(--grid-basis)] gi-max-w-[var(--grid-basis)]',
+      ].join(' '),
```
DmitryB-NF: I think `join` is redundant

packages/react/src/atoms/Grid.tsx
```diff
@@ -0,0 +1,155 @@
...
+            ? [
+                resolveResponsive(props.gap ?? 0, fromMap(GAP_CLASSES)),
+                resolveResponsive(props.columns ?? FigmaColumns, fromMap(PARENT_COL_CLASSES)),
+                resolveResponsive(props.gap ?? 0, fromMap(PARENT_GAP_CLASSES)),
+              ].join(' ')
+            : resolveResponsive(props.size ?? 1, fromMap(SPAN_CLASSES)),
+          props.className,
+        ]).join(' '),
+      })}
+      style={props.container ? buildContainerStyle(props.columns, props.gap) : buildItemStyle(props.size)}
```
DmitryB-NF: I think we may need to add an additional layer to the TW theme for this. Making TW process it by crawling styles does not seem a good and readable way to go

examples/nextjs/src/app/page.tsx
```diff
@@ -936,6 +937,27 @@ export default function Home() {
                     </div>
                   </div>
 
+                  <div>
+                    <h5 className="font-semibold mb-2">Grid (Responsive Size)</h5>
+                    <Grid container gap={4} dataTestId="grid-responsive-size">
+                      <Grid size={{ base: 4, xs: 4, sm: 4, md: 6, lg: 8 }} dataTestId="grid-rsize-content">
```
DmitryB-NF: `base: 4, xs: 4, sm: 4` - can we do it like `base: 4, md: 6` to avoid passing all the values?

packages/core/atoms/Grid.lite.tsx
ramonmulia: yes, adding padding here feels a bit strange to me, but I am following the [Figma](https://www.figma.com/design/ZEqfpXhocdZs1K5YfsE5ws/Design-system?node-id=4632-2926)

maybe we can add a note in the docs to suggest using it this way and remove the padding? so it is clearer when to use it.

packages/react/src/atoms/Grid.tsx
ramonmulia: that’s a good point, in the current pattern, the inline style only sets css variables.. so the tw classes use them. this works well and keeps a clear separation in the developer tools.
But I agree that having a tw approach without inline styles would be much better.

Maybe we can have a tw plugin where each of class maps to a css variables, set by tw plugin.
example: `gi-grid-columns-4` would be `--grid-columns: 4` and so on, we would have zero inline style, everything in classes, and more readable, but a new plugin file, export, safelist entries, and config changes in both react and angular.

keeping the inline style approach, It is self contained and has no external dependencies, it is kind of trade 3-4 css variable definitions in style for a plugin + safelist + inheritance fix.

I lean towards creating the plugin for consistency, but open to keeping inline styles if the added setup is not worth it. 
what do you think?

packages/core/atoms/Grid.lite.tsx
DmitryB-NF: Figma describes the overall experience of the Page when using grid display, but not an isolated component (especially with the note, there is no existing component). So, at the component level, we need a proper layer of isolation. The current approach of margins and max width has already been implemented in Container, where this behaviour is already encapsulated

packages/core/atoms/Grid.lite.tsx
ramonmulia: hmm good point, these maps exist because tw needs static strings to generate the css rules, so we're encoding the `var()` as value classes, it works but it is verbose.
I think a tw plugin would clean this.

packages/react/src/atoms/Grid.tsx
DmitryB-NF: I do not think we need a full plugin when we control the actual theme and all utilities in a separate public package. The simplest way is to add something like
```
@layer utilities {
  .gi-grid-columns-1  { --grid-columns: 1; }
  ...
  .gi-grid-gap-0      { --grid-gap: 0; }
  .gi-grid-gap-1      { --grid-gap: var(--gieds-space-1); }
  ...
  .gi-grid-span-1     { --grid-span: 1; }
 ...
}
```
This will cover the case the same way as it is done in `layout.css` right now. 
We can revise and extend the approach during the TW4 migration to remove all extrenal utilities consistently if required

ramonmulia: hmm yeah, that was my first thought too, but here inline styles only set css variables, they don’t control layout, layout is still handled by tailwind classes that read those variables

utility classes would add setup, we would need a new css file and imports in react and angular, like fonts.css, it feels too much for just 3–4 variables

the complex part `[&>*]:[--grid-parent-columns:var(...)]` still stays, we need it for nested grids to work,
with utility classes we need many variants, `gi-grid-columns-1` to 12, gaps, spans, breakpoints, so yeah a lot of classes, and it’s a fixed list, if someone uses `columns={16}` it just breaks

with inline styles it just works, `--grid-columns: 16` works, new spacing tokens also work automatically

tailwind v4 removed safelist, so this will change anyway later..

the current approach is simple and self-contained, no extra css files, i still think it’s better, but i’m ok to go with grid.css anyway..

DmitryB-NF: First of all, our design system is based on 12 columns max. So, it should not be any number, and `columns={16}` should be constrained to type `1 | ... | 12`. So basically do not need so many custom classes.

The entire `grid.css` file can be significantly simplified if using the `@layer utilities ` and `var()`, making the file much simpler, like:
```
@layer utilities {
  /* container & item base — written once, no media queries */
  .gi-grid-container {
    gap: var(--grid-gap, 0px);
    @apply gi-flex gi-flex-wrap;
  }
  .gi-grid-container > * {
    --grid-parent-columns: var(--grid-columns, 4);
    --grid-parent-gap: var(--grid-gap, 0px);
  }
  .gi-grid-item {
    --grid-basis: min(
      calc(var(--grid-span, 1) / var(--grid-parent-columns) * 100%
        - var(--grid-parent-gap) * (var(--grid-parent-columns) - var(--grid-span, 1))
          / var(--grid-parent-columns)),
      100%
    );
    @apply gi-shrink-0 gi-basis-[var(--grid-basis)]; // gi-grow-0 and gi-max-w-[var(--grid-basis)] will kill еntire purpose of the Flex Grid
  }

  .gi-grid-columns-1 { --grid-columns: 1; }
  .gi-grid-columns-2 { --grid-columns: 2; }
  /* … 3..12 */
  .gi-grid-span-1    { --grid-span: 1; }
  /* … 2..12 */
  .gi-grid-gap-0     { --grid-gap: 0; }
  .gi-grid-gap-1     { --grid-gap: var(--gieds-space-1); }
  /* … 2..12 */
}
  ```
Tailwind then auto-generates `md:gi-grid-span-4`, `xs:gi-grid-columns-2` and so on through its native variant system.

Also, this file should not be copied into each package but stored directly in the `design-system-tailwind` as a utility extension.

Comment (DmitryB-NF) [reply packages/react/src/atoms/Grid.tsx]: Simpler implementation of utilities
```
 .gi-grid-container {
    display: flex;
    flex-wrap: wrap;
    gap: var(--grid-gap, 0px);
  }
  .gi-grid-container > * {
    --grid-parent-columns: var(--grid-columns, 4);
    --grid-parent-gap: var(--grid-gap, 0px);
  }
  .gi-grid-item {
    flex-basis: min(
      calc(var(--grid-span, 1) / var(--grid-parent-columns) * 100%
        - var(--grid-parent-gap) * (var(--grid-parent-columns) - var(--grid-span, 1))
          / var(--grid-parent-columns)),
      100%
    );
    flex-shrink: 0;
  }
  ```

packages/react/styles.css
DmitryB-NF, line 9: Does it make sense to replace this with `grid.css`? It looks like `layout.css` is no longer used after replacing the Grid props

examples/nextjs/src/app/page.tsx
```diff
@@ -936,6 +937,27 @@ export default function Home() {
                     </div>
                   </div>
 
+                  <div>
```
DmitryB-NF: It would be really nice to use the `Box` component later on for all examples.

packages/angular/src/atoms/storybook/Grid.meta.ts
```diff
@@ -0,0 +1,311 @@
...
+        defaultValue: {
+          summary: 'false'
+        }
+      }
+    },
+    columns: {
+      control: {
+        type: 'number'
+      },
+      description: 'Override the number of columns. Accepts a number or a responsive breakpoint object `{ base?, xs?, sm?, md?, lg?, xl?, 2xl? }`. Defaults to Figma spec (4 → 4 → 6 → 8 → 12 → 12 → 12).',
```
DmitryB-NF: `Defaults to Figma spec` - I am not sure this is an accurate comment when Figma is not fully publicly available. Let's make these docs fully sufficient so they can serve as a single source of truth.

packages/angular/src/atoms/storybook/Grid.meta.ts
```diff
@@ -0,0 +1,311 @@
...
+      }
+    },
+    gap: {
+      control: {
+        type: 'number'
+      },
+      description: 'Gap between grid items in Tailwind spacing scale (0–12). Accepts a number or a responsive breakpoint object `{ base?, xs?, sm?, md?, lg?, xl?, 2xl? }`.',
+      table: {
+        type: {
+          summary: 'SpacingScale | ResponsiveValue<SpacingScale>'
```
DmitryB-NF: `SpacingScale` may not be clear here. Does it make sense to use `1 | ... | 12` with a comment that it can be responsive for breakpoints?

packages/angular/src/stories/Grid.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+};
+
+export default meta;
+
+const itemClasses = 'gi-bg-gray-300 gi-p-1 gi-flex gi-items-center gi-justify-center gi-font-primary gi-rounded';
+
+export const Default: StoryObj = {
+  ...gridDefault,
+  render: (props) => ({
+    props: { ...props, itemClasses },
```
DmitryB-NF: ⚠️ Do we really need `itemClasses` as part of props?

packages/angular/angular.json
```diff
@@ -23,7 +23,7 @@
             "tsConfig": ".storybook/tsconfig.json",
             "port": 6006,
             "compodoc": false,
-            "styles": ["fonts.css", "styles.css", ".storybook/global.css", "@ogcio/theme-govie/theme.css"]
+            "styles": ["fonts.css", "dist/styles.css", ".storybook/global.css", "@ogcio/theme-govie/theme.css"]
```
DmitryB-NF: This is a nice workaround, but it will break the Storybook deployment via Dockerfile. I added a specific webpack processing script to include the `postcss` plugin in https://github.com/ogcio/govie-ds/pull/944.
Could you review the PR and confirm it resolves the issue?

packages/angular/styles.css
```diff
@@ -1,3 +1,5 @@
+@import 'node_modules/@ogcio/design-system-tailwind/css/grid.css';
```
DmitryB-NF, line 1: ❓ Should it be after all other utils to make sure it has more priority?

packages/angular/tailwind.config.ts
```diff
@@ -1,15 +1,28 @@
...
   content: ['./src/**/*.{ts,html}', './.storybook/**/*.{ts,html}'],
   theme: createTheme(),
-  plugins: [require('@tailwindcss/typography')],
+  plugins: [typography],
   safelist: [
     { pattern: /gap-./, variants: ['xs', 'sm', 'md', 'lg', 'xl', '2xl'] },
     { pattern: /flex-row/, variants: ['xs', 'sm', 'md', 'lg', 'xl', '2xl'] },
     { pattern: /flex-col/, variants: ['xs', 'sm', 'md', 'lg', 'xl', '2xl'] },
+    {
+      pattern: /gi-grid-columns-\d+/,
```
DmitryB-NF: It can be simplified as in https://github.com/ogcio/govie-ds/pull/944/changes#diff-9b738afe8945078cf2b603cec98eec8d68df6f8c80aeccfc50301ba074c67072R14

packages/core/atoms/storybook/Grid.meta.ts
```diff
@@ -0,0 +1,269 @@
...
+      control: false,
+      description: 'Optional id for linking/targeting and aria references.',
+      table: { type: { summary: 'string' } },
+    },
+    dataTestId: {
+      control: false,
+      description: 'Test id for targeting the element in automated tests.',
+      table: { type: { summary: 'string' } },
+    },
+    role: {
```
DmitryB-NF: Would be really nice to reuse Box `argTypes`

packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,67 @@
...
+          : _.compact(['gi-grid-item', buildGridClasses(props.size, 'gi-grid-span'), props.className]).join(' ')
+      }
+      data-testid={props.dataTestId}
+    >
+      {props.children}
+    </div>
+  );
+}
+
+const FIGMA_COLUMNS: Partial<Record<BreakpointKey, SpacingScale>> = {
```
DmitryB-NF: Does it make sense to name it `default` or similar? I am not sure FIgma reference in the code make sense while this is an internal requirement

packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,67 @@
...
+  lg: 12,
+};
+
+const buildGridClasses = (value: ResponsiveValue<SpacingScale> | undefined, prefix: string): string => {
+  if (!value) {
+    return '';
+  }
+
+  const source = _.isNumber(value)
+    ? { base: Math.max(0, value) as SpacingScale }
```
DmitryB-NF: Just in case, `_.clamp(n, 0, 12)` can also work

DmitryB-NF (changes requested):
The Grid looks solid, and the Storybook demo is really nice!
We need to base it on the Box and Angular Storybook PRs to make sure the implementation is clean.

packages/react/styles.css
ramonmulia: not sure if we should, I just noticed there is a specific grid doc https://ds.services.gov.ie/foundations/design-tokens/grid/ and maybe a client is using those? 
maybe adding `gi-grid-12-column, gi-grid-8-column, gi-grid-4-column and gi-grid-responsive` into `grid.css` and update the grid docs with the new system?

packages/angular/src/stories/Grid.stories.ts
ramonmulia: no we don't, i'm going to add it directly

examples/nextjs/src/app/page.tsx
ramonmulia: yeap, these `div` were replaced by Box in Box pr

packages/angular/styles.css
ramonmulia: if I put after tailwind related it doesn't work, I'm not sure why.

apps/docs/content/3-components/2-library/grid/design.mdx
```diff
@@ -0,0 +1,173 @@
...
+description: 'Grid Component'
+status: 'stable'
+component:
+  id: 'grid'
+  status: 'stable'
+---
+
+# Grid
+
+Grid is a flex-based layout component that distributes children into columns with responsive behaviour. By default it follows the Figma grid specification (4 columns on mobile, 6 on small, 8 on medium, 12 on large), giving designers pixel-perfect layouts out of the box.
```
DmitryB-NF: _"By default it follows the Figma grid specification"_ it would be nice to avoid Figma references because Docs should be the source of truth for users

apps/docs/content/3-components/2-library/grid/design.mdx
```diff
@@ -0,0 +1,173 @@
...
+      <div className="gi-bg-gray-300 gi-p-1 gi-text-center">12</div>
+    </Grid>
+  </Grid>
+</ComponentPreview>
+
+## Variants
+
+### Automatic Mode
+
+Without any `size` prop, children distribute evenly according to the Figma column spec. Items wrap automatically as the viewport changes.
```
DmitryB-NF: _"Figma column spec"_

apps/docs/content/3-components/2-library/grid/react.mdx
```diff
@@ -0,0 +1,145 @@
...
+  status: 'stable'
+  properties:
+    - name: 'GridProps'
+      fields:
+        - name: 'container'
+          ofType: 'boolean'
+          description: 'When true, activates grid container mode with responsive Figma defaults. When false, the element acts as a grid item.'
+          required: false
+        - name: 'columns'
+          ofType: 'number | ResponsiveValue<number>'
```
DmitryB-NF: ⚠️ Does it make sense to do that `1-12` as it is in the Storybook?

packages/core/atoms/storybook/Grid.meta.ts
```diff
@@ -0,0 +1,194 @@
...
+        type: { summary: 'boolean' },
+        defaultValue: { summary: 'false' },
+      },
+    },
+    columns: {
+      control: { type: 'number' },
+      description:
+        'Override the number of columns. Accepts a number (0–12) or a responsive breakpoint object `{ base?, xs?, sm?, md?, lg?, xl?, 2xl? }`. Without this prop the container defaults to `{ base: 4, sm: 6, md: 8, lg: 12 }`.',
+      table: {
+        type: { summary: 'number (0–12) | ResponsiveValue<number>' },
```
DmitryB-NF: ❓ Should it be '1–12 | ResponsiveValue<1–12>'`? I am not sure 0 columns makes sense

examples/nextjs/src/app/page.tsx
```diff
@@ -701,7 +702,9 @@ export default function Home() {
 
                 <Box className="grid grid-cols-1 md:grid-cols-2 gap-6">
                   <Box>
-                    <h5 className="font-semibold mb-2">Radio Group</h5>
+                    <Heading as="h5" className="font-semibold mb-2">
```
DmitryB-NF, line 705: 💡 I am wondering if we can also use `H5` as more short alias

packages/core/atoms/storybook/utilities.ts
```diff
@@ -47,6 +47,13 @@ export function enumType<const T extends Readonly<Record<string, string>>>(
  * Returns an object with assertion methods for common play-function checks.
  */
 export const checker = (testId: string, canvas: ReturnType<typeof within>, step: StoryContext<Renderer>['step']) => ({
+  exists: async (tag = 'DIV') =>
```
DmitryB-NF: ❓ Does it make sense to name it `tag` or `is` to be `is('div')` (better to make it uppercase in the checker)? Because it is nore expected that `exists('div')` checks that the `testId` _is a specific tag_, but more likely includes a tag as a child

packages/core/atoms/Grid.lite.tsx
```diff
@@ -0,0 +1,50 @@
...
+export default function Grid(props: Props) {
+  return (
+    <GiBox
+      id={props.id}
+      role={props.role}
+      ariaLabel={props.ariaLabel}
+      ariaLabelledBy={props.ariaLabelledBy}
+      styles={props.styles}
+      className={
+        props.container
```
DmitryB-NF: I am wondering if we can make thу Grid item and container at the same time to improve the `Nested` example in the storybook. I think instead of checking for `container`, we can do something like
```
[
  props.container && _.compact(....),
  props.size != 'null' && `gi-grid-item ${getGridClasses(props.size, 'gi-grid-span')}`
  props.className,
].join(' ')
```
This way, we can make it much simpler to be an item by default with size and container on demand

packages/design/tailwind/css/grid.css
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  .gi-grid-container > * {
+    --grid-parent-columns: var(--grid-columns, 4);
+    --grid-parent-gap: var(--grid-gap, 0px);
+  }
+
+  .gi-grid-item {
+    --grid-item-span: initial;
+    --grid-span: var(--grid-item-span, 1);
```
DmitryB-NF: We basically have 2 variables with exactly the same sense. Does it make sense to declare only one `--grid-span`?

I think we can also remove `gi-grid-container > *` because we can use `--grid-gap` and `--grid-columns` directly, e.g.:
```
@layer utilities {
  .gi-grid-container {
    display: flex;
    flex-wrap: wrap;
    gap: var(--grid-gap, 0px);
  }

  .gi-grid-item {
    flex-basis: min(
      calc(
        var(--grid-span, 1) / var(--grid-columns, 4) * 100%
          - var(--grid-gap, 0px) * (var(--grid-columns, 4) - var(--grid-span, 1))
            / var(--grid-columns, 4)
      ),
      100%
    );
    flex-shrink: 0;
  }

  .gi-grid-columns-1  { --grid-columns: 1; }
  /* ... up to 12 */
  .gi-grid-gap-0      { --grid-gap: var(--gieds-space-0); }
  /* ... up to 12 */
  .gi-grid-span-1     { --grid-span: 1; }
  /* ... up to 12 */
}
```

DmitryB-NF: Based on https://github.com/ogcio/govie-ds/pull/939/changes#r3218788873 to make it container and items at the same time, we can do it something like
```
 @layer utilities {
   .gi-grid-container {
     display: flex;
     flex-wrap: wrap;
     gap: var(--grid-gap, 0px);
   }

   .gi-grid-item {
     flex-basis: min(
       calc(
+        var(--grid-span, 1) / var(--grid-parent-cols, 4) * 100% - var(--grid-parent-gap, 0px) *
+          (var(--grid-parent-cols, 4) - var(--grid-span, 1)) / var(--grid-parent-cols, 4)
       ),
       100%
     );
     flex-shrink: 0;
   }

+  .gi-grid-columns-1  > * { --grid-parent-cols: 1;  }
+  .gi-grid-columns-2  > * { --grid-parent-cols: 2;  }
+  /* ... up to 12 */

+  .gi-grid-gap-0  { --grid-gap: var(--gieds-space-0); }
+  .gi-grid-gap-0  > * { --grid-parent-gap: var(--gieds-space-0); }
+  /* ... up to 12 */

+  .gi-grid-span-1  { --grid-span: 1;  }
+  /* ... up to 12 */
 }
````

File: packages/react/tests/visual.spec.ts-snapshots/layout-grid--nested-webkit-linux.png
Comment (DmitryB-NF) [inline packages/react/tests/visual.spec.ts-snapshots/layout-grid--nested-webkit-linux.png:1]: ❓ Is this the expected view for the example? I suppose the nested `<Grid container/>` is not flex and uses the minimal width?

File: packages/react/styles.css
Comment (DmitryB-NF) [reply packages/react/styles.css:9]: It would be nice to get rid of this when adding the Grid component, but we should leave this for backward compatibility for now. Let's add a TODO here to remove when migrating

File: packages/core/atoms/storybook/Grid.meta.ts [hunk 24]
```diff
@@ -0,0 +1,194 @@
...
+      }
+    });
+  },
+};
+
+export const ColumnOverride = {
+  tags: ['skip-playwright'],
+  args: {
+    ...gridMeta.args,
+    columns: 3 as const,
```
DmitryB-NF: ❓ What is the reason for this case? If this is for checking that the `column` prop works as expected, I think we have a name like `CustomGrid` and add more tests for different grid props

examples/nextjs/src/app/page.tsx
ramonmulia: none of H1–H6 are currently exported from the package root, adding just H5 would be inconsistent, so this would need all heading exported together, should we export H1-H5 on this pr?

packages/react/tests/visual.spec.ts-snapshots/layout-grid--nested-webkit-linux.png
ramonmulia: refactored the nested example

packages/angular/styles.css
DmitryB-NF: Hmm, I suppose it is because of using `@tailwind` instead of `@import 'tailwindcss/base'` as in React.
I think this is correct and suggested by TW. The only concern is having a different priority set in React and Angular.

Overall, let's move this way and eventually fix React to have a fully up-to-date setup in Angular.

packages/core/atoms/storybook/Grid.meta.ts
DmitryB-NF: I think it has not been resolved or commented.

DmitryB-NF: I would suggest moving `await step('renders all *N* child items')` to the Default, where this case is already covered. So, we can remove this and add later with specific tests if any

ramonmulia: oh ok, I just changed the story's name, in Default this case is already covered , so I'm removing this from the `CustomGrid` story, I don't think we have specific tests to add here. I let visual regression cover the columns behavior.

packages/angular/src/atoms/storybook/Grid.meta.ts
```diff
@@ -0,0 +1,203 @@
...
+import type { Props } from '../Grid';
+import { boxMeta } from './Box.meta';
+import { checker } from './utilities';
+export const gridMeta = {
+  tags: ['autodocs'] as string[],
+  title: 'Layout/Grid',
+  args: {
+    container: true,
+    gap: 4 as const,
+    dataTestId: 'grid-test',
```
DmitryB-NF: I think the file was processed by prettier separately from the build. So there are differences after rebuilding.
@ramonmulia could you run a clear `build` in core to validate all the files?

packages/core/atoms/storybook/Grid.meta.ts
DmitryB-NF: ⚠️ But the story still does not make sense, even with the changed name. Exactly the same case is already covered in Default, so the Story could be removed entirely if there are no valuable unit tests

ramonmulia: removed.

DmitryB-NF (approved):
Well done, finally we have a fully encapsulated Grid 🎉

examples/nextjs/src/app/page.tsx
DmitryB-NF: It would be really nice to export `HN` components for better usage.

Overall, it is better to avoid adding unrelated changes to a PR to avoid needing to fix external code. We still need to reorganise all examples to follow a new Atoms architecture, so this will be changed later as well.

PR #941, opened 2026-05-01
feat(AB#37145): headertitle link

Ademsk1:
## Description

Adds in an optional `href` prop to the `HeaderTitle` and to the `HeaderLogo` . Includes Aria-labelling for better accessibility on the links. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#37145](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37145)


DmitryB-NF (dismissed):
@Ademsk1, this approach is not correct according to the initial requirements. As mentioned in the original investigation AB#36981, we need to use the `<a>` tag as a root for the title and header if the `href` property is provided. 
As suggested in the workaround in the comments, using `<a>` as a quick solution can resolve all issues without extra layer of classes to override the Link component behaviour

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,25 +1,40 @@
...
   if (!context) {
     throw new Error('HeaderTitle must be used within a Header');
   }
-
   return (
     <div
-      className={cn(
+      className={headerTitleVariants({
+        appearance: context.variant,
+        isLink: !!href,
```
Ademsk1: I'm unopinionated when it comes to putting the padding-2 in headerTitleVariants, or in the className directly. But searching through, we only use headerTitleVariants once, so I didn't see any harm in passing it in here

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -29,3 +44,21 @@ Object.defineProperty(HeaderTitle, 'componentType', {
   writable: false,
   enumerable: false,
 });
+
+const styles = tv({
+  base: [
+    'gi-truncate gi-py-1 gi-max-w-fit gi-block gi-w-full gi-underline gi-p-1',
```
Ademsk1: Please note, I have added a `gi-p-1` which has not been formally agreed upon. However, the focus state outline was touching the text, and so I added this here to make it look less compact.

packages/react/src/header/variants.ts
```diff
@@ -65,6 +65,10 @@ export const headerTitleVariants = tv({
   base: 'gi-header-title',
   variants: {
     appearance: AppearanceText,
+    isLink: {
+      true: 'gi-p-2',
```
Ademsk1: Unlike the other `gi-p-1`, this is important to allow the outline to not be cropped out by the HeaderTitle div.

packages/react/src/header/header-next/components/header-logo.tsx
```diff
@@ -1,18 +1,32 @@
...
   if (!context) {
     throw new Error('HeaderLogo must be used within a Header');
   }
 
   return (
-    <div className={headerLogoVariants({ appearance: context?.variant })}>
-      {children}
+    <div
+      className={cn(
+        !!href && 'gi-p-1',
```
Ademsk1: again, adding a `gi-p-1` to give it more space. This padding on both the logo and title only exists when an href is passed into them.

packages/react/src/header/header-next/components/header-logo.tsx
```diff
@@ -1,18 +1,32 @@
...
+        headerLogoVariants({
+          appearance: context?.variant,
+        }),
+      )}
+    >
+      {href ? (
+        <a href={href} aria-label={ariaLabel}>
```
DmitryB-NF: I do not think we need to wrap `<a>` into `<div>`, we simply use one node per props like:
```
{!href && <div className={headerLogoVariants}>{children}</div>}
{!!href && <a className={headerLogoVariants} href={href}>{childrent}</a>}
```
This way, there will be clear styling and nesting of nodes

packages/react/src/header/header-next/components/header-title.tsx
DmitryB-NF: See the approach of a single node in https://github.com/ogcio/govie-ds/pull/941/changes#r3227446600. This way, you do not need an additional parameter for styles or you can use different `tv` functions for `<div>` and `<a>`

packages/react/tests/visual.spec.ts-snapshots/layout-header--with-title-and-logo-as-links-android-linux.png
DmitryB-NF, line 1: ⚠️ Title should not have underline styling

DmitryB-NF (changes requested):

packages/react/src/header/header-next/components/header-logo.tsx
Ademsk1: Ah I forgot to do this for header-logo! Will update now

packages/react/src/header/header-next/components/header-logo.tsx
```diff
@@ -2,15 +2,34 @@
 import type { HeaderLogoProps } from '../../types.js';
 import { headerLogoVariants } from '../../variants.js';
 import { useHeaderContext } from '../header-context.js';
+import { cn } from '../../../cn.js';
```
ramonmulia: should we import clsx directly?

`import clsx from 'clsx'`

packages/react/src/header/header-next/components/header-logo.tsx
```diff
@@ -2,15 +2,34 @@
...
     throw new Error('HeaderLogo must be used within a Header');
   }
-
-  return <div className={headerLogoVariants({ appearance: context?.variant })}>{children}</div>;
+  if (href) {
+    return (
+      <a
+        href={href}
+        className={headerLogoVariants({ appearance: context.variant, className: 'gi-p-1 gi-block' })}
+        aria-label={ariaLabel}
```
ramonmulia: as the logo link typically points to the homepage what do you think to add `aria-current` here?

DmitryB-NF: This is a good note, but I think we should implement it as part of the Mitosis migration. The current API may be changed, where the `ariaLabel` may be part of the content (Image/Logo) rather than the link itself
@Ademsk1 could you revert this one to avoid discrepancies?

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,22 +1,87 @@
...
   if (!context) {
     throw new Error('HeaderTitle must be used within a Header');
   }
-
-  return <div className={cn(className, headerTitleVariants({ appearance: context.variant }))}>{children}</div>;
+  if (href) {
+    return (
+      <a
+        href={href}
+        aria-label={ariaLabel}
```
DmitryB-NF: `arialLabel` is excessive when there is an atual Label in children

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,22 +1,87 @@
...
 };
 
 Object.defineProperty(HeaderTitle, 'componentType', {
   value: 'HeaderTitle',
   writable: false,
   enumerable: false,
 });
+
+const styles = tv({
+  base: ['gi-heading-sm gi-ml-4 md:gi-ml-6 lg:gi-ml-12 gi-min-w-0 lg:gi-block', 'gi-truncate gi-w-full gi-p-1'],
```
DmitryB-NF: ⚠️ Does it make sense to add it as a simple string 
```suggestion
  base: 'gi-heading-sm gi-ml-4 md:gi-ml-6 lg:gi-ml-12 gi-min-w-0 lg:gi-block gi-truncate gi-w-full gi-p-1',
```
or split all values by array for readability?

```suggestion
  base: [
  'gi-heading-sm',
  'gi-ml-4 md:gi-ml-6 lg:gi-ml-12', 
  'gi-min-w-0 gi-w-full', 
  'gi-truncate', 
  'lg:gi-block', 
  'gi-p-1', 
],,
```

The mixture is not really readable

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -1,6 +1,9 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
+import type { IconId } from '../../icon/icon.js';
```
DmitryB-NF: Could we avoid adding `.js` imports in all new code?

packages/react/src/header/header-next/components/header-logo.tsx
Ademsk1: I don't think so... looking at the docs:
"
When you have a group of related elements, such as several links in a breadcrumb or steps in a multi-step flow, with one element in the group styled differently from the others to indicate to the sighted user that this is the current element within its group, the aria-current should be used to inform the assistive technology user what has been indicated via styling.
"
We don't have an indication via styling for the logo that there's a "current" element, like in breadcrumbs for example. Lmk if you disagree.

packages/react/src/header/header-next/components/header-logo.tsx
Ademsk1: i actually don't need this import anymore

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +528,118 @@ export const Light: StoryObj = {
...
+      expect(header).toBeInTheDocument();
+      const links = await canvas.findAllByRole('link');
+      const linkLogo = links[0];
+      const linkTitle = links[1];
+      expect(linkLogo).toHaveAttribute('aria-label', 'Gov.ie logo');
+      expect(linkTitle).toHaveAttribute('aria-label', 'Title as a link');
+      expect(linkLogo).toHaveAttribute('href', '#');
+      expect(linkTitle).toHaveAttribute('href', '#');
+      await userEvent.click(header);
+      await userEvent.tab();
```
DmitryB-NF: ⚠️ we should avoid adding stories with actions to Storybook to avoid flaky tests. Can we add one visual static story and one for E2E tests excluded from Playwright?

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,22 +1,87 @@
 'use client';
 
-import { cn } from '../../../cn.js';
 import type { HeaderTitleProps } from '../../types.js';
 import { headerTitleVariants } from '../../variants.js';
 import { useHeaderContext } from '../header-context.js';
+import { tv } from 'tailwind-variants';
 
-export const HeaderTitle = ({ children, className }: HeaderTitleProps) => {
+export const HeaderTitle = ({ children, href, className, ariaLabel }: HeaderTitleProps) => {
```
DmitryB-NF: ⚠️ `HeaderTitleProps` includes `ComponentPropsWithoutRef<'div'>` which may not be correct for `<a>`. Does it make sense to remove this from the type, as it is not used anyway?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +528,118 @@ export const Light: StoryObj = {
...
   },
 };
+
+const LinkExamples = ({ focused }: { focused?: boolean }) => {
+  const [variant, setVariant] = useState<HeaderAppearance>('default');
+  const toggleVariant = () => setVariant(variant === 'light' ? 'default' : 'light');
+  return (
+    <div className={cn('gi-p-4', variant === 'light' ? 'gi-bg-black' : 'gi-bg-white')}>
+      <Header variant={variant} aria-label="Site header">
+        <HeaderLogo href="#" ariaLabel="Gov.ie logo">
```
DmitryB-NF: `ariaLabel` is exessvie here as it is in the Logo itself

packages/react/src/header/header-next/header-next.tsx
```diff
@@ -52,8 +52,10 @@ export const HeaderNext = forwardRef<HTMLElement, HeaderNextProps>(
         >
           <Container className="gi-order-2 gi-min-w-0">
             <div className={headerMenuVariants({ appearance: variant })}>
-              {headerLogo}
-              {headerTitle}
+              <div className="gi-flex gi-items-center gi-min-w-0">
+                {headerLogo}
+                {headerTitle}
+              </div>
```
Ademsk1: see here for where I put couple the logo and the title. Because the header menu uses justify-between, this felt like the cleanest way to say "in this header, logo and title always sit next to each other"

packages/react/src/header/header-next/components/header-logo.tsx
ramonmulia: hmm interesting, ok then

packages/react/tests/visual.spec.ts-snapshots/layout-header--default-webkit-linux.png
Ademsk1, line 1: I see somethings happened with the default header here. I'll wait for the response from the pipeline to identify where it is...

packages/react/src/header/header-next/header-next.tsx
DmitryB-NF: ❓ Does it make sense to wrap only the title here to make sure it flexes properly? It is still possible to have Title as `div` inlide to make sure styles applied properly without additional grouping

DmitryB-NF (dismissed):

packages/react/src/header/header-next/header-next.tsx
DmitryB-NF: I think we can remove the `justify-between` to make sure we have a clear display pattern

packages/react/src/header/header-next/header-next.stories.tsx
Ademsk1: The Logo does have an aria-label, however, this aria-label is for describing the image. We want to be able to describe the link (i.e. where it takes us).

DmitryB-NF: But why do we need to describe the link? Aria is required for screen readers to identify content. In the current example, the screen reader will say something like _"Gov.ie logo link, Gov.ie logo image"_, which seems excessive

packages/react/src/header/header-next/components/header-logo.tsx
DmitryB-NF: @Ademsk1 we still need to revert this

packages/react/src/header/header-next/header-next.tsx
Ademsk1: wrapped just the headerTitle

apps/docs/src/components/navigation/header.tsx
```diff
@@ -58,7 +58,7 @@ export const DocsHeader = () => {
           <span className="gi-sr-only">Gov.ie logo</span>
         </HeaderLogo>
 
-        <HeaderTitle id="site-title">Design System</HeaderTitle>
```
Ademsk1, line 61: removed the id prop as it isn't used in the component

packages/react/src/header/header-next/header-next.tsx
Ademsk1: brought out the styles from the CSS into tailwind for `headerMenuVariants` and omitted justify-between. This way, `justify-between` will still work for HTML versions. `justify-between` added in Header-Legacys menu div to maintain visual consistency. Let me know if this feels the right way to go about it.

packages/react/tests/visual.spec.ts-snapshots/layout-header--default-webkit-linux.png
Ademsk1: fixed

packages/react/src/header/variants.ts
```diff
@@ -18,7 +18,7 @@ export const headerVariants = tv({
 });
 
 export const headerMenuVariants = tv({
-  base: 'gi-header-menu',
+  base: ['gi-h-20', 'gi-items-center', 'gi-flex', 'gi-flex-row', 'gi-flex-nowrap', 'gi-py-3', 'sm:gi-py-4'],
```
Ademsk1, line 21: brought out the styles here so we can continue to use `justify-between` in the header menu for HTML.

packages/react/src/header/header-legacy.tsx
```diff
@@ -181,7 +181,7 @@ export function HeaderLegacy({
       {...props}
     >
       <Container id="HeaderContainer" className="gi-order-2 gi-min-w-0">
-        <div className={headerMenuVariants({ appearance })}>
+        <div className={headerMenuVariants({ appearance, class: 'gi-justify-between' })}>
```
Ademsk1, line 184: manually added the justify-between here in the `HeaderLegacy` to maintain visual consistency.

packages/react/src/header/header-next/header-next.tsx
```diff
@@ -53,7 +53,7 @@ export const HeaderNext = forwardRef<HTMLElement, HeaderNextProps>(
           <Container className="gi-order-2 gi-min-w-0">
             <div className={headerMenuVariants({ appearance: variant })}>
               {headerLogo}
-              {headerTitle}
+              <div className="gi-flex-1 gi-min-w-0 gi-pr-2">{headerTitle}</div>
```
Ademsk1, line 56: There was not a lot of space between the menu buttons and the ellipsis - so I added a pr-2 here.

packages/react/src/header/header-next/header-next.stories.tsx
Ademsk1: skip-playwright added here.

packages/design/tailwind/css/components/header.css
```diff
@@ -55,13 +55,12 @@ focus:gi-bg-black
 }
 
 .gi-header-menu {
-  @apply gi-h-20 gi-items-center 
+  @apply gi-h-20 gi-items-center
```
DmitryB-NF: Can we revert this change? It will cause the Tailwind lib to be repuboished without any valuable change

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +528,115 @@ export const Light: StoryObj = {
...
+          </HeaderMenuItemLink>
+          <HeaderMenuItemSeparator />
+          <HeaderMenuItemButton>FAQ</HeaderMenuItemButton>
+          <HeaderMenuItemButton>Search</HeaderMenuItemButton>
+
+          <HeaderMenuItemButton showItemMode="desktop-only">Language</HeaderMenuItemButton>
+
+          <HeaderMenuItemButton onClick={toggleVariant} showItemMode="always">
+            {variant === 'default' ? 'Light' : 'Dark'}
+          </HeaderMenuItemButton>
```
DmitryB-NF: ❓ Does it make sense to simply display 2 headers in different variants without switching to make sure the example is fully sufficient and can be covered by Playwright?

packages/react/src/header/types.ts
```diff
@@ -179,11 +181,10 @@ export type HeaderSlotContainerProps = PropsWithChildren<
...
 >;
-export type HeaderTitleProps = PropsWithChildren<
-  {
-    className?: string;
-  } & ComponentPropsWithoutRef<'div'>
->;
+export type HeaderTitleProps = PropsWithChildren<{
+  className?: string;
+  href?: string;
+}>;
```
DmitryB-NF: ⚠️ This is a breaking change to remove all types. It is better to replace it with basic HTML attributes

Ademsk1: Actually, I'm going to replace it back the way it was - using `ComponentPropsWithoutRef<'div'>`. 
Reason being : 
`<div>`'s attributes are all `global` attributes. This means that we can pass in any `div` prop, and it would work for the `a` tag too. Not that any other props will actually exist in the DOM. I think this fix is fine, as we will migrate to mitosis anyways.

packages/react/src/header/header-next/header-next.stories.tsx
Ademsk1: showing both in 1 story now for each.

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +529,138 @@ export const Light: StoryObj = {
...
+  parameters: {
+    ...dualHeaderA11yParameters,
+    docs: {
+      description: {
+        story: 'By passing in an `href` prop to the `HeaderTitle` component, the title becomes a link.',
+      },
+    },
+  },
+  argTypes: {
+    variant: {
```
ramonmulia: not sure if argTypes is used here on this story since the render function always renders `<LinkExamples />` without reading args, I would remove it.

packages/react/src/header/types.ts
```diff
@@ -179,11 +181,11 @@ export type HeaderSlotContainerProps = PropsWithChildren<
     className?: string;
   } & ComponentPropsWithoutRef<'div'>
 >;
-export type HeaderTitleProps = PropsWithChildren<
-  {
-    className?: string;
-  } & ComponentPropsWithoutRef<'div'>
->;
+export type HeaderTitleProps = PropsWithChildren<{
```
ramonmulia, line 185: hmm when `href` is provided, the component renders `<a>`, but the type still extends `ComponentPropsWithoutRef<'div'>` so is semantically incorrect. I believe `<div>` attr don't include some `<a>` attrs like `href`.
could you check with this is correct?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +529,138 @@ export const Light: StoryObj = {
     );
   },
 };
+
+const LinkExamples = ({ focused }: { focused?: boolean }) => {
```
ramonmulia: the `focused` prop here only changes the text content, it doesn't trigger focus state, not sure but I think the name is misleading.. maybe rename it to something like `showFocusLabel`?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -25,6 +27,7 @@ import { HeaderMenuItemSlot } from './components/menu/components/header-menu-ite
 import { HeaderPrimaryMenu } from './components/menu/header-primary-menu.js';
 import { HeaderSecondaryMenu } from './components/menu/header-secondary-menu.js';
 import { HeaderNext as Header, HeaderSlotContainer } from './header-next.js';
+import { F } from 'node_modules/@faker-js/faker/dist/airline-CLphikKp.js';
```
ramonmulia: acidental import?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -1,4 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
+import clsx from 'clsx';
```
ramonmulia: I think `clsx` is not used on this file.

packages/react/src/header/types.ts
Ademsk1: Semantically yes, you're right. However, `div` props are global - meaning that they exist in `a` as well. Regardless, this is a temporary typescript fix to avoid rebuilding the docs (id is used in a HeaderTitle example, likely by mistake). The only props actually implemented in the DOM are `href`, `className` and `children` - and these are the minimum viable props necessary for the title to be a functioning title or link with style changes.

`href` is added manually in in this type, so we don't need to worry there. 

I think this way is fine for now. When we migrate this component to mitosis, this issue will not exist. Also in mitosis we can revisit the structure of HeaderTitle and consider what is the best way to implement it.

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,22 +1,93 @@
...
   }
-
-  return <div className={cn(className, headerTitleVariants({ appearance: context.variant }))}>{children}</div>;
+  if (href) {
+    return (
+      <a
+        href={href}
+        className={styles({
+          appearance: context.variant,
+          link: true,
```
DmitryB-NF: It seems like `link` is always true here. Does it make sense to simplify it and add those tiles to the base?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -526,3 +526,138 @@ export const Light: StoryObj = {
     );
   },
 };
+
+const LinkExamples = ({ showFocusedLabel }: { focused?: boolean }) => {
```
DmitryB-NF: ⚠️ types are invalid here. We need to use our `focused` parameter, or change the typing to `showFocusedLabel`.

packages/react/src/header/types.ts
DmitryB-NF: ⚠️ Types are invalid, and we should not ship invalid typings into production. The actual used type here is `AnchorHTMLAttributes<HTMLAnchorElement>`, but we can use the most common type `HTMLAttributes<HTMLElement>` to make sure it processes `div` and `a` with used props like

DmitryB-NF (changes requested):
The Header Title looks much more usable now!
We need to fix the typings and simplify the styles, so we are good to go

packages/react/src/header/header-next/components/header-title.tsx
```diff
@@ -1,22 +1,93 @@
...
+    {
+      link: true,
+      appearance: 'light',
+      class: [
+        'focus:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
+        'focus-visible:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
+      ],
+    },
+    {
+      link: true,
```
DmitryB-NF: As link is always `true`, we can simplify it like 
```
const styles = tv({
  base: [
    'gi-heading-sm',
    'gi-ml-4 md:gi-ml-6 lg:gi-ml-12',
    'gi-min-w-0 gi-max-w-fit',
    'gi-block',
    'gi-truncate',
    'gi-p-1',
    'focus:gi-rounded-sm',
    'focus-visible:gi-rounded-sm focus-visible:gi-no-underline focus-visible:gi-outline-none',
  ],
  variants: {
    appearance: {
      default: [
        'gi-text-white gi-stroke-white',
        'focus:gi-shadow-[0_0_0_2px_var(--gieds-color-yellow-400)]',
        'focus-visible:gi-shadow-[0_0_0_2px_var(--gieds-color-yellow-400)]',
      ],
      light: [
        'gi-text-gray-950',
        'focus:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
        'focus-visible:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
      ],
    },
  },
});
```

DmitryB-NF (approved):

PR #942, opened 2026-05-01
feat(AB#37418): Adding box atom component

ramonmulia:
## Description

Adding box atom component 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues
[37418](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37418)


examples/nextjs/src/app/page.tsx
```diff
@@ -913,6 +914,13 @@ export default function Home() {
                     </div>
                   </div>
 
+                  <Box>
+                    <h5 className="font-semibold mb-2">Box</h5>
```
DmitryB-NF: Can we use `Heading` or `H5` to make sure we rely fully on the Design System components?

packages/core/atoms/Box.lite.tsx
```diff
@@ -0,0 +1,18 @@
+import { useMetadata } from '@builder.io/mitosis';
+
+useMetadata({ angular: { selector: 'gi-box' } });
+
+export type Props = {
+  className?: string;
+  children?: any;
+  id?: string;
+  dataTestId?: string;
+};
```
DmitryB-NF, line 14: ⚠️ Does it make sense to add default `role`, `aria*, `tabIndex`, `hidden`, etc. props the same as for all Layout components? We can also create a `LayoutBase` interface to make sure all props follow the same consistent structure of a11y props

`styles` would also be nice to have to make sure we can use the default styling if necessary

packages/core/atoms/storybook/Box.meta.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    await step('does not interfere with child focus order', async () => {
+      const element = canvas.getByTestId('box-test');
+      expect(element).not.toHaveAttribute('tabindex');
+    });
+
+    await step('has no implicit ARIA role', async () => {
+      const element = canvas.getByTestId('box-test');
+      expect(element).not.toHaveAttribute('role');
+    });
```
DmitryB-NF: It would be nice actually to test that passing `aria` props works as expected

packages/react/src/stories/Box.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </Box>
+      <Box>
+        <p className="gi-text-sm gi-font-bold gi-mb-2">gi-m-4 gi-p-4</p>
+        <Box className="gi-m-4 gi-p-4 gi-bg-gray-100">Margin + Padding</Box>
+      </Box>
+    </Box>
+  ),
+};
+
+export const Background: Story = {
```
DmitryB-NF: ⚠️ This Story basically tests the Taiwind styling rather than the Layout. I do not think it is really useful in the context of the Box. It would be nice to focus on Layout where it make sense to demonstrate where Box can be natively used within other Layout Components

apps/docs/content/3-components/2-library/box/design.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+<ComponentPreviewLayout>
+  <ComponentPreviewItem className="h-auto w-full">
+    <Box className="gi-p-4 gi-bg-gray-100">
+      Padding 4
+    </Box>
+  </ComponentPreviewItem>
+</ComponentPreviewLayout>
+
+### Background Wrapper
```
DmitryB-NF: ⚠️ Background is not really a specification of the Box if it has no props. It is possible to add a background styling to any Layout component, and this doc could be misleading into thinking that the Box should be used for the background. 
Overall, it is possible to have this Doc simple and focused on Layout only. E.g. add an example of Comtainer + Box for layout demo

DmitryB-NF (changes requested):
The Box component could be extended to be consistent with other Layout components.
It would be nice to make Documentation focused on Layout primarily

packages/core/atoms/Stack.lite.tsx
```diff
@@ -35,7 +27,7 @@ export default function Stack(props: Props) {
         wrap: props.wrap ?? false,
         class: [
           resolveResponsive(props.direction ?? Direction.COLUMN, directionToClass),
-          resolveResponsive(props.gap ?? 0, (gap: number, prefix: string): string => `${prefix}gi-gap-${gap}`),
+          resolveResponsive(props.gap ?? 0, gapToClass),
```
ramonmulia, line 32: added back due a JIT problem.
angular JIT compiler can't parse ts type annotations inside template expressions..

DmitryB-NF: Hmm, this should be added in a separate PR to solve the issue before the release. We should not fix `main` branch issues in the unrelated PRs

DmitryB-NF:
@ramonmulia could you make sure all PRs follow the same convention as commits?

examples/nextjs/src/app/page.tsx
```diff
@@ -913,6 +906,15 @@ export default function Home() {
...
                   </div>
 
+                  <Box>
+                    <Heading as="h5" className="gi-font-semibold gi-mb-2">
+                      Box
+                    </Heading>
+                    <Box className="gi-p-4 gi-bg-gray-100">
+                      <Box className="gi-p-2 gi-bg-emerald-700 gi-text-white gi-mb-2">Box Item</Box>
+                    </Box>
+                  </Box>
```
DmitryB-NF: ⚠️ I do not think we need this example externally. It is better to use Box everywhere instead of a pure `div` on the page to make a proper demo for the Layout.
We can add this in this PR or move this work to a new PR to make sure the demo page is fully redesigned

packages/core/atoms/Box.lite.tsx
```diff
@@ -0,0 +1,22 @@
+import { useMetadata } from '@builder.io/mitosis';
+import type { LayoutBaseProps } from './constants';
+
+useMetadata({ angular: { selector: 'gi-box' } });
+
+export type Props = LayoutBaseProps;
```
DmitryB-NF: Does it make sense to use `BoxProps` as a pivot for all Layout components? We may not need extranal interface when the Box is already a good example of Base Layout component

packages/core/atoms/storybook/Box.meta.ts
```diff
@@ -0,0 +1,119 @@
...
+      description:
+        'Points to the id of an element that labels the container. Preferred over `ariaLabel` when a visible heading exists. Only applied when `role` is set. Maps to `aria-labelledby`.',
+      table: { type: { summary: 'string' } },
+    },
+    styles: {
+      control: false,
+      description:
+        'Inline styles applied directly to the container element. Use for truly dynamic values that cannot be expressed as Tailwind classes.',
+      table: { type: { summary: 'Record<string, string>' } },
+    },
```
DmitryB-NF, line 55: I am wondering if we can make this block generic across all Layout components, as we did for Layout/Box props. Basically, all Layout components extend Box, and we can reuse Box storybook meta as a source of `argTypes` for all Layout components

packages/core/atoms/storybook/Box.meta.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    await step('renders role attribute', async () => {
+      const element = canvas.getByTestId('box-test');
+      expect(element).toHaveAttribute('role', 'region');
+    });
+
+    await step('renders aria-label when role is set', async () => {
+      const element = canvas.getByTestId('box-test');
+      expect(element).toHaveAttribute('aria-label', 'Example region');
+    });
```
DmitryB-NF: The same set of tests can also be used the same way as `argTypes` for https://github.com/ogcio/govie-ds/pull/942/changes#r3195083768

DmitryB-NF (changes requested):
The component looks good!
It would be nice if we could improve reusage of props/sotrybook meta across Core components to make it cleaner

packages/react/src/box/box.tsx
```diff
@@ -0,0 +1,17 @@
+import GiBox, { type Props as GiBoxProps } from '../atoms/Box.js';
```
DmitryB-NF: Let's follow the new convention to use `src/Box.tsx` in React components as well. I do not think having the `box` folder for the component itself make sense

packages/react/src/box/box.tsx
```diff
@@ -0,0 +1,17 @@
...
+export type BoxProps = Omit<GiBoxProps, 'children'> &
+  React.PropsWithChildren<{
+    style?: React.CSSProperties;
+  }>;
+
+export function Box({ style, styles, ...props }: BoxProps) {
+  return (
+    <GiBox
+      {...props}
+      styles={(style ?? styles ?? {}) as Record<string, string>}
```
DmitryB-NF: `?? {}` is not really required.  `undefined` should be processed correctly

packages/react/src/box/box.tsx
```diff
@@ -0,0 +1,17 @@
...
+    />
+  );
+}
+
+Box.displayName = 'Box';
```
DmitryB-NF: ⚠️ This is excessive as the `function` already has the `Box` name

packages/react/src/container/container.tsx
```diff
@@ -21,34 +21,27 @@ type ContainerProps = Omit<GiContainerProps, 'children'> &
...
-    </GiContainer>
+      {...props}
+      className={containerStyles({ insetTop, insetBottom, class: className })}
+      styles={(style ?? styles ?? {}) as Record<string, string>}
+    />
   );
 }
 
-const styles = tv({
+const containerStyles = tv({
```
DmitryB-NF: 💡 to be consistent across all components, it could be `getClasses` or `useClasses` while thius generates a `className` string

packages/core/atoms/storybook/Box.meta.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    await check.attributes({ 'aria-label': args.ariaLabel, role: args.role, id: args.id });
+    await check.children();
+  },
+};
+
+export const WithContainerAndStack = {
+  args: {
+    ...boxMeta.args,
+    dataTestId: 'box-layout',
```
DmitryB-NF: ⚠️ it seems like it is never used in the actual props

packages/react/src/stories/Box.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const WithContainerAndStack: Story = {
+  ...boxWithContainerAndStack,
+  render: () => (
+    <Container>
+      <Stack direction="row" gap={4}>
+        <Box
+          className="gi-p-4 gi-bg-gray-300 gi-flex-1"
+          dataTestId="box-layout"
```
DmitryB-NF: It does not seem to be used

packages/react/src/stories/Container.stories.tsx
```diff
@@ -13,16 +13,21 @@ import {
 const meta: Meta<typeof Container> = {
   ...containerMeta,
   title: 'Layout/Container',
-  component: Container,
```
DmitryB-NF, line 16: I think we can still use `component` to simplify at least 2 stories. You can even try to do like this to avoid object duplication:

```
export const Default: Story = defaultStory;
export const WithInset: Story = withInset;
```

packages/react/src/stories/Stack.stories.tsx
```diff
@@ -153,6 +147,7 @@ export const Responsive: Story = {
       dataTestId={dataTestId}
       ariaLabel={ariaLabel}
       role={role}
+      id={id}
```
DmitryB-NF: Replace with `{ ...props }`?

packages/core/atoms/storybook/Box.meta.ts
```diff
@@ -0,0 +1,91 @@
...
+        'Accessible label for the container. Use when the component has a `role` prop to provide an accessible name for the landmark. Maps to `aria-label`.',
+      table: { type: { summary: 'string' } },
+    },
+    ariaLabelledBy: {
+      control: { disable: true },
+      description:
+        'Points to the id of an element that labels the container. Preferred over `ariaLabel` when a visible heading exists. Only applied when `role` is set. Maps to `aria-labelledby`.',
+      table: { type: { summary: 'string' } },
+    },
+    style: {
```
DmitryB-NF: ⚠️ `style` cannot be used as a prop in Mitosis. It should be `styles` only in the `Core` package. We can override it in React but it should be on the React side only.

DmitryB-NF (changes requested):
The Box components and inheritance in Layout components with utils look great!
We need to close small gaps to go 🚀

packages/react/src/stories/Container.stories.tsx
ramonmulia: hmm when the component is set in meta, sb auto inferrs `argTypes` and merges them with the manual ones... this brings unwanted `Container` wrapper props into controls, so removing it keeps `argTypes` fully controlled by meta and syncs angular and react sb.

DmitryB-NF: It is possible to control the list of props with `parameters`, but this seems reasonable and minimal change

packages/react/src/stories/Box.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  boxMeta,
+  Default as boxDefault,
+  WithContainerAndStack as boxWithContainerAndStack,
+} from '../atoms/storybook/Box.meta';
+
+const meta: Meta<typeof Box> = {
+  ...boxMeta,
+  title: 'Layout/Box',
+  argTypes: {
+    ...omit(boxMeta.argTypes, 'styles'),
```
DmitryB-NF, line 16: I am wondering if we can still use both and mark one as an alias

DmitryB-NF (approved):

PR #953, opened 2026-05-12
feat(AB#37422): Adding Divider mitosis component

ramonmulia:
## Description

Adding Divider mitosis component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#37422](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37422)


packages/react/src/section-break/section-break.stories.tsx
```diff
@@ -41,7 +42,7 @@ export const Small: Story = {
     await step('should have correct text size classes for "sm"', async () => {
       const canvas = within(canvasElement);
       const separatorElement = canvas.getByRole('separator');
-      expect(separatorElement).toHaveClass('gi-section-break-sm');
+      expect(separatorElement).toHaveClass('gi-m-0');
```
DmitryB-NF: Those tests do not bring a lot of value, especially with Playwright coverage, and it is better to remove them instead

packages/core/atoms/Divider.lite.tsx
```diff
@@ -0,0 +1,65 @@
...
+        inset: props.inset ?? false,
+        className: props.className,
+      })}
+      style={props.styles}
+    />
+  );
+}
+
+const dividerStyles = tv({
+  base: 'gi-border-gray-300 gi-border-0',
```
DmitryB-NF: ❓ The previous SectionBreak default was `400`. Is it intentional?

packages/core/react.config.ts
```diff
@@ -2,7 +2,7 @@ import { MitosisConfig } from '@builder.io/mitosis';
 import annotation from './plugins/annotation';
 import prettierFormat from './plugins/prettier-format';
 import fixForwardRef from './plugins/forward-ref';
-import normalizeReactSvgAttrs from './plugins/normalize-svg-attrs';
+import normalizeReactSvgAttributes from './plugins/normalize-svg-attrs';
```
DmitryB-NF, line 5: ⚠️ It would bу really nice to avoid any side changes unrelated to the PR goal. We can clean up the same file and functions, but better to move any external changes to a dedicated PRs

packages/react/src/section-break/section-break.tsx
```diff
@@ -1,39 +1,30 @@
...
+  className,
+  'data-testid': dataTestId,
+}: SectionBreakProps) {
   return (
-    <hr
-      className={cn(marginClass, color)}
-      role="separator"
-      aria-label={t('sectionBreak.sectionBreak', {
-        defaultValue: 'Section Break',
-      })}
```
DmitryB-NF, line 35: The Arial label for the divider is excessive and has been removed. Could you clean up the `sectionBreak.sectionBreak` from i18n as unused?

packages/react/src/stories/Divider.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  render: (props) => (
+    <Stack direction="row" gap={4} className="gi-h-20">
+      <Box className="gi-p-4">Left</Box>
+      <Divider {...props} />
+      <Box className="gi-p-4">Right</Box>
+    </Stack>
+  ),
+};
+
+export const Inset: Story = {
```
DmitryB-NF: Does it make sense to add a vertical inset as well?

packages/react/src/stories/Divider.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ),
+};
+
+export const Vertical: Story = {
+  ...dividerVertical,
+  render: (props) => (
+    <Stack direction="row" gap={4} className="gi-h-20">
+      <Box className="gi-p-4">Left</Box>
+      <Divider {...props} />
+      <Box className="gi-p-4">Right</Box>
```
DmitryB-NF: Does it make sense to add a horizontal example with some content to make it clearer how to use? The vertical is really valuable

packages/react/tests/visual.spec.ts-snapshots/layout-divider--vertical-android-linux.png
DmitryB-NF, line 1: I think the example is slightly misaligned. Can we make the divider size to match the content, or fix it if required?

packages/react/src/section-break/section-break.tsx
```diff
@@ -1,39 +1,30 @@
...
+};
+
+/** @deprecated Use `<Divider>` instead. */
+export function SectionBreak({
+  size = 'sm',
+  color = 'gi-border-gray-400',
+  id,
+  style,
+  className,
+  'data-testid': dataTestId,
```
DmitryB-NF: Not related, but can we add support for `dataTestId` as well to make surу all components are migrated to the camelCase attributes?

packages/react/src/footer/footer.stories.tsx
```diff
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
 import { within, expect } from 'storybook/test';
 import { Icon } from '../icon/icon.js';
 import { Link } from '../link/link.js';
-import { SectionBreak } from '../section-break/section-break.js';
+import { Divider } from '../Divider';
```
DmitryB-NF: It would be nice to use default imports to improve readability

packages/react/src/stories/Divider.stories.tsx
ramonmulia: hmm for horizontal inset we use `gi-w-11/12` which maps cporrectly to the `12-column` grid, for vertical inset there's no equivalent reference, so I'm not sure what the height value should be maybe `gi-h-[91.666667%]`?
or we could leave inset as horizontal only..

packages/core/react.config.ts
ramonmulia: for some reason, this file was failing when lint-staged ran on the branch, this change only updates the naming.. no functionality was added or modified.

packages/core/atoms/Divider.lite.tsx
ramonmulia: the `gi-border-gray-300` was in task Dev Notes, so I'm assuming 300 is the correct one?

packages/react/src/footer/footer.stories.tsx
ramonmulia: currently we don't use default exports on react wrappers (Box.tsx, Heading.tsx, etc.) all use named exports only, should we adopt default exports as the standard pattern across all wrappers?

packages/core/atoms/Divider.lite.tsx
DmitryB-NF: I think I added this based on values, but not the implementation. Let's make sure it matches Figma design to make sure the default behaviour is not changed

packages/core/react.config.ts
DmitryB-NF: Lint staged checks only the diff files (files added/changed in Git). It is possible that there was a minor space/accidental addition of the file to the git stage.
We can leave it as it is, but better to check the local setup to avoid similar issues

packages/react/src/stories/Divider.stories.tsx
DmitryB-NF: I am not sure that the horizontal-only approach is a good solution for consistency. The quickest option is to use `gi-h-5/6` (exists in the default TW Theme), or we can replace the implementation to use margins instead, like `gi-mx-4` + `gi-my-4` for different screens, as it is better.

DmitryB-NF: I think that the best and easiest approach is
```
{ orientation: HORIZONTAL, inset: true, class: 'gi-w-11/12 gi-mx-auto' },  // 91.7%
{ orientation: VERTICAL,   inset: true, class: 'gi-self-center gi-h-5/6' }, // 83.3%
```
This way, it will be consistent for all types of screens and visually noticeable from the Design perspective
cc: @LexaGallery

ramonmulia: After talked with Lexa we took out inset prop.

apps/docs/content/3-components/2-library/divider/design.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+### Do
+
+- Use Divider to create a thematic break between sections of content.
+- Control spacing around the Divider using the parent layout's gap or padding.
+- Use `className` to customise width and margins when the divider should not extend to the full width.
+
+### Don't
+
+- Don't add margins directly to the Divider — use the parent's spacing instead.
```
DmitryB-NF: ❓ Is it actually good advice? If we need a specific placement of the Divider without an `inset` attribute, adding margins directly to the `Hr` is the right approach

apps/docs/content/3-components/2-library/divider/react.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+<Section current="react">
+
+<ComponentPreview>
+  <Divider />
+</ComponentPreview>
+
+## Examples
+
+### Default
```
DmitryB-NF: Does it make sense to name it Horizontal? `Default` term is kind of odd for the documentation

packages/react/tests/visual.spec.ts-snapshots/layout-divider--inset-firefox-linux.png
DmitryB-NF, line 1: ⚠️ I suppose it should be removed after the `inset` decision. We can still add an example of different margins to have a clear example of how to make it smaller than the parent, but we may not add Playwright for it

packages/core/atoms/Divider.lite.tsx
```diff
@@ -0,0 +1,42 @@
...
+  id?: string;
+  dataTestId?: string;
+};
+
+export default function Divider(props: Props) {
+  return (
+    <hr
+      id={props.id}
+      data-testid={props.dataTestId}
+      aria-orientation={getOrientation(props.orientation) === Orientation.VERTICAL ? Orientation.VERTICAL : undefined}
```
Ademsk1, line 21: Can you expand on why `aria-orientation` is done like this? I see that this aria prop can be `horizontal`, `vertical` and is default `undefined`. But here we seem to only show `aria-orientation if its vertical?

DmitryB-NF (changes requested):
The Divider looks much cleaner and reusable now 🚀 
We need to polish the documentation, and we are good to go

packages/core/atoms/Divider.lite.tsx
ramonmulia: Setting aria-orientation="horizontal" is unnecessary as it is already the implicit default behavior.

packages/react/tests/visual.spec.ts-snapshots/layout-divider--inset-firefox-linux.png
ramonmulia: good call!

packages/react/src/footer/footer.stories.tsx
```diff
@@ -69,7 +69,7 @@ export const CompleteFooter: Story = {
         </div>
         <div className="gi-col-span-4 md:gi-col-span-4 lg:gi-col-span-3">
           <h3 className="gi-heading-sm gi-mb-4">Departments</h3>
-          <SectionBreak color="gi-border-color-border-system-neutral-subtle" size="md" />
+          <Divider className="gi-my-4" />
```
DmitryB-NF, line 72: ❓ this looks like a design change. Could you explicitly note this in the Ticket to create a new one for the Figma update?

packages/core/atoms/storybook/Divider.meta.ts
```diff
@@ -0,0 +1,95 @@
...
+          'Divider is a visual separator rendered as a native `<hr>`. It has no margin — spacing is controlled by the parent layout (gap, padding). Supports horizontal and vertical orientations.',
+      },
+    },
+  },
+};
+
+export const Horizontal = {
+  args: {
+    ...dividerMeta.args,
+  },
```
DmitryB-NF: This is excessive, as all args will be inherited from meta anyway

DmitryB-NF (dismissed):
Divider looks good!
Make sure the design chnage is reflected in the ticket, so we can merge this

DmitryB-NF (approved):

PR #955, opened 2026-05-13
chore(AB#37589): adopt `@/` path alias for internal imports

DmitryB-NF:
## Description

Wire `@/*` → `src/*` in React and Angular `tsconfig` (and Vite for React), switch internal React imports to the alias, and add `eslint-plugin-no-relative-import-paths` to prevent regressions. `vite-plugin-dts` and `ng-packagr` resolve the alias at build time, so published `.d.ts` and FESM bundles stay relative.

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#37589](https://dev.azure.com/OGCIO-Digital-Services/2aff229e-6625-4ad7-8f51-ff850dd878ec/_workitems/edit/37589)


packages/react/src/input-text/input-text.stories.tsx
```diff
@@ -1,11 +1,11 @@
...
 import { FormProvider, useForm } from 'react-hook-form';
 import { expect, userEvent, within } from 'storybook/test';
-import Button from '../atoms/Button';
-import { FormField, FormFieldError, FormFieldHint, FormFieldLabel } from '../forms/form-field/form-field.js';
-import { Icon } from '../icon/icon.js';
-import { Link } from '../link/link.js';
-import { Spinner } from '../spinner/spinner.js';
+import Button from '@/atoms/Button';
+import { FormField, FormFieldError, FormFieldHint, FormFieldLabel } from '@/forms/form-field/form-field.js';
+import { Icon } from '@/icon/icon.js';
```
Ademsk1, line 6: I notice in some of the imports in this list we append `.js`. Is this necessary to keep?

packages/react/src/modal/modal.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import { useState } from 'react';
 import { expect, within, waitFor, screen, userEvent } from 'storybook/test';
-import Button from '../atoms/Button';
-import Paragraph from '../atoms/Paragraph';
+import Button from '@/atoms/Button';
+import Paragraph from '@/atoms/Paragraph';
```
Ademsk1, line 5: Could import from just `atoms/` as the index file should contain these as exports. up to you.

packages/react/src/input-text/input-text.stories.tsx
DmitryB-NF: No, we definitely need to remove those. The current set of changes is a pure ESLint autofix without any manual changes to ensure full consistency. Fixing legacy `.js` could be a separate cleanup

Ademsk1 (commented):
LGTM. Check out the comments I left. If all good, will approve.

packages/react/src/modal/modal.stories.tsx
DmitryB-NF: It is preferable to use direct imports (` from '@/atoms/Paragraph'`) to simplify tree-shaking and optimise the build result. We can use named imports, but I think the default import is preferable for React Components in the convention of one component per file
Overall, this is an autofix without manual logic (exports are exactly the same as before). We can discuss a better approach of imports separatelly for all new components

ramonmulia (approved):

PR #957, opened 2026-05-13
chore(AB#37755): upgrade react vite

ramonmulia:
## Description

upgrade react vite

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [x] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#37755](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37755)


packages/react/vite.config.ts
```diff
@@ -6,6 +6,12 @@ import preserveDirectives from 'rollup-preserve-directives';
 import { defineConfig } from 'vite';
 import dts from 'vite-plugin-dts';
 import { libInjectCss } from 'vite-plugin-lib-inject-css';
+import packageJson from './package.json' with { type: 'json' };
+
+const externalPackages = [
+  ...Object.keys(packageJson.dependencies ?? {}),
+  ...Object.keys(packageJson.peerDependencies ?? {}),
+];
```
DmitryB-NF: I think we can replace this by `import { externalizeDeps } from 'vite-plugin-externalize-deps';` to make sure we do not make this processing manually

packages/react/vite.config.ts
```diff
@@ -35,10 +41,8 @@ export default defineConfig({
       entry: 'src/index.ts',
       formats: ['es'],
     },
-    rollupOptions: {
-      external: (id) => {
-        return id === 'react' || id === 'react-dom' || id.startsWith('react/') || id.startsWith('react-dom/');
-      },
+    rolldownOptions: {
+      external: (id) => externalPackages.some((name) => id === name || id.startsWith(`${name}/`)),
```
DmitryB-NF: I would try to use native features as much as possible like this

```
const BUILD_EXCLUDE = [
  'src/**/*.stories.tsx',
  'src/**/*.test.*',
  'src/test-utilities.ts',
  'src/atoms/storybook/**',
  'src/data-table/tanstack/tanstack-helpers.ts',
];

export default defineConfig({
  resolve: {
    alias: {
      '@': path.resolve(import.meta.dirname, 'src'),
    },
  },
  plugins: [
    react(),
    libInjectCss(),
    dts({
      include: ['src'],
      exclude: BUILD_EXCLUDE,
    }),
    preserveDirectives(),
    externalizeDeps(),
  ],
  build: {
    copyPublicDir: false,
    sourcemap: false,
    lib: {
      entry: Object.fromEntries(
        glob
          .sync('src/**/*.{ts,tsx}', {
            ignore: ['src/**/*.d.ts', ...BUILD_EXCLUDE],
          })
          .map((file) => [
            file.replace(/^src\//, '').replace(/\.tsx?$/, ''),
            path.resolve(file),
          ]),
      ),
      formats: ['es'],
    },
    rolldownOptions: {
      output: {
        assetFileNames: 'assets/[name][extname]',
      },
    },
  },
});
```
@ramonmulia Could you investigate the simplified config?

packages/react/vite.config.ts
ramonmulia: oh interesting!

packages/react/vite.config.ts
```diff
@@ -1,70 +1,64 @@
...
-      exclude: [
-        'src/**/*.stories.tsx',
-        'src/**/*.test.*',
-        'src/test-utilities.ts',
-        'src/atoms/storybook/**',
-        'src/data-table/tanstack/tanstack-helpers.ts',
-      ],
-    }),
-    preserveDirectives(),
+    ...(isLibraryBuild
```
DmitryB-NF: ⚠️ Hmm, I suppose Storybook does not work properly with a clean exclusion, right? If so, it is better to create 2 configs for local runs and for lib building to ensure the config is clean and readable without internal conditions. 
We can split it into 2 files and run Storybook with a separate `viteConfigPath` to make sure we can proprly build both outputs

packages/react/vite.config.ts
```diff
@@ -1,70 +1,64 @@
...
 import preserveDirectives from 'rollup-preserve-directives';
 import { defineConfig } from 'vite';
 import dts from 'vite-plugin-dts';
+import { externalizeDeps } from 'vite-plugin-externalize-deps';
 import { libInjectCss } from 'vite-plugin-lib-inject-css';
 
+// Storybook builds as an application, not a library. Rolldown rejects
+// function-based `external` in application mode, so library-only plugins
+// and build config must be skipped when Storybook is running.
+const isLibraryBuild = process.env.STORYBOOK !== 'true';
```
DmitryB-NF: ❓ I am not sure this env variable is documented. It is better to use an explicit split of configs for Storybook if it requires something specific

packages/react/vite.config.ts
```diff
@@ -1,70 +1,64 @@
...
+      '@': path.resolve(import.meta.dirname, 'src'),
     },
   },
   plugins: [
     react(),
-    libInjectCss(),
-    dts({
-      include: ['src'],
-      exclude: [
-        'src/**/*.stories.tsx',
```
DmitryB-NF, line 22: ❓ Btw, how has it worked before for the Storybook if there was an explicit exclusion of Stories?

ramonmulia: this error is due `externalizeDeps` that was recently added, not `dts`.
I'm going to include a separate storybook vite.config for storybook, that is the second option.

DmitryB-NF (dismissed):

DmitryB-NF (approved):

PR #958, opened 2026-05-14
feat(AB#35133): inline tailwind conditional components

Ademsk1:
## Description

Converts conditional components styles from using CSS files, to inline tailwind. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#35133](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/35133)


packages/react/src/forms/form-field/descendingErrorStyles.ts
```diff
@@ -0,0 +1,36 @@
+import { tv } from 'tailwind-variants';
+
```
Ademsk1: This is probably a sub-optimal way of specifying particular input types but not others to match what was in the CSS. I'd be very happy to hear better ideas!

packages/react/src/toast/ds-toast.tsx
```diff
@@ -73,14 +39,12 @@ function Toast({
   onClose,
   slotAction,
 }: ToastProps) {
-  const { base, heading, container, innerContainer, dismiss, baseDismissible } = toastVariants({
+  const { base, heading, container, innerContainer, dismiss } = toastVariants({
```
Ademsk1: unneeded baseDissmisable.

packages/react/src/toast/toast.tsx
```diff
@@ -60,7 +60,7 @@ export const ToastProvider = () => {
               defaultValue: `Toasts-${position.y}-${position.x}`,
             })}
             data-position={`${position.y}-${position.x}`}
```
Ademsk1, line 62: can i get rid of data-position here?

ramonmulia: check if there is any css rule for `data-position` in the .css file

packages/react/src/forms/form-field/descendingErrorStyles.ts
Ademsk1: I'm going to set back the form-field to its original implementation. This needs resolving in a better way.

packages/react/src/forms/form-field/descendingErrorStyles.ts
```diff
@@ -0,0 +1,36 @@
...
+const formFieldErrorDescendantClasses = [
```
ramonmulia: should we have a specific .css file to deal with this? not sure if the readability is good here.

packages/react/src/list-item/list-item.tsx
```diff
@@ -25,3 +25,30 @@ export function ListItem({ label, href, external, bold, slot, className, ...prop
...
+export const styles = tv({
+  base: [
+    'gi-block',
+    'gi-py-4',
+    'gi-border-color-border-system-neutral-subtle',
+    'gi-border-b-xs',
+    'gi-border-solid',
+    'focus:gi-no-underline',
+    'focus:gi-rounded-sm',
+    'focus:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
```
ramonmulia: should we use semantic tokens here instead of primitives?

packages/react/src/summary-list/summary-list.tsx
```diff
@@ -15,22 +16,37 @@ export const SummaryList = ({ children, className, withBorder, ...props }: Summa
...
           {rows ? <tbody>{rows}</tbody> : null}
         </table>
       </div>
     </SummaryListProvider>
   );
 };
 
+const styles = tv({
+  slots: {
+    container:
```
ramonmulia: the base slot already includes `gi-border gi-border-color-border-system-neutral-muted` so the `withBorder: true` variant repeats the same classes and `withBorder: false` has to override with `gi-border-none`

I think maybe the base should be border free and `withBorder: true` should add it?
this also could work with less code I guess:

```const styles = tv({
    slots: {
      container:
        'gi-rounded-md gi-overflow-hidden gi-antialiased gi-text-md gi-w-full',
      thead:
        'gi-bg-color-surface-system-neutral-layer1 gi-border-b gi-border-color-border-system-neutral-muted',
    },
    variants: {
      withBorder: {
        true: {
          container: 'gi-border gi-border-color-border-system-neutral-muted',
        },
      },
    },
  });

File: packages/react/src/toast/ds-toast.tsx [hunk 7]
```diff
@@ -73,14 +39,12 @@ function Toast({
...
     variant,
   });
 
-  const baseVariant = dismissible ? baseDismissible : base;
-
   return (
-    <div className={baseVariant()}>
+    <div className={base()}>
       {showIcon ? <Icon icon={icon({ variant })} className="gi-toast-icon" data-variant={variant} /> : null}
```
ramonmulia, line 84: should `git-toast-icon` be replaced for inline tw classes too?

packages/react/src/toast/types.ts
```diff
@@ -1,7 +1,7 @@
 import type { ComponentPropsWithoutRef } from 'react';
 
 export type ToastHorizontalPosition = 'left' | 'center' | 'right';
-export type ToastVerticalPosition = 'top' | 'center' | 'bottom';
```
ramonmulia, line 4: we don't need `center` anymore?

Ademsk1: `center` was never applicable to the stylings vertically (there was never a `y:center`). and so this is just removing something that wasn't possible.

packages/react/src/forms/form-field/descendingErrorStyles.ts
Ademsk1: i remove

packages/react/src/forms/form-field/form-field.tsx
```diff
@@ -62,7 +62,7 @@ const FormFieldBase = ({ children, className, ...props }: FormFieldBaseProps) =>
   );
 
   return (
-    <fieldset className={cn({ 'gi-error-state': !!error }, className)} {...props}>
+    <fieldset className={clsx({ 'gi-error-state': !!error }, className)} {...props}>
```
Ademsk1: leaving this error state for another PR

DmitryB-NF: Does it make sense to revert the entire file to make sure it is fully covered in one PR rather than in chunks?

packages/react/src/icon/icon.tsx
```diff
@@ -29,6 +29,7 @@ import Threads from './svgs/threads.js';
 import Tiktok from './svgs/tiktok.js';
 import X from './svgs/x.js';
 import Youtube from './svgs/youtube.js';
+import { tv } from 'tailwind-variants';
```
DmitryB-NF: It seems like the variable is not used. Does it make sense to revert all the changes in the file if there are no valuable ones?

packages/react/src/list-item/list-item.tsx
DmitryB-NF: ⚠️ @Ademsk1, the list of styles seems different from what is declared in `gi-list-item`. Is it expected?

packages/react/src/summary-list/summary-list-header.tsx
```diff
@@ -82,6 +84,10 @@ export const SummaryListHeader = ({
   );
 };
 
+const styles = tv({
```
DmitryB-NF: ⚠️ it does not seem to be used

packages/react/src/summary-list/summary-list.tsx
```diff
@@ -15,22 +16,54 @@ export const SummaryList = ({ children, className, withBorder, ...props }: Summa
...
+        container: 'gi-border gi-border-color-border-system-neutral-muted',
+      },
+      false: {
+        container: 'gi-border-none',
+      },
+    },
+  },
+});
+
+export const cellVariants = tv({
```
DmitryB-NF: I am not really sure we need this to reuse in 2 files. Can we avoid refactoring and inlining styles to keep them as clean as possible?

packages/react/src/toast/types.ts
DmitryB-NF: ⚠️ Even if styles are not applied, this can be a breaking change for clients when the previous valid value shows TS errors. We should not change types without proper validation

packages/react/src/tabs/tab-list.tsx
```diff
@@ -183,15 +183,32 @@ export const TabList = ({
...
 };
+
+const styles = tv({
+  base: [
+    'gi-flex',
+    'gi-items-center',
+    'gi-overflow-visible',
+    'gi-whitespace-nowrap',
+    'gi-relative',
+    'gi-shadow-[inset_0_-1px_0_0_var(--gieds-color-border-system-neutral-muted)',
```
DmitryB-NF: ⚠️ class is broken, no closing `]`
```suggestion
    'gi-shadow-[inset_0_-1px_0_0_var(--gieds-color-border-system-neutral-muted)]',
```

packages/react/src/toast/toast.tsx
```diff
@@ -157,3 +169,36 @@ export const Toast = ({
     </div>
   );
 };
+
+const toastVariants = tv({
```
DmitryB-NF: ⚠️ styles are changed, no `:has` processing like
```
.gi-toast:has(.gi-toast-info) {
  @apply gi-shadow-[0px_0px_25px_-5px_rgba(7,33,85,0.15)];
}
```

packages/react/src/toast/ds-toast.tsx
```diff
@@ -73,15 +39,20 @@ function Toast({
...
+    heading,
+    container,
+    innerContainer,
+    dismiss,
+    icon: iconStyle,
+  } = toastVariants({
     variant,
   });
 
-  const baseVariant = dismissible ? baseDismissible : base;
```
DmitryB-NF, line 80: ⚠️ dismissible is missing now

packages/react/src/summary-list/summary-list-header.tsx
```diff
@@ -36,14 +38,14 @@ export const SummaryListHeader = ({
 
   return (
     <tr {...props} className={cn('gi-h-14', className)}>
-      <th scope="col" className="gi-text-center gi-align-middle">
+      <th scope="col" className={cellVariants({ type: 'head', header: true, className })}>
```
DmitryB-NF: ⚠️ className is applied twice now

packages/design/tailwind/src/create-theme.ts
```diff
@@ -328,6 +328,99 @@ export function createTheme(
...
+        'toast-fadeoutup': 'toast-fadeoutup 0.3s ease-out forwards',
+
+        'toast-fadeinleft': 'toast-fadeinleft 0.3s ease-in forwards',
+        'toast-fadeoutleft': 'toast-fadeoutleft 0.3s ease-in forwards',
+
+        'toast-fadeinright': 'toast-fadeinright 0.3s ease-in forwards',
+        'toast-fadeoutright': 'toast-fadeoutright 0.3s ease-in forwards',
+
+        'toast-fadeoutdown': 'toast-fadeoutdown 0.3s ease-in forwards',
+      },
```
DmitryB-NF: ❓ we need to validate the HTML lib also to make sure there is no override of the same styles from toast and progress-bar. 
It would be better to move it into React `tailwind.config.ts` to add a custom toast/progress animation styles

DmitryB-NF: I would also suggest moving it to a separate ticket to make sure we can validate it properly in HTML as well

packages/react/src/list-item/list-item.tsx
Ademsk1: I'd like to leave this for another PR. boxShadows aren't configured properly in the `create-theme` file. When they are, I can remove all the primitive focus styles from the entire react package.

packages/react/src/toast/toast.tsx
Ademsk1: I'll keep it in for now - but we should consider removing it in the mitosis migration.

DmitryB-NF (dismissed):
@Ademsk1, considering the too large scope of PR and a lot of potential breaking changes, it is better to split this into multiple PR and tickets to isolate specific changes.
I would definitely suggest splitting
1. Animations for Toast and ProgressBar (could be 2 separate tickets to make sure we can process and validate them properly)
2. The Summary List seems overcomplicated. Let's have a dedicated PR for it
3. Forms could be a separate set of tickets
4. Leave all simple and processed components here as done and confirmed.

Could you split the PR and let me know if you need help with tickets?
Thanks!

packages/react/src/list-item/list-item.tsx
Ademsk1: @DmitryB-NF there are two `gi.list-item` classes declared. one in `header.css` and one in `list-item.css`. To avoid any issues here, I superimposed both class styles. Due to this added complexity, we can move it into another ticket.

packages/react/src/toast/ds-toast.tsx
Ademsk1: I was exploring this and came to the conclusion that `baseDismissable` is not necessary. Here's why. 

1. The `base` uses the class `gi-toast-base`, and `baseDismissable` uses `gi-toast-base-dismissable`. 
2. Searching in the codebase produces one result for `gi-toast-base-dismissable`, and is isolated to the toast. It's class is this:
```
.gi-toast-base-dismissible {
  @apply gi-toast-base;
}
```

Hence, `baseDismissable` is a redundant slot, and we can remove this.

packages/react/src/toast/types.ts
Ademsk1: reverted

DmitryB-NF (approved):

PR #963, opened 2026-05-19
fix(AB#37422): Divider docs example

ramonmulia:
## Description

Divider docs example

## Type of Issue

- [ ] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues
[AB#37422](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37422)


apps/docs/next.config.mjs
```diff
@@ -10,6 +10,7 @@ const nextConfig = {
   images: {
     unoptimized: true, // TODO: review image optimisation
   },
+  serverExternalPackages: ['react-i18next', 'i18next'],
```
ramonmulia: `react-i18next` uses `use-sync-external-store` which has dynamic `require()` calls.
Turbopack does not support this and throws a runtime error, added `react-i18next` and `i18next` to `serverExternalPackages` so Turbopack skips bundling them.

Ademsk1 (changes requested):
Getting a Not Found in the React tab for the examples... is this expected?

apps/docs/next.config.mjs
Ademsk1: Hmm, what's the expectation here now? I guess we need to check the `node_modules` are present in the deployed version too now right? I built the app then served it, and found the source code of i18 in `standalone/node_modules`. Although it could be elsewhere as well...

ramonmulia: we don't need this, going to remove..

PR #964, opened 2026-05-20
feat(AB#38086): migrate header icons to mitosis

Ademsk1:
## Description

Deprecates the use of `icon` for Header Menu Item Button/Link. Introduces use of Menu and Search into the icons.tsx. Replaces the storybook use of Icons with the SVG's directly. 

I opted to avoid any significant changes to the legacy header and its functions. Let me know if this is not right. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#38086](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38086)


packages/react/src/header/header-next/components/menu/components/header-menu-item-button.tsx
```diff
@@ -11,12 +11,10 @@ import { useHeaderMenuSection } from '@/header/header-next/components/menu/heade
...
-    <>
-      <PrimitiveButton data-item-mode={showItemMode} {...props}>
-        {children}
-        {icon && <Icon icon={icon} ariaHidden={true} />}
-      </PrimitiveButton>
-    </>
+    <PrimitiveButton data-item-mode={showItemMode} {...props}>
+      {children}
+      {icon && <Icon icon={icon} ariaHidden={true} />}
+    </PrimitiveButton>
```
Ademsk1: removed due to linting. No changes. React docs: 
<blockquote>
Grouping elements in Fragment has no effect on the resulting DOM; it is the same as if the elements were not grouped. 
</blockquote>
this doesn't need to be grouped.

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -232,14 +232,13 @@ export const Default: StoryObj = {
...
 
-    await step('FAQ drawer button updates aria + icon', async () => {
+    await step('FAQ drawer button updates aria', async () => {
       const faqButton = await canvas.findByRole('button', {
         name: /toggle frequently asked questions/i,
       });
       expect(within(faqButton).getByTestId(/info/i)).toBeInTheDocument();
-
       await userEvent.click(faqButton);
-      expect(within(faqButton).getByTestId('close')).toBeInTheDocument();
```
Ademsk1, line 242: I removed the Close icon in the FAQ drawer when the drawer is open for this example.

packages/react/src/icon/icon.tsx
```diff
@@ -93,6 +96,8 @@ const ICON_REGISTRY: Record<
   arrow_right_alt: { Component: ArrowRightIcon },
   first_page: { Component: FirstPageIcon },
   last_page: { Component: LastPageIcon },
+  menu: { Component: MenuIcon },
+  search: { Component: SearchIcon },
```
Ademsk1, line 99: Adding these here enables `HeaderLegacy` to **also** use the SVG icons rather than the material symbols. I opted to avoid significant structural changes to `HeaderLegacy` - as its mechanics of slots are quite complex. Let me know if `HeaderLegacy` needs structural changes as well.

packages/react/src/icon/icon.tsx
```diff
@@ -112,7 +117,7 @@ export const Icon = forwardRef<HTMLSpanElement, IconProps>(
         className,
       );
 
-      return <Component size={fontSize} className={svgClass} />;
+      return <Component id={props?.id} size={fontSize} className={svgClass} />;
```
Ademsk1, line 119: `HeaderLegacy` compatibility: `helper.ts` uses a `querySelector` to toggle the CSS display of this component - and require id to be passed in.

packages/react/tests/visual.spec.ts-snapshots/layout-header--default-firefox-linux.png
Ademsk1, line 1: 1 pixel difference in the info icon. This is expected - we had the same thing when migrating the SVG's for the Alert component.

packages/react/src/header/components/header-search.tsx
```diff
@@ -2,12 +2,15 @@ import Heading from '@/Heading.js';
 import Button from '@/atoms/Button';
 import { translate as t } from '@/i18n/utility.js';
 import type { IconId } from '@/icon/icon.js';
-import { IconButton } from '@/icon-button/icon-button.js';
+import { IconButton } from '@/icon-button/icon-button';
+import GiIconButton from '@/atoms/IconButton';
+import Close from '@/atoms/icons/Close';
```
DmitryB-NF: It is better to use `CloseIcon` notation in all new code to make the structure clear

packages/react/src/header/components/header-search.tsx
```diff
@@ -2,12 +2,15 @@ import Heading from '@/Heading.js';
...
 import { InputText } from '@/input-text/input-text.js';
 
 export type HeaderSearchProps = {
   action?: string;
   serverAction?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  /** @deprecated variable icon will default to "search" icon in future releases */
```
DmitryB-NF: ⚠️ It may not be expected behaviour to hardcode an icon here and still make it flexible. We can replace the string with a component slot instead if required during the Mitosis migration, e.g to use like 
```
<HeaderSearch Icon={<SearchIcon color="black">}>.
```
Let's remove this for future improvements to avoid public API changes in the internal SVG migtation

packages/react/src/header/components/header-search.tsx
```diff
@@ -42,7 +45,17 @@ export function HeaderSearch({ action, serverAction, icon = 'search' }: HeaderSe
...
-          <IconButton aria-label={t('header.search', { defaultValue: 'Search' })} aria-hidden="true" icon={{ icon }} />
+          {icon ? (
+            <IconButton
+              aria-label={t('header.search', { defaultValue: 'Search' })}
+              aria-hidden="true"
+              icon={{ icon }}
+            />
+          ) : (
+            <GiIconButton ariaLabel={t('header.search', { defaultValue: 'Search' })}>
+              <Close className="gi-shrink-0" />
```
DmitryB-NF: ⚠️ The default icon should be Search, not Close.

packages/react/src/header/header-next/components/menu/components/header-menu-item-button.tsx
DmitryB-NF: Good one, but let's try to avoid refactoring changes if there are no other required changes in the file

packages/react/src/icon/icon.tsx
DmitryB-NF: Great, this is a good decision. I think the Legacy components should be deprecated to avoid changes

packages/react/src/header/components/header-search.tsx
```diff
@@ -42,7 +45,17 @@ export function HeaderSearch({ action, serverAction, icon = 'search' }: HeaderSe
           </Button>
         </div>
         <div className="gi-ml-1 gi-flex-none gi-block md:gi-hidden">
-          <IconButton aria-label={t('header.search', { defaultValue: 'Search' })} aria-hidden="true" icon={{ icon }} />
+          {icon ? (
```
DmitryB-NF: ⚠️ This `if` is always true because the icon has the 'search' default value. We need to remove it to actually use SVG here

DmitryB-NF (dismissed):

packages/react/src/header/components/header-search.tsx
Ademsk1: Removed `icon` default parameter value "search" to remedy this.

packages/react/src/header/components/header-search.tsx
```diff
@@ -42,7 +44,17 @@ export function HeaderSearch({ action, serverAction, icon = 'search' }: HeaderSe
...
+            <IconButton
```
DmitryB-NF: @Ademsk1 could we also migrate this to GiIconButton? We can use `Icon` for named icons and it will simplify the future migtation

DmitryB-NF (dismissed):
Migration looks good, but it would be nice to fully migrate on Atoms whre possible

packages/react/src/header/components/header-search.tsx
```diff
@@ -42,7 +43,9 @@ export function HeaderSearch({ action, serverAction, icon = 'search' }: HeaderSe
           </Button>
         </div>
         <div className="gi-ml-1 gi-flex-none gi-block md:gi-hidden">
-          <IconButton aria-label={t('header.search', { defaultValue: 'Search' })} aria-hidden="true" icon={{ icon }} />
+          <IconButton ariaLabel={t('header.search', { defaultValue: 'Search' })}>
+            {icon ? <Icon icon={icon} /> : <SearchIcon className="gi-shrink-0" />}
+          </IconButton>
```
DmitryB-NF, line 48: It seems like there are no regression tests for this case. Can we add some?

Ademsk1: Added in a story describing HeaderSearch abilities - with visual regressions as well for either version. Let me know if this needs to be dialled back.

packages/react/src/icon/icon.tsx
```diff
@@ -18,6 +18,9 @@ import {
   FirstPageIcon,
   LastPageIcon,
 } from '@/atoms/icons';
+import MenuIcon from '@/atoms/icons/Menu';
+import SearchIcon from '@/atoms/icons/Search';
+import { IconProps as GiIconProps } from '@/atoms/icons';
```
DmitryB-NF: Imported but unused

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -25,6 +23,15 @@ import { HeaderMenuItemSlot } from './components/menu/components/header-menu-ite
...
 import { HeaderNext as Header, HeaderSlotContainer } from './header-next.js';
+import InfoIcon from '@/atoms/icons/Info';
+import LogoutIcon from '@/atoms/icons/Logout';
+import SearchIcon from '@/atoms/icons/Search';
+import MicrophoneIcon from '@/atoms/icons/Mic';
+import CloseIcon from '@/atoms/icons/Close';
+import MenuIcon from '@/atoms/icons/Menu';
+import Divider from '@/atoms/Divider';
+
+const SM = '16px';
```
DmitryB-NF: I think we can inline this as `16`

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +652,102 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+        '[&_[data-testid=header-search-form]>div>div:nth-last-child(2)]:md:!gi-hidden',
+        'gi-max-w-sm',
+      )}
+    >
+      {children}
+    </div>
+  );
+}
+
+const nonUniqueRoles = {
```
DmitryB-NF: Can it be inlined?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +652,102 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+  a11y: {
+    config: {
+      rules: [{ id: 'landmark-unique', enabled: false }],
+    },
+  },
+};
+
+export const UsingSearch: StoryObj = {
```
DmitryB-NF: Does it make sense to name it `HeaderSearch` and add one more story where we can customise the icon? It is better to have stories small and clean to make sure it could be also readable documentation for users

Ademsk1: Adjusted the stories now to have 2 examples, updated the description to be more informative - enabled controls as well for the icon being passed in, and disabled meta controls for header. :D

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +649,110 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+    const searchButton = await canvas.findByRole('button');
+    await userEvent.click(searchButton);
+    expect(handleOnSubmitMock).toHaveBeenCalled();
+    form.removeEventListener('submit', handleOnSubmitMock);
+    handleOnSubmitMock.mockClear();
+    await userEvent.click(form);
+  });
+};
+
+export const HeaderSearch_: StoryObj = {
```
ramonmulia, line 682: maybe a more descriptive name like `HeaderSearchDefault`, or alias the import: `import { HeaderSearch as HeaderSearchComponent }` ? I think  it would make it clearer.

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -114,8 +113,8 @@ export const Default: StoryObj = {
             </HeaderMenuItemLink>
             <HeaderMenuItemSlot className="gi-flex gi-items-center">
               <label>Hello John &nbsp;| </label>
-              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-default">
-                <Icon icon="logout" size="sm" />
+              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-light" aria-label="logout">
```
ramonmulia: is the light background on the Default story intentional?

Ademsk1: it is not, good catch!

packages/react/src/header/header-next/header-next.stories.tsx
Ademsk1: I think `HeaderSearchDefault` is slightly misleading, as I've always thought that appending `Default` suggests that maybe there's another variant that styles the entire component differently? Like Default and Light for the Header.
I also don't want to use an import name that's different from what we export from the package (that the user will use) as I want the "Show Code" section to be as close to what they will have as possible. 
I'm hoping the description for it in the docs will be clear enough.

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +649,110 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+  render: function Render() {
+    return (
+      // Pass in an action to specify a URL different to your own
+      <HeaderSearch action="/search-api" />
+    );
+  },
+  play: headerSearchTests,
+};
+
+function HeaderSearchShowIconOnMd({ children }: { children: React.ReactNode }) {
```
DmitryB-NF: Could it be inlined for one-time usage?

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +649,110 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+    },
+  },
+  args: {
+    icon: 'menu',
+  },
+
+  render: function Render({ icon }) {
+    return (
+      <div>
+        <p>Mobile view</p>
```
DmitryB-NF: ⚠️ it is better to add this into the Story description to make sure the code is a full busseness example only

DmitryB-NF (commented):
Minor changes in the Storybook and we're good to go!

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -419,8 +413,8 @@ export const Light: StoryObj = {
             </HeaderMenuItemLink>
             <HeaderMenuItemSlot className="gi-flex gi-items-center">
               <label>Hello John &nbsp;| </label>
-              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-light">
-                <Icon icon="logout" size="sm" />
+              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-light" aria-label="logout">
```
DmitryB-NF: It is better to add a readable aria label for the icon itself `<LogoutIcon label="Log out"/>`

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -655,3 +649,110 @@ export const WithTitleAsLinkFocusState: StoryObj = {
...
+};
+
+function HeaderSearchShowIconOnMd({ children }: { children: React.ReactNode }) {
+  return (
+    <div
+      className={clsx(
+        // IconButton column: force visible from md up
+        '[&_[data-testid=header-search-form]>div>div:last-child]:md:!gi-block',
+        // hide the text Button column on md+
+        '[&_[data-testid=header-search-form]>div>div:nth-last-child(2)]:md:!gi-hidden',
```
DmitryB-NF: ❓ Can we use the `viewport` parameter for Storybook instead of patching the internal implementation of the component? I think it is better to have the code as clean as possible for the demo and use all the abilities of the Storybook to change the viewport

packages/react/src/header/header-next/header-next.stories.tsx
```diff
@@ -114,8 +113,8 @@ export const Default: StoryObj = {
             </HeaderMenuItemLink>
             <HeaderMenuItemSlot className="gi-flex gi-items-center">
               <label>Hello John &nbsp;| </label>
-              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-default">
-                <Icon icon="logout" size="sm" />
+              <a href="#" className="gi-header-secondary-item gi-header-secondary-item-default" aria-label="logout">
```
DmitryB-NF: aria label is not required here as it is on the Icon

packages/react/src/header/header-next/header-next.stories.tsx
DmitryB-NF: Still not resolved?

DmitryB-NF (approved):

PR #965, opened 2026-05-21
feat(AB#38088): migrate tick icon

Ademsk1:
## Description

Migrates over the Check icon from using material symbols to the icon directly for ProgressStepper and Select

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#38088](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38088)


packages/react/src/progress-stepper/progress-stepper.tsx
```diff
@@ -49,12 +49,12 @@ const Connector = ({
 const getIndicatorClasses = (indicator: ProgressStepperIndicatorType) => {
   const indicatorClasses = {
     [ProgressStepperIndicator.Hashtag]: {
-      completed: <Icon icon="check" />,
+      completed: <Check />,
```
Ademsk1: As there is only 1 child in the Step, we don't need a shrink-0 here.

packages/react/src/progress-stepper/progress-stepper.tsx
```diff
@@ -1,8 +1,7 @@
 import type { FC } from 'react';
 import { Children } from 'react';
 import { cn } from '@/cn.js';
-import { Icon } from '@/icon/icon.js';
-import type { StepFillLevelType } from './types.js';
+import Check from '@/atoms/icons/Check';
```
DmitryB-NF: `CheckIcon` to follow the convention?

packages/react/tests/visual.spec.ts-snapshots/application-progressstepper--default-android-linux.png
DmitryB-NF, line 1: @Ademsk1 the new icon is slightly different. Could you make sure that the SVG Atom has the correct paths?
This could be totally expected to be slightly different. If paths are ok, we can move with a new icon

DmitryB-NF (commented):
The migration looks good!
Can we check the icon difference before merging?

packages/react/tests/visual.spec.ts-snapshots/application-progressstepper--default-android-linux.png
Ademsk1: Was very mystified by this. Here are the steps I took to evaluate this
1. Checked the `material-symbols` configuration was correct when I got the `path` ✅ 
2. Checked `icon.tsx` to ensure that styles like `opsz` and `GRAD` were matching as well ✅ 
3. Zoomed in and measured each icon to ensure stroke weight was correct ✅ 

the material symbols used a weight of 400 that maps to a particular strokeWidth. But the path is definitely defined for font weight 400. 🤷

Ademsk1: There are some additional styles that is applied by `material-symbols` anti-aliasing, font-features etc.

DmitryB-NF (approved):

PR #969, opened 2026-05-25
feat(AB#38397): Adding Link atom component

ramonmulia:
## Description

Adding Link atom component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#38397](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38397)


packages/react/tests/visual.spec.ts-snapshots/components-link--interaction-states-iphone-linux.png
Ademsk1, line 1: Looking at the focus state in this image and in others, it looks like the text is too close to the outline. Does this match the Figma?

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,77 @@
...
+  href: string;
+  className?: string;
+
+  external?: boolean;
+  target?: '_self' | '_blank' | '_parent' | '_top';
+  rel?: string;
+  download?: string | boolean;
+
+  ariaCurrent?: 'page' | 'step' | 'location' | 'date' | 'time' | boolean;
+  ariaLabel?: string;
```
Ademsk1, line 45: This feels at odds with what we use to prescribe other aria-labels - simply calling it `label`. I personally prefer `ariaLabel`, as `label` is too ambiguous to immediately understand its an accessibility label. @DmitryB-NF what do you think?

packages/react/tests/visual.spec.ts-snapshots/components-link--interaction-states-iphone-linux.png
Ademsk1: As discussed with @LexaGallery , all is fine here, some strange mobile view focus state spacing. Might still be worth double checking there's nothing strange going on here @ramonmulia for mobile view.

ramonmulia: good call Adam! Fixed.

packages/react/src/stories/Link.stories.tsx
```diff
@@ -0,0 +1,88 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
```
DmitryB-NF, line 1: The Storybook is great, but I do not think we can provide it separately, especially since it has no component directly exported. 
Could you check if it is possible to use this Atom directly in the existing React Link to make sure we can provide a pure Atom directly without any wrappers, or remove the Storybook from the React side to add it later? We can even move it in a separate Draft PR to reuse later

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  dataTestId?: string;
+};
+
+export default function Link(props: Props) {
+  return (
+    <a
+      id={props.id}
+      href={props.href}
+      class={styles({ class: props.className })}
```
DmitryB-NF: Does it make sense to encapsulate `underline: false` styles here as well?

DmitryB-NF (changes requested):

packages/core/atoms/Link.lite.tsx
DmitryB-NF: I think `ariaLabel` is better especially if ew have all other props named with `aria`

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+    'focus-visible:gi-no-underline',
+    'focus-visible:gi-rounded-sm',
+    'focus-visible:gi-outline-none',
+    'focus-visible:gi-text-color-text-tone-convention-hover',
+    'aria-[current=page]:gi-text-color-icon-tone-convention-disabled',
+    'aria-[current=page]:gi-pointer-events-none',
+    'aria-[current=page]:gi-no-underline',
+    'aria-[current=page]:focus-visible:gi-outline-none',
+    'aria-[current=page]:focus-visible:gi-shadow-none',
+    'supports-[-moz-appearance:none]:gi-underline-offset-[0.23em]',
```
Ademsk1: Super cool to see working through tailwind! However, we will also need these styles as well! 
```css
  :where(.gi-underline),
  a:where(.gi-underline) {
    text-decoration-thickness: max(1px, 0.08em);
    text-underline-offset: 0.2em;
  }
  /* required, as <a/> tag by default uses text-decoration: inherit */

  :where(.gi-underline:hover),
  a:where(.gi-underline:hover) {
    text-decoration-thickness: max(2px, 0.12em);
  }
```
it was a very broad-sweeping underline fix. Hopefully you can apply these here. You can most definitely not include the `:where`. That was to allow users to override the styles originating from CSS files by passing in their own classes. CSS files got rendered last so always won.

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+  ],
+  variants: {
+    underline: {
+      true: 'gi-underline',
+      false: 'gi-no-underline hover:gi-underline focus:gi-no-underline',
```
Ademsk1: Doesn't focus already have `gi-no-underline`?

packages/react/src/link/link.tsx
```diff
@@ -76,6 +77,31 @@ export const Link = forwardRef<HTMLElement, LinkProps>(
     },
     ref,
   ) => {
+    const isBaseLink =
```
Ademsk1: Minor;  would be nice to use `isLegacy` to be in line with other components we've done.

ramonmulia: I see, but in this scenario it's the opposite, the intention here is to render the Atom link (the new one), not the legacy one.

packages/core/atoms/Link.lite.tsx
ramonmulia: true! we have it in base, don't need to be added here. removed.

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,101 @@
...
+    'focus-visible:gi-text-color-text-tone-convention-hover',
+    'aria-[current=page]:gi-text-color-icon-tone-convention-disabled',
+    'aria-[current=page]:gi-pointer-events-none',
+    'aria-[current=page]:gi-no-underline',
+    'aria-[current=page]:focus-visible:gi-outline-none',
+    'aria-[current=page]:focus-visible:gi-shadow-none',
+    'gi-decoration-[max(1px,0.08em)]',
+    'gi-underline-offset-[0.2em]',
+    'hover:gi-decoration-[max(2px,0.12em)]',
+    'hover:gi-underline-offset-[0.2em]',
```
Ademsk1: I think this `hover:gi-underline-offset-[0.2em]` is redundant.

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,101 @@
...
+    'hover:gi-underline',
+    'supports-[-moz-appearance:none]:gi-underline-offset-[0.23em]',
+    'supports-[-moz-appearance:none]:hover:gi-underline-offset-[0.23em]',
```
Ademsk1: I think this is redundant too. Let me know if I'm wrong.

packages/react/src/link/link.tsx
```diff
@@ -76,6 +77,31 @@ export const Link = forwardRef<HTMLElement, LinkProps>(
...
+      !asChild && !asButton && !iconStart && !iconEnd && !disabled && !noVisited && !noColor && !appearance;
+
+    if (isBaseLink && props.href) {
+      return (
+        <GiLink
+          href={props.href}
+          external={external}
+          underline={!noUnderline}
+          className={cn(
```
Ademsk1: clsx?

packages/react/src/link/link.tsx
Ademsk1, line 75: Given we leave props to be really open ended on what the user can pass in, it is worth including in the GiLink below **all** the props that it could possible accept. e.g.

```jsx
<GiLink
  href={props.href}
  ....
  download={props.download}
  ariaDescribedBy={props.ariaDescribedBy}
  dataTestId={props['data-test-id']}
```

My justification for this is that the `isBaseLink` requirement only considers props that we destructure in the function. if a user just used 

`<Link href="/some-link" download/>
then `isBaseLink` would still be true , yet download wouldn't be propegated.

Ademsk1 (dismissed):

packages/react/src/link/link.tsx
Ademsk1: There are 3 streams here.
1. The use passes in props that all can be used by the `<GiLink/>`. Then we can use `<GiLink>` - no issues.
2. The user passes in a mixture - legacy props, and props that can be accepted by `<GiLink/>`. Then we use Legacy. no issues either,
3. The user passes in props that aren't LegacyLink specific, but also aren't available in GiLink. Yet it goes to GiLink. `<GiLink>` does not propegate to the `<a>` tag. This is a problem.

Effectively, one solution might be to reverse the logic of `isBaseLink` - and instead ask, are there any props that `<GiLink>` does not accept - rather than, are there props that LegacyLink only accepts. 
e.g.
```jsx
if (Object.keys(props).every((prop) => giLinkProps.includes(prop)) {
  return <GiLink/>
} else {
  return <LegacyLink/>
}
```

I think this is better because `GiLink` props are are a smaller subset than what is possible for our `LegacyLink` to accept.

packages/core/atoms/Link.lite.tsx
ramonmulia: true!

packages/react/src/link/link.tsx
ramonmulia: good catch! The `LinkNext` path already explicitly maps every `GiLink` prop so nothing gets lost, the `useLinkNext` guard only checks some needed legacy features, if none are truthy, it routes to `LinkNext`

packages/react/src/Link.tsx
```diff
@@ -0,0 +1,21 @@
...
+  inline?: boolean;
+  underline?: (typeof Underline)[keyof typeof Underline];
+  appearance?: (typeof Appearance)[keyof typeof Appearance];
+};
+
+const Link = ({ inline, underline, appearance, className, ...props }: LinkProps) => {
+  return (
+    <GiLink
+      className={
+        _.some({ underline, appearance, inline }) ? linkStyles({ underline, appearance, class: className }) : className
```
DmitryB-NF: 💡 `inline` prop here makes it a bit misleading because `<Link underline="full">` will have the same set of styles as for `<Link inline>`, which seems unclear. Does it make sense to change the `inline` to `variant: 'inline' | 'plan'` to make sure all variables are independent and `<Link underline="full" variant='plain'>` applies only styles related to underline but without colour?

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,117 @@
...
+    'aria-[current=page]:gi-pointer-events-none',
+    'aria-[current=page]:gi-no-underline',
+    'aria-[current=page]:focus-visible:gi-outline-none',
+    'aria-[current=page]:focus-visible:gi-shadow-none',
+    'gi-decoration-[max(1px,0.08em)]',
+    'gi-underline-offset-[0.2em]',
+    'hover:gi-decoration-[max(2px,0.12em)]',
+    'hover:gi-underline',
+    'supports-[-moz-appearance:none]:gi-underline-offset-[0.23em]',
+  ],
```
DmitryB-NF, line 100: What if we could split styles into 2 groups to make this whole base only for the `inline` or `variant` prop? It could be slightly misleading to have `base` if it's not used in the component?

DmitryB-NF: We can also move most of the styles into `appearance: Default` and have no default value not to apply those

ramonmulia: yeap, it was moved all the structural styles (colors, focus ring, underline, font) into `variant: 'inline'` and left base empty.

packages/react/src/Link.tsx
ramonmulia: good idea, the `inline` got replaced with `variant: 'inline' | 'plain'` so now all three props (variant, underline, appearance) are fully independent.. this way consumers can compose styles more freely, each prop controls one thing only, so no hidden coupling or unexpected side effects when you override a single prop.

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,133 @@
...
+export const Underline = {
+  ALWAYS: 'always',
+  HOVER: 'hover',
+  NONE: 'none',
+} as const;
+
+export const Appearance = {
+  LIGHT: 'light',
+  INHERIT: 'inherit',
+} as const;
```
DmitryB-NF: Can we move constants before the function, especially if they are public?

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,133 @@
...
+      data-testid={props.dataTestId}
+    >
+      {props.children}
+    </a>
+  );
+}
+
+export const Variant = {
+  INLINE: 'inline',
+  PLAIN: 'plain',
```
DmitryB-NF: `default`? Because `Plan` may not be clear.
We can aslo make it `inline` or `undefined` to remove `default`/`plain` options

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,133 @@
...
+      [Underline.NONE]: 'gi-no-underline hover:gi-no-underline',
+    },
+    appearance: {
+      [Appearance.LIGHT]: 'gi-text-white hover:gi-text-white focus:gi-text-white',
+      [Appearance.INHERIT]: 'gi-text-inherit hover:gi-text-inherit',
+    },
+  },
+  defaultVariants: {
+    variant: Variant.INLINE,
+  },
```
DmitryB-NF, line 158: ❓ Can we make the inline optional to use a general Link for general purposes instead of text by default?

packages/react/src/link/link.tsx
```diff
@@ -76,6 +79,50 @@ export const Link = forwardRef<HTMLElement, LinkProps>(
...
+      return (
+        <LinkNext
+          href={props.href}
+          external={external}
+          underline={noUnderline ? Underline.HOVER : Underline.ALWAYS}
+          appearance={noColor ? Appearance.INHERIT : appearance === 'light' ? Appearance.LIGHT : undefined}
+          className={clsx(
+            {
+              'visited:gi-text-color-text-tone-convention-default': noVisited,
+            },
```
DmitryB-NF: if having fully isolated styles, we can also add the `visited` parameter (the same default/none)

packages/core/atoms/storybook/Link.meta.ts
```diff
@@ -0,0 +1,251 @@
...
+    visited: 'default' as const,
+    external: false,
+    id: 'link-example',
+    dataTestId: 'link-test',
+  },
+  argTypes: {
+    children: {
+      control: 'text',
+      description:
+        'Visible link text. Should describe the destination or purpose of the link — avoid generic text like "Click here" or "Read more".',
```
DmitryB-NF: ⚠️ Generally, you can pass any HTML element here. Does it make sense to note it?

packages/core/atoms/storybook/Link.meta.ts
```diff
@@ -0,0 +1,251 @@
...
+    },
+    href: {
+      control: 'text',
+      description:
+        'The URL the link points to. An anchor without href has no implicit link role. Maps to the HTML href attribute.',
```
DmitryB-NF: ⚠️ `An anchor without href has no implicit link role` - what is this actually? I am not sure we can use our link without a href as it should be a required parameter

packages/core/atoms/storybook/Link.meta.ts
```diff
@@ -0,0 +1,251 @@
...
+import { within, expect, userEvent } from 'storybook/test';
+import { checker } from './utilities';
+
+export const linkMeta = {
+  tags: ['autodocs'] as string[],
+  title: 'Navigation/Link',
+  args: {
+    children: 'Link',
+    href: 'https://www.gov.ie',
+    variant: 'inline' as const,
```
DmitryB-NF: We may need to demo that the plain link is a default behaviour for all components

packages/core/atoms/storybook/Link.meta.ts
ramonmulia: fixed.,

packages/core/atoms/storybook/Link.meta.ts
```diff
@@ -0,0 +1,196 @@
...
+      description:
+        'Controls text decoration independently of variant. "always" shows underline, "hover" shows on hover only, "none" removes underline entirely.',
+      table: {
+        type: { summary: "'always' | 'hover' | 'none'" },
+      },
+    },
+    appearance: {
+      control: 'select',
+      options: ['light', 'inherit'],
+      description: 'Overrides the text colour. Use "light" on dark backgrounds, "inherit" to match parent text colour.',
```
DmitryB-NF: Does it make sense to notice the default behaviour as well? When undefined, it is basically the default set of styles?

packages/react/src/Link.tsx
```diff
@@ -0,0 +1,34 @@
...
+  className,
+  children,
+  href,
+  ...rest
+}: LinkProps) {
+  const styleProps = { variant, underline, appearance, visited };
+
+  if (asChild) {
+    return (
+      <Slot {...rest} className={linkStyles({ ...styleProps, class: className })}>
```
DmitryB-NF: ⚠️ The current type passes `aria-` props camelCase and those will be dropped as Radix cannot process it as it is. We may need add a manual processing the same as for `GiLink`

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,172 @@
...
+        'aria-[current=page]:gi-pointer-events-none',
+        'aria-[current=page]:gi-no-underline',
+        'aria-[current=page]:focus-visible:gi-outline-none',
+        'aria-[current=page]:focus-visible:gi-shadow-none',
+        'gi-decoration-[max(1px,0.08em)]',
+        'gi-underline-offset-[0.2em]',
+        'gi-underline',
+        'hover:gi-decoration-[max(2px,0.12em)]',
+        'hover:gi-underline',
+        'supports-[-moz-appearance:none]:gi-underline-offset-[0.23em]',
```
DmitryB-NF: ❓ I see those styles are moved here, which duplicates the ones in `gi-underline`. The initial approach to fixing all styles focused on ensuring the `gi-underline` utility in the Tailwind theme worked properly across all uses of `gi-underline`, not just for links. 
So I would suggest **removing duplicate styles** to let the `gi-underline` work correctly here, as it did initially. All internal utilities should still be stored in the Tailwind setup to work independently rather than only in the `Link` component.

packages/core/atoms/Link.lite.tsx
```diff
@@ -0,0 +1,172 @@
...
+export const Underline = {
+  ALWAYS: 'always',
+  HOVER: 'hover',
+  NONE: 'none',
+} as const;
+
+export const Appearance = {
+  LIGHT: 'light',
+  INHERIT: 'inherit',
+} as const;
```
DmitryB-NF, line 21: ⚠️ I would still suggest adding a default `here` to be consistent across all the types (e.g. `Visited`). There are 2 different consistent paradigms:

### Undefined + Type:
```
const Visited = { NONE: 'none' }
visited: Visited | undefined // explicit passing `none` or nothing
```
### Fully explicit type
```
const Visited = { DEFAULT: 'default', NONE: 'none' }
visited: Visited = Visited.Default // no value means Default
```

it is better to pick one across all types rather than using 1 for Appearance and 2 for Visited

packages/react/src/link/link.tsx
```diff
@@ -76,6 +79,46 @@ export const Link = forwardRef<HTMLElement, LinkProps>(
...
+      iconEnd,
+      disabled,
+      size,
+    });
+
+    if (useLinkNext) {
+      return (
+        <LinkNext
+          variant="inline"
+          {...(asChild ? { asChild: true } : { href: props.href ?? '' })}
```
DmitryB-NF: ⚠️ `props.href ?? ''` is slightly suspicious here because we can use a new link only when passing the actual href to avoid issues. We can leave it as it is with a comment, or just avoid using `LinkNext`

DmitryB-NF (changes requested):

packages/react/src/Link.tsx
ramonmulia: good call!

packages/core/atoms/Link.lite.tsx
ramonmulia: removed the duplicate styles, `gi-underline` now handles thickness and offset, the Firefox `@supports` fix is kept inline because the one in `typography.css` loses due to a source order issue within `@layer utilities` (both selectors use :where at specificity 0, and the base rule ends up after the `@supports` block in compiled output), I think because of that we also have this `support` in `link.css`.

packages/react/src/link/link.tsx
ramonmulia: `LinkNext` has been removed from the legacy Link.

packages/react/src/stories/Link.stories.tsx
```diff
@@ -0,0 +1,225 @@
...
+export const Default: Story = {
+  ...stories.Default,
+};
+
+export const ExternalLink: Story = {
+  ...stories.ExternalLink,
+};
+
+export const Underlines: Story = {
+  render: () => (
```
DmitryB-NF: ⚠️ we need to pass props to each render function to make sure the "Show Code" feature works as expected
Better to name `Underline` by the name of the feature

DmitryB-NF (commented):
All good, we just need to clean up Storybook a bit!

packages/react/tests/visual.spec.ts-snapshots/navigation-link-link--inline-link-webkit-linux.png
DmitryB-NF, line 1: ⚠️ @ramonmulia is it expected to have no theme focus mode? I think it should be the default styling, but still canonical focus mode

packages/angular/src/atoms/storybook/Link.meta.ts
```diff
@@ -0,0 +1,296 @@
...
+    const check = checker(args.dataTestId, canvas, step);
+    await check.is('a');
+    await check.attributes({
+      href: args.href,
+    });
+    await check.children();
+    await step('is keyboard focusable', async () => {
+      await userEvent.tab();
+      expect(canvas.getByTestId(args.dataTestId)).toHaveFocus();
+    });
```
DmitryB-NF: ⚠️ I think it is better to exclude this from the playwright because it can produce flaky tests and flickering on the UI

packages/react/tests/visual.spec.ts-snapshots/navigation-link-link--plain-link-android-linux.png
DmitryB-NF, line 1: ⚠️ I think this style is not consistent with [navigation-link-link--default-webkit-linux.png‎](https://github.com/ogcio/govie-ds/pull/969/changes#diff-2ad55fad56fd10b11f5093dd05e9deb0fcb22f08ee13d13e632af55fbfb98904) in https://github.com/ogcio/govie-ds/pull/969/changes#r3354616067

DmitryB-NF (changes requested):

packages/angular/src/atoms/storybook/Link.meta.ts
DmitryB-NF: I think it is better to have one more test for the focusable state and maybe one more story for unit tests
**Note:** I cannot see the focus state when opening the Storybook, but all screenshots havу the link with focus

packages/react/tests/visual.spec.ts-snapshots/navigation-link-link--inline-link-android-linux.png
DmitryB-NF, line 1: ⚠️ I think next-navigation does not exist anymore and should be removed

packages/angular/src/atoms/Link.ts
```diff
@@ -0,0 +1,207 @@
...
+    },
+    {
+      variant: Variant.DEFAULT,
+      visited: Visited.NONE,
+      class: 'visited:gi-text-inherit hover:visited:gi-text-inherit',
+    },
+    {
+      appearance: Appearance.LIGHT,
+      visited: Visited.DEFAULT,
+      class: 'visited:gi-text-color-text-tone-light-visited hover:visited:gi-text-color-text-tone-light-visited',
```
DmitryB-NF: `'focus-visible:gi-text-color-text-tone-convention-hover'` has been lost here

packages/angular/src/atoms/Link.ts
```diff
@@ -0,0 +1,207 @@
...
+        'hover:gi-text-color-text-tone-convention-hover',
+        'focus:gi-no-underline',
+        'focus:gi-rounded-sm',
+        'focus:gi-text-color-text-tone-convention-hover',
+        'focus:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
+        'focus-visible:gi-shadow-[0_0_0_2px_var(--gieds-color-gray-950),0_0_0_5px_var(--gieds-color-yellow-400)]',
+        'focus-visible:gi-no-underline',
+        'focus-visible:gi-rounded-sm',
+        'focus-visible:gi-outline-none',
+        'focus-visible:gi-text-color-text-tone-convention-hover',
```
DmitryB-NF: I think this should be removeв from here to avoid interaction with the Light theme.
Due to initial `.gi-link:not(.gi-link-inherit):not(.gi-link-disabled):not(.gi-link-light)` conditions, we may need a more careful split of all styles to conditions to avoid complicated conditions

packages/angular/src/atoms/Link.ts
ramonmulia: is not needed in this compound variant, for light appearance, the focus text color is already set to white in the  `Appearance.LIGHT` variant. this is only relevant for inline + default appearance, which is handled separately.

packages/react/src/stories/Link.stories.tsx
```diff
@@ -0,0 +1,217 @@
...
+export const PlainLink: Story = {
+  ...stories.PlainLink,
+  render: (_props) => (
+    <Box className="gi-flex gi-flex-col gi-gap-4 gi-items-start">
+      <Link href="#" ariaLabel="Home" className="gi-inline-flex gi-w-fit">
+        <HomeIcon />
+      </Link>
+      <Link asChild>
+        <a href="#" className="gi-font-primary">
+          via asChild
```
DmitryB-NF: Does it make sense to name it as "Router Link" or "Styleв Native link" because `asChild` is an internal concept?

packages/core/atoms/Link.lite.tsx
```diff
@@ -143,30 +155,56 @@ export const linkStyles = tv({
...
     {
-      variant: Variant.DEFAULT,
-      visited: Visited.NONE,
-      class: 'visited:gi-text-inherit hover:visited:gi-text-inherit',
+      variant: Variant.INLINE,
+      appearance: Appearance.LIGHT,
+      class: 'aria-[current=page]:gi-text-color-text-tone-light-disabled',
     },
     {
+      variant: Variant.INLINE,
```
ramonmulia: I’m not sure we should restrict this to inline only. If we have a `light` appearance with the `default` variant in a `visited` state, the text could become unreadable, but on focus, the text displays correctly.

packages/react/src/Link.tsx
```diff
@@ -43,8 +53,9 @@ export default function Link({
 
   return (
     <GiLink
-      {...styleProps}
       {...rest}
+      {...styleProps}
+      {...anchorProps}
```
ramonmulia, line 58: `{...anchorProps}` I don’t think this is necessary, since these rules are already handled at the atom level.

packages/core/atoms/Link.lite.tsx
```diff
@@ -92,31 +92,32 @@ export default function Link(props: Props) {
...
         'aria-[current=page]:gi-no-underline',
         'aria-[current=page]:focus-visible:gi-outline-none',
         'aria-[current=page]:focus-visible:gi-shadow-none',
-        'gi-underline',
-        'hover:gi-underline',
+        // Firefox & Safari clip descenders at 0.2em offset — bump to 0.23em.
+        // Kept as explicit utilities (not only the :where rule in typography.css)
+        // so the offset survives even when other classes touch underline styles.
         'supports-[-moz-appearance:none]:gi-underline-offset-[0.23em]',
+        'supports-[background:-webkit-named-image(i)]:gi-underline-offset-[0.23em]',
```
ramonmulia, line 120: should we also have this in typography.css?

packages/core/atoms/Link.lite.tsx
DmitryB-NF: Yes, if passing a text within a non-inlined link with a light appearance, it will have an `inherit` colour. This is a kind of edge case where we may not need to apply colours. We need to reconsider the default `'gi-text-white'` for the light appearance because, for non-inlined links, we may not add context colours

packages/react/src/Link.tsx
DmitryB-NF: This is done for readability and consistency. Otherwise, we need to see the same logic in 2 different components and it is not clear whu we need this for Slot but not for GiLink

packages/core/atoms/Link.lite.tsx
DmitryB-NF: Yes, we should, but not within the Link Atom PR

packages/core/atoms/Link.lite.tsx
DmitryB-NF: @ramonmulia I updated the Inherit styles to be consistent across Light appearence

DmitryB-NF (approved):

PR #970, opened 2026-05-26
feat(AB#37991): Adding actions to SideNavItem

ramonmulia:
## Description

- Adding actions to SideNavItem
- SideNavItem height fixes
- Storybook example added 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#37991](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/37991)


packages/react/src/side-nav/side-nav.tsx
```diff
@@ -110,51 +110,48 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
+              onClick={handleClick}
+            >
+              {asChild ? (
+                children
+              ) : (
+                <ItemContent icon={icon} label={label} showExpandableIcon={showExpandableIcon} isOpen={isOpen} />
+              )}
+            </Link>
+          ) : (
+            <Button
```
DmitryB-NF, line 126: ⚠️ I suppose having the button with actions aside without styles override will have invalid focus styles. it would be nice to have the entire nav item focusable rather than multiple items - when clicking the button, the only button will be selected instead of the entire panel

ramonmulia: This is solved by the padding added to the button/link, it reserves space for trailing actions, so focus and hover styles cover the full row.. same pattern MUI uses.

packages/react/src/side-nav/side-nav.stories.tsx
```diff
@@ -1,5 +1,10 @@
 import type { Meta } from '@storybook/react-vite';
+import { Icon } from '@/icon/icon.js';
+import { IconButton } from '@/icon-button/icon-button.js';
+import { Paragraph } from '@/paragraph/paragraph.js';
+import { Tag } from '@/tag/tag.js';
 import { SideNav, SideNavHeading, SideNavItem } from './side-nav.js';
+import { Box } from '@/Box.js';
```
DmitryB-NF: It is better to avoid legacy `js` extensions in all new imports

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--basic-android-linux.png
DmitryB-NF, line 1: It looks like the positioning has been slightly changed. Is it expected or we need to adjust styles?

packages/react/src/side-nav/side-nav.stories.tsx
```diff
@@ -53,10 +58,85 @@ export const ParentChild = {
 
 export const FullExample = {
```
DmitryB-NF: Does it make sense to add one more example for actions only? It would be much more readable if there were a simple code snippet to demonstrate how users can use additional actions.
It would be nice to have a similar one for `label` also, to demo how we can add a node with tags.

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-webkit-linux.png
DmitryB-NF, line 1: ⚠️ The text weight is changed. This could not be expected. Could you check that it matches the design?

DmitryB-NF (changes requested):
The change is really nice but у need to make sure the design is the same

packages/react/src/side-nav/side-nav.stories.tsx
```diff
@@ -1,5 +1,10 @@
 import type { Meta } from '@storybook/react-vite';
+import { Icon } from '@/icon/icon.js';
```
Ademsk1: Lets use the icons directly from `atoms/icons` :D

ramonmulia: good call!

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-webkit-linux.png
ramonmulia: yes this is expected, this is due the `primary` prop I added in the `WithIcons` story, I think this was not added by mistake, now it matches with Figma example.

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--basic-android-linux.png
ramonmulia: I believe this is due the SideNavItem height change, it was wrong. I checked with @LexaGallery and `primary` and `secondary` items should be 48px (btn lg) .

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-webkit-linux.png
DmitryB-NF: Hmm, so, `primary` is not the default, and it can be possible to use it in different modes. Do we need to fix it later or add an additional example without the `primary`?

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-actions-android-linux.png
DmitryB-NF, line 1: ❓ It looks like the actions menu is after the dropdown icon. Does it make sense to change the order, if possible, to make it more consistent?
cc: @LexaGallery

DmitryB-NF (dismissed):
The implementation looks good but it would be nice to confirm the design

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-webkit-linux.png
ramonmulia: I don't think we need to fix this, but to keep consistent with what we already tested I'm going to revert this `WithIcon` change.

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-actions-android-linux.png
ramonmulia: good call! the order was changed.

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -110,51 +110,69 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
+              className={itemClassName}
+              onClick={handleClick}
+            >
+              {asChild ? (
+                children
+              ) : (
+                <ItemContent
+                  icon={icon}
+                  label={label}
+                  showExpandableIcon={showExpandableIcon && !hasActionsWithExpandable}
```
DmitryB-NF: Having the dropdown migrate, can we just remove this one entirely?

DmitryB-NF (commented):
The change looks good!
Can we maybe clean up icons if possible, while we have already migrated to a proper structure?

packages/react/src/side-nav/side-nav.tsx
ramonmulia: yes, I kept it because it was there, but I removed cause does not make sense to have link as dropdown.

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-android-linux.png
DmitryB-NF, line 1: @ramonmulia I think overall paddings are changed in a new implementation

DmitryB-NF: I think the text align is changed when using SVG directly here which could be expected but better to check

DmitryB-NF (dismissed):
The Implementation is much more flexible and readable now!
@ramonmulia, could you confirm the reason for the differences in the screenshots, so we are good to merge?

packages/react/tests/visual.spec.ts-snapshots/navigation-sidenav--with-icons-android-linux.png
ramonmulia: good catch! I'm checking this

ramonmulia: fixed.

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -109,52 +89,53 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
+            >
+              <ItemContent icon={icon} label={label} />
+            </Button>
+          )}
+          {actions && (
+            <div className={cn('gi-absolute', hasActionsWithExpandable ? 'gi-right-12' : 'gi-right-3')}>{actions}</div>
+          )}
+          {showExpandableIcon && (
+            <div className="gi-absolute gi-right-4 gi-pointer-events-none">
+              <KeyboardArrowDown className={cn(isOpen && 'gi-rotate-180')} />
```
Ademsk1: I'm not sure if its valid here, but more a general FYI : be aware that if the parent is a flex item and can be squeezed by sibling elements, the KeyboardArrowDown will also be squeezed. This is possible especially if the sibling element is controlled by the consumer (I had this when migrating the tick icon in `<Select>`. The SelectItem text was defined by the consumer, if they wrote text that was too long, it would squeeze the icon down. Using `gi-shrink-0` fixes it).

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -109,52 +89,53 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
+              size="lg"
+              onClick={handleButtonClick}
+              className={itemClassName}
+              id={itemId}
+            >
+              <ItemContent icon={icon} label={label} />
+            </Button>
+          )}
+          {actions && (
+            <div className={cn('gi-absolute', hasActionsWithExpandable ? 'gi-right-12' : 'gi-right-3')}>{actions}</div>
```
Ademsk1: clsx?

packages/react/src/side-nav/side-nav.tsx
ramonmulia: that's is a valid concern! but it does not apply here, since it's `absolutely positioned`, it's removed from the flex flow entirely,  it can't be squeezed by flex siblings regardless of how long the label text is.
The `gi-shrink-0` fix is needed when an icon is a flex child but here the chevron is overlaid via absolute positioning, so flex shrinking doesn't affect it.

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -109,55 +89,58 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
+              <ItemContent icon={icon} label={label} />
+            </Button>
+          )}
+          {actions && (
+            <div className={clsx('gi-absolute', hasActionsWithExpandable ? 'gi-right-12' : 'gi-right-3')}>
+              {actions}
+            </div>
+          )}
+          {showExpandableIcon && (
+            <div className="gi-absolute gi-right-4 gi-pointer-events-none">
```
DmitryB-NF: I think, it should be `right-3` to be the same as it was before the change

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -109,55 +89,58 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
-
-    const buttonClassName = cn('gi-side-nav-item', {
+    const hasActionsWithExpandable = !!actions && showExpandableIcon;
+    const itemClassName = clsx('gi-w-full !gi-h-12 gi-border-transparent !gi-px-3 !gi-py-2 ', {
       'gi-side-nav-item-selected': isSelected,
       'gi-side-nav-item-primary': primary,
-      'gi-side-nav-item-secondary': secondary,
+      '!gi-px-6': secondary,
+      '!gi-pr-12': !!actions !== !!showExpandableIcon,
+      '!gi-pr-[84px]': hasActionsWithExpandable,
```
DmitryB-NF, line 102: Can we add a comment for this specific pixel padding? It could be a correct decision but it is better to explain math behind this

packages/react/src/side-nav/types.ts
```diff
@@ -1,16 +1,18 @@
...
-  label: string;
+  label: ReactNode;
   href?: string;
   asChild?: boolean;
   primary?: boolean;
   secondary?: boolean;
   open?: boolean;
   expandable?: boolean;
-  icon?: IconId;
+  icon?: IconId | Exclude<ReactNode, string>;
```
DmitryB-NF, line 14: Just in case, could you validate that the usage of plain strings still works to avoid breaking changes?

packages/react/src/side-nav/side-nav.tsx
```diff
@@ -109,55 +89,58 @@ export const SideNavItem: React.FC<PropsWithChildren<SideNavItemProps> & { open?
...
-            appearance="dark"
-            size="md"
-            onClick={handleButtonClick}
-            className={buttonClassName}
-            id={itemId}
-          >
-            <ItemContent icon={icon} label={label} showExpandableIcon={showExpandableIcon} isOpen={isOpen} />
-          </Button>
-        )}
+      <div aria-label={`${typeof label === 'string' ? label : ''} ${primary && expandable ? 'dropdown' : 'item'}`}>
```
DmitryB-NF: `role="group"` may still be required to make the `aria-label` applicable. Can we still have it to avoid any regression before the migration?

ramonmulia: good call! adding it back.

packages/react/src/side-nav/types.ts
ramonmulia: yap, confirmed working, the plain string icon values render correctly, and intellisense still restricts suggestions to the IconId union, so there's no breaking change for existing consumers.

packages/react/src/side-nav/side-nav.tsx
ramonmulia: true, reverted to `right-3` to match the original spacing.

DmitryB-NF (approved):

PR #972, opened 2026-05-28
feat(AB#38089): migrate table icons

Ademsk1:
## Description

Migrating over Table icons that use material symbols to directly using SVG. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#38089](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38089)


packages/core/atoms/constants.ts
```diff
@@ -11,6 +11,13 @@ export const Size = {
   XL: 'xl',
 } as const;
 
+export const ICON_SIZE: Record<keyof typeof Size, string> = {
```
Ademsk1: This will be useful as we migrate more icons with differing sizes to using SVG.

packages/react/src/table/table-pagination.tsx
```diff
@@ -91,23 +91,19 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
         disabled={currentPage === totalPages}
         appearance="dark"
         variant="flat"
-        className="gi-ml-2"
-        icon={{
-          icon: 'chevron_right',
-          useFontIcon: true,
-        }}
-      />
+        className="gi-ml-2 gi-p-2"
```
Ademsk1, line 98: Here we add a padding of 8px. It actually used to be 12px in the earlier version due to how `<IconButton>` handled icons. But I checked the designs and 8px is correct.

packages/core/atoms/constants.ts
ramonmulia: cool! since the default size for all icons is a number, does it make sense to keep the sizes as numbers as well?

packages/react/src/table/table-pagination.tsx
```diff
@@ -88,26 +92,25 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
...
         appearance="dark"
         variant="flat"
-        className="gi-ml-2"
-        icon={{
-          icon: 'last_page',
-          useFontIcon: true,
-        }}
-      />
+        className="gi-ml-2 gi-p-2"
+        ariaLabel="Last page"
```
Ademsk1, line 109: Please note, I've added in an aria-label here. Before - when using material symbols the text "last_page" was rendered in the span and therefore captured by the screen reader. 

However, I recognise a better way to implement aria-labels may be with i18 here. Would we like this?

packages/react/tests/visual.spec.ts-snapshots/components-table--table-with-footer-android-linux.png
Ademsk1, line 1: Despite running the tests for **all** devices, android was the only one that was failing. For some reason it renders the icon slightly taller when in span than in SVG. My eyes could be playing tricks on me, but it seems like this new delete icon from SVG is visually closer to the other device renders. 
If we'd like this investigated please say so. This is the first instance I've seen where converting from material-symbols to SVG only changed some device views but not others.

packages/core/atoms/constants.ts
Ademsk1: done!

packages/react/src/table/table-header.tsx
```diff
@@ -2,8 +2,11 @@
 import type { ThHTMLAttributes } from 'react';
 import { Children } from 'react';
 import { cn } from '@/cn.js';
-import { Icon } from '@/icon/icon.js';
 import type { TableAlign, VerticalAlign } from './table.js';
+import SwapVertical from '@/atoms/icons/SwapVertical';
+import { ICON_SIZE as size } from '@/atoms/constants';
```
ramonmulia: i think you can use `ICON_SIZE` directly (no alias), or at minimum se a more descriptive alias like `iconSize`..

packages/react/src/table/table-header.tsx
```diff
@@ -21,11 +24,14 @@ const getSortedIcon = (isChildrenString: boolean, sorted: SortedType) => {
   if (!isChildrenString) {
     return null;
   }
-
-  if (!sorted) {
-    return <Icon inline icon="swap_vert" size="sm" />;
+  switch (sorted) {
```
ramonmulia, line 26: hmm the `switch` replace is good, but the `default` case handles `'desc'`, maybe a `case 'desc':` would be more explicit and self-documenting.

packages/react/src/table/table-pagination.tsx
```diff
@@ -88,26 +92,25 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
       </div>
       <IconButton
         onClick={handleNext}
-        disabled={currentPage === totalPages}
+        disabled={atLastPage}
+        aria-disabled={atLastPage}
```
ramonmulia: `aria-disabled={atLastPage}` is set only here, but not on the other three pagination buttons, and I think since `IconButton` renders a native `<button disabled>`, the browser already exposes `ariadisabled="true"` making `aria-disabled` redundant, would be nice to check this..

packages/react/src/table/table.stories.tsx
```diff
@@ -585,7 +586,9 @@ export const WithSmallRowSize: Story = {
               <Tag text={row.status.text} type={row.status.type} />
             </TableData>
             <TableData>
-              <IconButton {...iconProps} size="md" />
+              <IconButton size="md" variant="flat" appearance="dark" ariaLabel="Download">
+                <DownloadIcon />
+              </IconButton>{' '}
```
ramonmulia: unintentional empty space?f

packages/react/src/table/table-header.tsx
Ademsk1: I agree on the self-documenting side. I wanted to match exactly the logical structure we had previously. In it, if a user passes anything **but** `false` or `asc` then `desc` is always the default.

packages/react/src/table/table-pagination.tsx
Ademsk1: Turns out we don't have aria-disabled available for IconButton - but screen readers already declare that a button is unavailable if the button is disabled anyways.

packages/core/atoms/constants.ts
```diff
@@ -11,6 +11,13 @@ export const Size = {
   XL: 'xl',
 } as const;
 
+export const ICON_SIZE: Record<keyof typeof Size, number> = {
```
DmitryB-NF: ⚠️ I think we got rid of this approach during the migration. Let's use this locally where it is required rather than adding global constans

packages/react/src/table/table-data.tsx
```diff
@@ -3,8 +3,10 @@
 import type { TdHTMLAttributes, ReactNode } from 'react';
 import { useRef, useEffect, useState } from 'react';
 import { cn } from '@/cn.js';
-import { IconButton } from '@/icon-button/icon-button.js';
+import IconButton from '@/atoms/IconButton';
 import type { TableAlign, VerticalAlign } from './table.js';
+import KeyboardArrowUp from '@/atoms/icons/KeyboardArrowUp.js';
+import KeyboardArrowDown from '@/atoms/icons/KeyboardArrowDown.js';
```
DmitryB-NF: ⚠️ No legacy JS extensions in all new code.
Can we use `XxxIcon` notation to indicate the Icon context?

packages/react/src/table/table-header.tsx
```diff
@@ -21,11 +24,16 @@ const getSortedIcon = (isChildrenString: boolean, sorted: SortedType) => {
...
+    case false:
+      return <SwapVertical size={ICON_SIZE.SM} className="gi-shrink-0" />;
+    case 'asc':
+      return <ArrowUpward size={ICON_SIZE.SM} className="gi-shrink-0" />;
+    case 'desc':
+      return <ArrowDownward size={ICON_SIZE.SM} className="gi-shrink-0" />;
```
DmitryB-NF: This case is not required as it will be processed by default

packages/react/src/table/table-header.tsx
```diff
@@ -21,11 +24,16 @@ const getSortedIcon = (isChildrenString: boolean, sorted: SortedType) => {
   if (!isChildrenString) {
     return null;
   }
-
-  if (!sorted) {
-    return <Icon inline icon="swap_vert" size="sm" />;
+  switch (sorted) {
+    case false:
+      return <SwapVertical size={ICON_SIZE.SM} className="gi-shrink-0" />;
```
DmitryB-NF: It seems like only `ICON_SIZE.SM` is used here. We can directly replace it with `16` locally rather than creating a map

packages/react/src/table/table-pagination.tsx
```diff
@@ -1,7 +1,11 @@
 import { cn } from '@/cn.js';
 import { translate as t } from '@/i18n/utility.js';
-import { IconButton } from '@/icon-button/icon-button.js';
 import { SelectItem, SelectNative } from '@/select/select-native.js';
+import IconButton from '@/atoms/IconButton';
+import LastPage from '@/atoms/icons/LastPage';
```
DmitryB-NF: `LastPageIcon`?

packages/react/src/table/table-pagination.tsx
```diff
@@ -27,6 +31,8 @@ export const TablePagination: React.FC<TablePaginationProps> = ({
       onPageChange(currentPage + 1);
     }
   };
+  const atFirstPage = currentPage === 1;
+  const atLastPage = currentPage === totalPages;
```
DmitryB-NF, line 35: Variables seem redundant. The existing seem clear

packages/react/src/table/table-pagination.tsx
DmitryB-NF: We do need to introduce i18n, but this should be a separate task

packages/react/src/table/table.stories.tsx
```diff
@@ -15,6 +15,9 @@ import { TableHead } from './table-head.js';
 import { TableHeader } from './table-header.js';
 import { TableRow } from './table-row.js';
 import { Table } from './table.js';
+import DownloadIcon from '@/atoms/icons/Download';
+import Edit from '@/atoms/icons/Edit';
+import Delete from '@/atoms/icons/Delete';
```
DmitryB-NF: ⚠️ `EditIcon`?

packages/react/tests/visual.spec.ts-snapshots/components-table--table-with-footer-android-linux.png
DmitryB-NF: If the icon is correct, we can rely on SVG as a source of truth

DmitryB-NF (changes requested):

packages/react/src/table/table-header.tsx
Ademsk1: The intention here was for the code to be more self-documenting - but happy to remove.

Ademsk1 (commented):
> Variables seem redundant. The existing seem clear 

I disagree, I think they add clarity to the intentions of why the buttons  may be disabled, and the statement only needs to be evaluated once rather than twice.

packages/react/src/table/table-pagination.tsx
Ademsk1: I disagree, I think they add clarity to the intentions of why the buttons may be disabled, and the statement only needs to be evaluated once rather than twice.

DmitryB-NF: Overall, for developers, `currentPage === totalPages` vs `atLastPage` will have exactly the same meaning, where the named constant is redundant. But it is ok still to have it even it is verbose

DmitryB-NF (approved):
Nice done! 
it is good to see new Atom components in action 🚀

PR #974, opened 2026-06-01
feat(AB#36276): animate chevrons for displaying content

Ademsk1:
## Description

Rotation styles and animation applied to several chevrons contained within well-defined components for a more visually appealing UX. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#36276](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/36276)


packages/react/src/header/components/header-menu.tsx
```diff
@@ -47,7 +48,11 @@ export const MenuItemAccordion = ({ index, item }: MenuItemAccordionProps) => {
       >
         <div>
           <span className="gi-text-sm gi-font-bold gi-ml-1">{item.label}</span>
-          <KeyboardArrowRightIcon className="gi-block gi-shrink-0 gi-accordion-item-icon" />
```
Ademsk1, line 50: removed gi-block as it is default display: block

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -350,11 +351,14 @@ const autocompleteStyles = tv({
         iconEnd: 'gi-cursor-not-allowed gi-pointer-events-none',
       },
     },
+    isOpen: {
+      true: {
+        iconEnd: 'gi-rotate-180',
```
Ademsk1, line 364: opted for using slot variants over clsx as there are other styles applied to iconEnd in line 351.

packages/react/tests/visual.spec.ts-snapshots/form-autocomplete--default-android-linux.png
Ademsk1, line 1: For all these screenshots, there is a hair of a difference between them. I closely examined this difference and investigated all possible effects. However, after some digging, it seems that this slight change is part of how the browser renders rotated vectors

packages/react/src/select/select-next.stories.tsx
```diff
@@ -107,16 +107,18 @@ export const Default: StoryObj = {
...
+      expect(keyboardArrowDown).toBeInTheDocument();
     });
 
     await userEvent.click(input);
     await waitFor(() => {
       expect(canvas.getByRole('listbox')).toBeInTheDocument();
     });
     await waitFor(() => {
-      expect(canvas.getByTestId('keyboard-arrow-up')).toBeInTheDocument();
+      expect((keyboardArrowDown.parentElement as HTMLElement).classList.contains('gi-rotate-180')).toBe(true);
```
DmitryB-NF: Better to avoid checking precise classes in tests and replace it with something stable or remove

packages/react/src/combo-box/dropdown-item.tsx
```diff
@@ -84,7 +84,10 @@ export const DropdownItem = ({
           {selected?.length > 0 && <Tag type={TagTypeEnum.Counter} text={selected.length.toString()} />}
         </div>
 
-        <Icon className={cn({ 'gi-rotate-180': isOpen })} icon="keyboard_arrow_down" />
+        <Icon
+          className={cn('motion-safe:gi-transition-transform motion-safe:gi-duration-100', { 'gi-rotate-180': isOpen })}
+          icon="keyboard_arrow_down"
+        />
```
DmitryB-NF: Replace with `KeyboardArrowDownIcon` consistently?

packages/react/src/accordion/accordion-item.tsx
```diff
@@ -90,7 +85,18 @@ const accordionVariants = tv({
...
+      true: {
+        icon: 'gi-rotate-180',
       },
+      false: '',
     },
   },
+  compoundVariants: [
+    { isExpanded: true, variant: 'default', class: { icon: 'gi-pb-[1.5px]' } },
+    { isExpanded: false, variant: 'default', class: { icon: 'gi-pt-[1.5px]' } },
+  ],
```
DmitryB-NF: I think we can simplify this by applying it with a margin
```
variant: {
    default: { header: 'gi-px-2 gi-py-4 gi-text-md gi-font-bold', icon: 'gi-mt-[1.5px]' },
    small: { header: 'gi-py-2 gi-px-2 gi-text-sm gi-font-bold' },
 },
...
isExpanded: {
  true: { icon: 'gi-rotate-180' },
},
...

Comment (Ademsk1) [reply packages/react/src/accordion/accordion-item.tsx]: This doesn't produce the same desired effect, as rotating the item also means rotating the margin / padding. what do you think about this instead:
```jsx
        <div data-testid="accordion-header" className={header()}>
          {label}
          <div className={clsx({ 'gi-pt-[1.5px]': variant === 'default' })}>
            <KeyboardArrowDownIcon className={icon()} />
          </div>
        </div>
```
now the icon is wrapped in a div that always has a padding of 1.5px for default variant. Shorter and simpler no?

DmitryB-NF: This is not actually true. Rotating affects the **padding** as an internal structure, but leaves the margin as it is, so changing _padding to marging_ avoid the need to change it manually. We can still go with the additional `div`, but I would suggest creating an example with margin instead to see how it goes

packages/react/tests/visual.spec.ts-snapshots/layout-accordion--default-android-linux.png
Ademsk1, line 1: The original accordion keyboard down icons had a top padding of 1.5px. This actually squeezed the path size - which does not match the Figma. Now we have a div wrapped with 1.5px and the icons are 24x24.

DmitryB-NF (approved):

PR #975, opened 2026-06-03
feat(AB#38090): render all icons as mitosis svg for Icon

Ademsk1:
## Description

Adds in all icons used within mitosis SVG to the `<Icon>` components ICON_REGISTRY. Updates screenshots and removes `filled` as a prop now that it is ineffective. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#38090](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38090)


packages/react/src/icon/icon.tsx
```diff
@@ -60,75 +110,138 @@ const SIZE_MAP: Record<IconSize, string> = {
...
+  attach_file: { Component: AttachFileIcon },
+  block: { Component: BlockIcon },
+  call: { Component: CallIcon },
+  cancel: { Component: CancelIcon },
+  candlestick_chart: { Component: CandlestickChartIcon },
+  chat_bubble: { Component: ChatBubbleIcon },
+  check: { Component: CheckIcon },
+  check_circle: { Component: CheckCircleIcon },
+  chevron_left: { Component: KeyboardArrowLeftIcon },
+  chevron_right: { Component: KeyboardArrowRightIcon },
```
Ademsk1: chevron left / right -> keyboard arrow left/right

packages/react/src/icon/icon.tsx
```diff
@@ -60,75 +110,138 @@ const SIZE_MAP: Record<IconSize, string> = {
...
+  placeholder: { Component: PlaceholderIcon },
   first_page: { Component: FirstPageIcon },
   last_page: { Component: LastPageIcon },
-  menu: { Component: MenuIcon },
-  search: { Component: SearchIcon },
 };
 
 export const Icon = forwardRef<HTMLSpanElement, IconProps>(
   (
-    { icon, size = 'md', filled, disabled, ariaHidden, ariaLabel, inline, className, onClick, useFontIcon, ...props },
```
Ademsk1, line 104: note: `filled` has been removed. This is intentional.

packages/react/src/icon/icon.tsx
```diff
@@ -60,75 +110,138 @@ const SIZE_MAP: Record<IconSize, string> = {
...
+      size = 'md',
+      disabled,
+      id,
+      ariaHidden,
+      ariaLabel,
+      inline,
+      className,
+      onClick,
+      useFontIcon,
+      dataTestId,
```
Ademsk1, line 211: added `dataTestId` here to pass through to the SVG component.

packages/react/src/icon/icons.ts
```diff
@@ -80,4 +80,7 @@ export const iconIds = [
   'last_page',
 ] as const;
 
+/**
+ * @deprecated Icon list will be removed in the next major version update. Use type IconId to identify the icons available for `<Icon/>`
+ */
 export const Icons = [...iconIds];
```
Ademsk1: we export this directly in the `react` entry file, otherwise I would have liked to have removed it in this PR.

packages/react/tests/visual.spec.ts-snapshots/components-icon--extra-large-android-linux.png
Ademsk1, line 1: Optical size was directly linked to `fontSize` used. It is therefore predicted that the icon would render differently for large/small fonts outside of medium.

packages/react/tests/visual.spec.ts-snapshots/form-inputfile--with-list-of-uploaded-files-android-linux.png
Ademsk1, line 1: Identified this regression with the trash icon - seen before in previous android runs. 

Android has the biggest visual differences between the icons rendered and the material symbols. Whereas iphone/webkit/chromium/firefox were very minimal. Android had differences with nearly all the icons in local testing.

packages/react/src/icon/icon.tsx
```diff
@@ -60,75 +110,138 @@ const SIZE_MAP: Record<IconSize, string> = {
...
-      const svgClass = cn(
+      const svgClass = clsx(
         { 'gi-block': !inline, 'gi-inline-block': inline },
         'gi-shrink-0',
         disabled && (disabledClass || 'gi-fill-gray-700'),
         className,
       );
 
-      return <Component id={props?.id} size={fontSize} className={svgClass} />;
+      return <Component id={id} size={fontSize} className={svgClass} label={ariaLabel} dataTestId={dataTestId} />;
```
ramonmulia, line 226: i noticed the `ariaHidden` is destructed but never passed to the svg component, and the Component has `aria-hidden={!props.label}` so when `ariaLabel` is underfined the icon is hidden, but when `ariaHidden=true` is set with an `ariaLabel`, the icon will be announced by screen readers if `ariaLabel`is also set.

 I'm not sure if we should fix this updating the icons in core or clear `ariaLabel` when `aria-hidden` is also set something like: `label={ariaHidden ? undefined : ariaLabel}`

packages/react/src/error-text/error-text.stories.tsx
```diff
@@ -107,7 +107,7 @@ export const WithRichText: Story = {
     text: (
       <div className="gi-flex">
         Error message &nbsp;
-        <Icon icon="error" />
+        <ErrorIcon label="error" />
```
DmitryB-NF: As we have a message here, `label` is not really required for the icon

packages/react/src/hint-text/hint-text.stories.tsx
```diff
@@ -91,7 +91,7 @@ export const WithRichText: Story = {
     children: (
       <div className="gi-flex">
         Here is a rich hint &nbsp;<Link href="#">Click here </Link>
-        <Icon icon="arrow_drop_up" />
+        <ArrowDropUpIcon label="arrow drop up" />
```
DmitryB-NF: It is better to remove the `label` as the icon should be hidden after the message

packages/react/src/index.ts
```diff
@@ -64,7 +64,7 @@ export { default as Heading, type Props as HeadingProps } from './Heading.js';
 export type { HintTextProps } from './hint-text/types.js';
 export * from './i18n/config.js';
 export { IconButton, type IconButtonProps } from './icon-button/icon-button.js';
-export { Icon, type IconProps } from './icon/icon.js';
+export { Icon, type IconProps, type IconId } from './icon/icon.js';
```
DmitryB-NF: `IconId` is excessive as it can be used as `IconProps['icon']`. It is better to avoid public types that will be deprecated

packages/react/src/modal/modal.tsx
```diff
@@ -43,21 +43,22 @@ const ModalCloseButton = ({ label, size = 'sm', ...props }: ModalCloseButtonProp
     >
       <>
         {label}
-        <Icon icon="close" size={iconSize} />
+        <CloseIcon size={iconSize} label="Close" />
```
DmitryB-NF: ⚠️ The label should be removed as the button itself already has the `label`. Only actionable/explicitly visible elements should have `ariaLabel`.

packages/react/src/modal/modal.tsx
```diff
@@ -43,21 +43,22 @@ const ModalCloseButton = ({ label, size = 'sm', ...props }: ModalCloseButtonProp
...
       onClick={props.onClick}
       variant="flat"
       size={normalizedSize}
       appearance="dark"
-      dataTestid="modal-close-button"
+      dataTestId="modal-close-button"
       {...props}
-    />
+    >
+      <CloseIcon size={iconSize} label="Close" />
```
DmitryB-NF: ⚠️ the `label` is excessive

packages/react/src/icon/icon.tsx
```diff
@@ -150,3 +263,86 @@ export const Icon = forwardRef<HTMLSpanElement, IconProps>(
 );
 
 export default Icon;
+
+export type IconId =
```
DmitryB-NF: ⚠️ I do not think we need one more "source of truth" here for icon IDs. We can use `keyof typeof ICON_REGISTRY`, so the type always consistently reflects actual icons

packages/react/src/icon/icon.tsx
```diff
@@ -60,75 +110,138 @@ const SIZE_MAP: Record<IconSize, string> = {
 };
 
 const ICON_REGISTRY: Record<
-  string,
+  IconId,
   {
-    Component: ComponentType<{ id?: string; size: string; className: string }>;
+    Component: ComponentType<GiIconProps>;
     disabledClass?: string;
   }
```
DmitryB-NF: 💡 This type is excessive now. We can do that as a clear map now
```
const ICON_REGISTRY = {
  accessibility_new: AccessibilityIcon,
  ...
}
```
to let the compiler set the type and use keys as a source of truth for the type. This way, we do not need even the `iconIds` map, which is basically `_.keys(ICON_REGISTRY)`

packages/react/src/icon/icon.tsx
DmitryB-NF: For backward compatibility, does it make sense to mark the prop as `@deprecated` to use together with `useFontIcon` (deprecated as well)? This way, it is still possible to use font icons with fill, but as a backup if any of the clients need this for some reason. 
With this approach, in the next ticket for full deprecation, we can still allow the workaround to load fonts, but as a temp solution for migration (marked as deprecated migration path in Docs)

packages/react/src/icon/icon.tsx
DmitryB-NF: I would suggest marking `ariaHidden` as deprecated to suggest using `ariaLabel` to calculate `hidden` automatically and explain as hidden will not be used if label is passed

packages/react/src/icon/icons.ts
DmitryB-NF: Let's remove this file and replace the export with `_.keys(ICON_REGISTRY)` as it is a public API with a message to use types like `IconProps['icon']`

packages/react/src/label/label.stories.tsx
```diff
@@ -115,7 +115,7 @@ export const WithRichText: Story = {
     children: (
       <div className="gi-flex">
         Here is a rich label &nbsp;<Link href="#">Click here </Link>
-        <Icon icon="arrow_drop_up" />
+        <ArrowDropUpIcon label="arrow_drop_up" />
```
DmitryB-NF: `label` is excessive

packages/react/src/select/select-native.tsx
```diff
@@ -27,10 +27,14 @@ export const SelectNative: FC<SelectProps & { containerProps?: any }> = ({
 }) => {
   return (
     <div className="gi-select-container" {...containerProps}>
-      <select className={cn('gi-select', className)} {...props}>
+      <select className={clsx('gi-select', className)} {...props}>
         {children}
       </select>
-      <Icon className="gi-select-icon" icon="keyboard_arrow_down" data-disabled={props?.disabled?.toString()} />
+      <KeyboardArrowDownIcon
+        className={clsx('gi-select-icon', {
```
DmitryB-NF: 💡Just in case, if changing the style, we can inline `gi-select-icon` as well as a quick fix

packages/react/tests/visual.spec.ts-snapshots/form-inputfile--with-list-of-uploaded-files-android-linux.png
DmitryB-NF: The new SVG looks better and more consistent than fonts. So, this is a good change

DmitryB-NF (changes requested):

packages/react/src/icon/icon.tsx
Ademsk1: done! `if` statement now checks if either `useFontIcon` **or** `filled` is there, and if so, defaults to the fallback font icon.

packages/react/src/icon/icon.tsx
```diff
@@ -1,56 +1,109 @@
...
-   * Use font icon instead of svg
+  /** @deprecated This prop will be removed in future versions */
+  filled?: boolean;
+  /**Use font icon instead of svg
    * Used as a fallback for consistency during Mitosis migration.
-   */
+   * @deprecated This prop will be removed in future versions */
   useFontIcon?: boolean;
-} & Omit<ComponentPropsWithoutRef<'span'>, 'children'>;
+} & Omit<ComponentPropsWithoutRef<'span'>, 'children' | 'onClick'>;
```
Ademsk1, line 104: ComponentPropsWithoutRef included `onClick` as well. And so in the type declaration for `IconProps`, the deprecated message would not appear.

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -16,7 +16,7 @@ export function IconGridLayout() {
           <li key={icon} className="inline-block w-32">
             <div className="border rounded-md transition-colors relative">
               <div className="flex justify-center px-10 py-12">
-                <Icon icon={icon} size="lg" />
+                <Icon icon={icon as IconProps['icon']} size="lg" />
```
DmitryB-NF, line 19: Does it make sense to change this usage too to use icons directly?

packages/react/src/icon/icon.stories.tsx
```diff
@@ -21,10 +21,6 @@ export const Default: Story = {
       options: ['sm', 'md', 'lg', 'xl'],
       description: 'Specify the size of the icon',
     },
-    filled: {
```
DmitryB-NF, line 24: ⚠️ should it be deprecated?

apps/docs/content/3-components/2-library/icon/react.mdx
```diff
@@ -37,6 +37,10 @@ component:
           ofType: 'boolean'
           description: 'Specifies whether the icon should be displayed as inline.'
           required: false
+        - name: 'dataTestId'
```
DmitryB-NF, line 44: 💡 We need to add a `useFontIcon` internal prop as well to allow an explicit fallback. We can add this in the next PR with full deprecation

DmitryB-NF (commented):
We need to clean up the removed `filled` prop, and we are good to go!
@Ademsk1 could you check other comments for removing the `Icon` usage in docs and `useFontIcon`? We need to add the `useFontIcon` fallback explicitly to the docs and update the Docs website in a separate PR.

apps/docs/content/3-components/2-library/icon/react.mdx
Ademsk1: updated it here.

apps/docs/src/components/document/resources/icon-grid-layout.tsx
Ademsk1: imported all and iterated through. Looks good.

apps/docs/content/3-components/2-library/icon/react.mdx
```diff
@@ -19,15 +19,20 @@ component:
           required: false
         - name: 'filled'
           ofType: 'boolean'
-          description: 'Specifies whether the icon has a filled style.'
+          description: 'Specifies whether the icon has a filled style. (**Deprecated**)'
```
ramonmulia: should we follow the Paragraph `deprecated` message pattern? https://github.com/ogcio/govie-ds/blob/5bdcf26005c7b6f76562ec6c81e8d59e2f1004e6/apps/docs/content/3-components/2-library/paragraph/react.mdx?plain=1#L18

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,29 +1,31 @@
...
 
-const icons = [...Icons.sort()];
-
 export function IconGridLayout() {
   return (
     <ul className="flex flex-wrap gap-2 p-0 gi-not-prose">
-      {icons.map((icon) => {
-        const iconName = titleCase(icon).replaceAll(' Alt', '');
+      {Object.entries(IconsList).map(([name, Icon]) => {
+        console.log(IconsList);
```
ramonmulia: remove `console.log`

apps/docs/content/3-components/2-library/icon/react.mdx
```diff
@@ -19,15 +19,20 @@ component:
           required: false
         - name: 'filled'
           ofType: 'boolean'
-          description: 'Specifies whether the icon has a filled style.'
+          description: '@deprecated - Specifies whether the icon has a filled style.'
```
DmitryB-NF: Let's add a comment that is applicable only to Material Font icons with `useFontIcon`. It should be clear why the field is deprecated and how to replace it

apps/docs/content/3-components/2-library/icon/react.mdx
```diff
@@ -19,15 +19,20 @@ component:
...
+          required: false
+        - name: 'useFontIcon'
+          ofType: 'boolean'
+          description: '@deprecated - Enforce icon source fallback to material symbols font icons. '
```
DmitryB-NF: It is better to change the message to explain the fallback to Font Icons instead of the default SVG approach and required only for transition period

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,29 +1,30 @@
...
-          <li key={icon} className="inline-block w-32">
+          <li key={name} className="inline-block w-48">
             <div className="border rounded-md transition-colors relative">
               <div className="flex justify-center px-10 py-12">
-                <Icon icon={icon} size="lg" />
+                <Icon />
               </div>
               <div className="absolute bottom-1 right-1">
-                <CopyToClipboardButton text={icon} />
+                <CopyToClipboardButton text={iconName} />
```
DmitryB-NF: ⚠️ Should this be a correct component/icon name rather than a UI label? I suppose this is the main idea to be able to copy the actual reference of the Icon

packages/react/src/icon/icon.tsx
```diff
@@ -150,3 +259,7 @@ export const Icon = forwardRef<HTMLSpanElement, IconProps>(
 );
 
 export default Icon;
+
+export type IconId = keyof typeof ICON_REGISTRY;
```
DmitryB-NF, line 261: I suppose we can remove this after fixing Docs

DmitryB-NF (changes requested):
@Ademsk1 I think it is better to move the Docs changes to a separate PR, since they require more changes. There is a new ticket AB#39308 for it already

apps/docs/src/components/document/resources/icon-grid-layout.tsx
DmitryB-NF: Replacement of `Icon` requires more changes here. Let's move those changes to a separate PR

packages/react/src/icon/icon.tsx
```diff
@@ -1,56 +1,109 @@
...
   ariaHidden?: boolean;
   ariaLabel?: string;
   inline?: boolean;
   className?: string;
+  dataTestId?: string;
+  /** @deprecated Use `IconButton` to add interactivity to your icons */
   onClick?: MouseEventHandler<HTMLSpanElement>;
-  /**
-   * Use font icon instead of svg
+  /** @deprecated This prop will be removed in future versions */
```
DmitryB-NF: ⚠️ all deprecation warnings should have an explanation and a migration path

apps/docs/content/3-components/2-library/icon/react.mdx
DmitryB-NF: Something like
> Specifies whether the icon has a filled style. **Deprecated:** uses `useFontIcon` Font fallback instead of SVG to render the component.

packages/react/src/icon/icon.tsx
```diff
@@ -1,56 +1,109 @@
...
 
-export type IconId = (typeof iconIds)[number];
 export type IconSize = 'sm' | 'md' | 'lg' | 'xl';
 
 export type IconProps = {
   icon: IconId;
   size?: IconSize;
-  filled?: boolean;
   disabled?: boolean;
+  /** @deprecated For SVG components, ariaHidden is implicitly true for icons with no `ariaLabel`/`label` */
```
DmitryB-NF: Could be something
> `ariaHidden` will be calculated based on `ariaLabel` prop. `True` if `ariaLabel` is not passed

apps/docs/content/3-components/2-library/icon/react.mdx
```diff
@@ -94,4 +102,12 @@ Icons help users quickly recognise actions, states and categories. Use them to c
 
 <ComponentPropertiesTable componentId="icon-react" />
 
+### SVG and font icons
+
+Earlier versions of the package rendered all icons using Material Symbols font icons. The package now ships SVG icons internally, and `<Icon />` uses these by default. If you pass `filled` or `useFontIcon` as `true`, the component falls back to Material Symbols font icons instead.
+
+The package will soon stop including the Material Symbols stylesheet. If you still need font icons - for example, for filled styles not yet available as SVGs — add the font resource to your project so `<Icon filled={true} />` continues to work. Follow these [instructions](https://developers.google.com/fonts/docs/material_symbols#use_in_web) to set this up.
+
+Future work is planned to add filled SVG variants for all supported icons, which will remove the need for the deprecated `filled` and `useFontIcon` props and the Material Symbols font dependency.
```
DmitryB-NF: ⚠️ Let's decide this within the PR for disabling Material Symbols support. We still should have a workaround to enable Font Icons, so this comment will not be correct.

packages/react/src/icon/icon.stories.tsx
```diff
@@ -2,15 +2,19 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
 import { within, expect } from 'storybook/test';
 import { Icon } from './icon.js';
 
+const FONT_ICON_MIGRATION_DOCS =
+  '**Migration:** Icons now render as SVGs by default. Remove `filled` and `useFontIcon` unless you explicitly need Material Symbols font icons. The package will stop including the Material Symbols stylesheet — [import fonts manually](https://developers.google.com/fonts/docs/material_symbols#use_in_web) if you still rely on these props. Filled SVG variants are planned, which will remove the need for both props. See the [Icon React docs](https://ds.services.gov.ie/components/library/icon/react/) for details.';
```
DmitryB-NF: > The package will stop including the Material Symbols stylesheet
This is outside of the PR work. Let's avoid adding it as it may not be correct based on the disabling of Font icons

packages/react/src/icon/icon.stories.tsx
```diff
@@ -68,13 +87,6 @@ export const ExtraLarge: Story = {
   },
 };
 
-export const Filled: Story = {
-  args: {
-    icon: 'thumb_up',
-    filled: true,
-  },
-};
```
DmitryB-NF, line 76: ⚠️ I am not sure we can remove this as there is a support for the fallback

DmitryB-NF (changes requested):

apps/docs/src/components/document/resources/icon-grid-layout.tsx
Ademsk1: reverted

packages/react/src/icon/icon.tsx
Ademsk1: added

DmitryB-NF (approved):
Well done!

PR #976, opened 2026-06-03
feat(AB#38534): Adding FooterSection atom

ramonmulia:
## Description

Adding FooterSetion atom

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#38534](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38534)


packages/core/atoms/FooterSection.lite.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { tv } from 'tailwind-variants';
+import { clamp } from './utilities';
+import GiBox from './Box.lite';
+import type { Props as BoxProps } from './Box.lite';
+
+useMetadata({ angular: { selector: 'gi-footer-section' } });
+
+export const FooterSectionVariant = {
+  DEFAULT: 'default',
+  META: 'meta',
```
DmitryB-NF: ⚠️ `Default` and `Meta` seem a bit confusing for the Section. It is difficult to understand that the `meta` option is a darker one named `Utility` in the Design. I would suggest using `Primary`/`Secondary` or maybe `Primary`/`Utility`

packages/core/atoms/FooterSection.lite.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export default function FooterSection(props: Props) {
+  return (
+    <GiBox
+      id={props.id}
+      className={styles({
+        variant: getVariant(props.variant),
+        className: props.className,
+      })}
+      role={props.role ?? (props.ariaLabel || props.ariaLabelledBy ? 'region' : undefined)}
```
DmitryB-NF: I'm wondering if we need `arialLabel` and a specifiс `role` here. In the correct way, it would be nice to place the `nav` tag inside the client application, where the `region` role may not make sense.
What is the reason for such a choice?

ramonmulia: good call! I think it makes sense to remove these props since `FooterSection` is a visual primitive, not a landmark. The parent `Footer` already owns content info and consumers would put their own `nav` inside, so these props just add noise and risk to double labelling.

DmitryB-NF (approved):

PR #982, opened 2026-06-09
feat(AB#38402): Adding LinkButton atom component

ramonmulia:
## Description

Adding LinkButton atom component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues
[AB#38402](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38402)


packages/core/atoms/LinkButton.lite.tsx
```diff
@@ -0,0 +1,154 @@
...
+  );
+}
+
+export const linkButtonStyles = tv({
+  extend: buttonBaseStyles,
+  base: ['gi-gap-2', 'gi-no-underline', 'hover:gi-no-underline', 'aria-[disabled=true]:gi-pointer-events-none'],
+  variants: {
+    size: buttonSizeVariants,
+  },
+  compoundVariants: [
```
DmitryB-NF: 💡 I suppose we need to override all `focus` styles to ensure they apply to the `<a>` tag, right? Does it make sense to move this into the `buttonBaseStyles` by replacing `enabled:focus` with `focus`? This way, we can remove all `compoundVariants` and make `buttonBaseStyles` agnostic to avoid `<button>` specific

packages/core/atoms/utilities.ts
```diff
@@ -17,6 +27,11 @@ export const getJustify = (x: (typeof Justify)[keyof typeof Justify] | undefined
 export const getOrientation = (x: (typeof Orientation)[keyof typeof Orientation] | undefined) =>
   clamp(x, Orientation, Orientation.HORIZONTAL);
 
+export const getVariant = (x: (typeof Variant)[keyof typeof Variant] | undefined) => clamp(x, Variant, Variant.PRIMARY);
```
DmitryB-NF: ❓ `Variant` seems related to the button only. Should it be part of the Button component rather than generic constants?

packages/core/atoms/LinkButton.lite.tsx
```diff
@@ -0,0 +1,154 @@
...
+      href={props.disabled ? undefined : props.href}
+      ref={props.ref}
+      class={linkButtonStyles({
+        variant: getVariant(props.variant),
+        appearance: getAppearance(props.appearance),
+        size: getSize(props.size),
+        disabled: !!props.disabled,
+        class: props.className,
+      })}
+      role={props.disabled ? 'link' : undefined}
```
DmitryB-NF: I think `role` is always `'link'` by default. So, this check has no effect

packages/core/atoms/LinkButton.lite.tsx
```diff
@@ -0,0 +1,154 @@
+import { tv } from 'tailwind-variants';
+import { useMetadata } from '@builder.io/mitosis';
+import { Variant, Appearance } from './constants';
+import { buttonBaseStyles, ButtonSize, buttonSizeVariants, getSize } from './Button.lite';
+import { getVariant, getAppearance } from './utilities';
+
+useMetadata({ angular: { selector: 'gi-link-button' } });
+
+export type Props = {
```
DmitryB-NF: 💡 Can we extenв `LinkProps` here to make sure all types are consistent? Basically, we can use the `Link` here as a base instead of `<a>` to validate types as well - even if it has no practical effect, validation and consistency could be valuable to make sure that changes to the `Link` require changes to the `LinkButton`

packages/core/atoms/storybook/LinkButton.meta.ts
```diff
@@ -0,0 +1,249 @@
...
+    children: 'LinkButton',
+    href: 'https://www.gov.ie',
+    variant: Variant.PRIMARY,
+    appearance: Appearance.DEFAULT,
+    size: ButtonSize.MD,
+    disabled: false,
+    dataTestId: 'link-button-test',
+    id: 'link-button-example',
+  },
+  argTypes: {
```
DmitryB-NF, line 21: Can we extend `Link` meta `here`, as the list should be fully the same except for new props?
I would suggest reusing the existing code as much as possible to preserve consistency

packages/react/src/index.ts
```diff
@@ -87,6 +87,7 @@ export { default as InsetText, type Props as InsetTextProps } from './atoms/Inse
 export { default as Text, type Props as TextProps } from './atoms/Text.js';
 export type { LabelTextProps } from './label/types.js';
 export { Link, type LinkProps } from './link/link.js';
+export { default as LinkButton, type Props as LinkButtonProps } from './atoms/LinkButton.js';
```
DmitryB-NF: ⚠️ I think we also need to add the `asChild` prop and Radix support to make sure `Link` can be Framework-native.
We may pass `style` prop as well to be consistent in all React components

packages/react/src/atoms/storybook/LinkButton.meta.ts
```diff
@@ -0,0 +1,301 @@
...
+      expect(element).toHaveAccessibleName('LinkButton');
+    });
+    await step('omits aria-disabled and aria-current when not set', async () => {
+      const element = canvas.getByTestId('link-button-test');
+      expect(element).not.toHaveAttribute('aria-disabled');
+      expect(element).not.toHaveAttribute('aria-current');
+    });
+  },
+};
+export const AllVariants = {
```
DmitryB-NF: 💡 It seems like the list of props exactly follows the `Button` Storybook. Does it make sense to reuse `Button` stories for the LinkButton component?
Basically, it tests the same Button styles once again. I would suggest reusing those tests and stories or adding a cross-link, e.g., reference `Button` Storybook to avoid duplication of the same stories

DmitryB-NF (changes requested):
The implementation is great but we may significantly simplify the approach

packages/core/atoms/LinkButton.lite.tsx
ramonmulia: yes that's correct, the `:enabled` prevents focus styles on `disabled` in `<button>` elements at the CSS level. Since`<a>` elements don't support `:enabled/:disabled` that's why `LinkButton` re-declares the focus styles with plain focus.
We can replace `enabled:focus:` with `focus:` in `buttonBaseStyles` but switching to `focus:` it would need the tv `disabled` variant to prevent focus styles on disabled buttons since `buttonBaseStyles` already has `disabled: true/false` variants, this is doable.
Based on that, should we refactor `Button` and `IconButton` on this PR, or in a dedicated one?

packages/core/atoms/storybook/LinkButton.meta.ts
ramonmulia: I think it can't extend directly,  `Link` has some props `LinkButton` doesn't use like (underline, visited, different variant/appearance enums), different styles and descriptions so extending it would need `Omit` props and overrides on most of the rest, I think it would be harder to read than the standalone definition, what do you think?

packages/core/atoms/LinkButton.lite.tsx
ramonmulia: good call, but composing with `Link` at the component level (instead primitive `<a>`) was also attempted and didn't work tv can't resolve `gi-` class conflicts when both style sets land on the same `<a>` (e.g. `Link` applies `focus-visible:gi-outline-none`which kills LinkButton's `focus:gi-outline`), their focus rings are also incompatible (box-shadow vs outline+border). The only shared logic is 2 lines of target and rel for external links.

packages/react/src/atoms/storybook/LinkButton.meta.ts
ramonmulia: yes, the story structure is similar to Button intentionally, same visual and enums. But stories can't be shared completely: `Button` asserts tagName === 'BUTTON' with native disabled, `LinkButton` asserts tagName 
  === 'A' with aria-disabled + href absence.
  
  For argTypes duplication I agreed. I'll extract shared button argTypes from Button.meta.ts so `LinkButton` and IconButton spread them instead of redeclaring.

packages/core/atoms/LinkButton.lite.tsx
ramonmulia: I think this is necessary, `<a>` only has the implicit link role when `href` is present, since we remove `href` when disabled to prevent keyboard navigation, the element loses its link semantics, so `role="link"` restores the screen reader announcement, without it, a disabled `LinkButton` would be announced as generic text rather than a disabled link.

packages/core/atoms/storybook/LinkButton.meta.ts
DmitryB-NF: We may extend only a couple of `args`. It is not all or nothing because `argTypes` is a simple map, where you can pick required props like
```
argTypes: {
  href: linkMeta.href,
  external: linkMeta.external,
  ...
}
```

packages/react/src/atoms/storybook/LinkButton.meta.ts
DmitryB-NF: We can still have an abstraction or helpers to reuse if it makes sense. I would rather suggest making a reference to the Button to avoid duplication if the extension is not straightforward

packages/core/atoms/LinkButton.lite.tsx
DmitryB-NF: ⚠️ Oh, I think this is the actual issue here. We should not remove `href` according to the `disabled` prop because there is no such behaviour in `Link`. `LinkButton` is supposed to be a **link** (`<a>` behaviour) that looks like a Button, but not a Button that accepts `href` for navigation.
I woulв suggest removing the `disabled` parameter entirely to make sure this is a pure Link

packages/core/atoms/LinkButton.lite.tsx
DmitryB-NF: As I understand, `enabled:focus` works exactly as `focus:` in form controls (including buttons). And those styles will not be picked up due to the native `disabled` prop `<button disabled=true>` by the native button behaviour.
So basically, minimal change in the `Button '.lite` can simplify the `LinkButton`. Something like
```
export const buttonBaseStyles = tv({
  base: [
    'gi-font-primary',
    'gi-border-solid',
    'gi-border-sm',
    'gi-flex',
    'gi-rounded-sm',
    'gi-items-center',
    'disabled:gi-cursor-not-allowed',
    'disabled:gi-pointer-events-none',
  ],
  variants: {
    variant: {
      primary: ['gi-border-transparent'],
      secondary: [],
      flat: ['gi-border-base-transparent'],
    },
    appearance: {
      default: '',
      light: '',
      dark: '',
    },
    disabled: {
      true: '',
      false: [
        'focus:gi-outline',
        'focus:gi-outline-sm',
        'focus:gi-outline-color-shadow-intent-focus-default',
        'focus:gi-outline-offset-0',
        'focus:gi-border-solid',
        'focus:gi-border-color-border-intent-focus-default',
        'focus:gi-border-sm',
        'focus:gi-rounded-sm',
      ],
    },
  },
  compoundVariants: [
    {
      variant: Variant.PRIMARY,
      disabled: false,
      class: ['focus:gi-shadow-color-border-intent-focus-light', 'focus:gi-shadow-[inset_0_0_0_2px]'],
    },
... // replace `enabled:focus:` with `focus:`
```
Will work correctly for LinkButton
```
export const linkButtonStyles = tv({
  extend: buttonBaseStyles,
  base: ['gi-gap-2', 'gi-no-underline', 'hover:gi-no-underline', 'aria-[disabled=true]:gi-pointer-events-none'],
  variants: {
    size: buttonSizeVariants,
  },
  defaultVariants: {
    size: ButtonSize.MD,
  },
});
```

ramonmulia: correct, so I'm going to include this change on this PR.

packages/react/src/atoms/storybook/LinkButton.meta.ts
ramonmulia: removed visual stories from `LinkButton`, kept only Default with link specific tests, also I added a link to `Button` storybook for variants, appearances, and sizes.

packages/core/atoms/storybook/LinkButton.meta.ts
```diff
@@ -0,0 +1,65 @@
...
+import { checker } from './utilities';
+import { linkMeta } from './Link.meta';
+import { buttonMeta } from './Button.meta';
+
+export const linkButtonMeta = {
+  tags: ['autodocs'] as string[],
+  title: 'Navigation/LinkButton',
+  args: {
+    children: 'LinkButton',
+    href: 'https://www.gov.ie',
```
DmitryB-NF: I would rather use DS webpage not to increase traffic to gov.ie in test purposes.

packages/core/atoms/storybook/LinkButton.meta.ts
```diff
@@ -0,0 +1,65 @@
...
+      description: {
+        component:
+          'LinkButton renders a native anchor element (`<a>`) styled like a Button. Use it for button-styled navigation links. It keeps the implicit `link` role so browser link behaviours (right-click "Open in new tab", Ctrl/Cmd+click) work correctly. Keyboard activation is Enter only (not Space) — this is native anchor behaviour.\n\nVisual variants, appearances, and sizes are identical to [Button](/docs/components-button--docs).',
+      },
+    },
+  },
+};
+
+export const Default = {
+  tags: ['skip-playwright'],
```
DmitryB-NF: We can still include it to Playwright to have a single base view check

packages/react/src/LinkButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+  className,
+  style,
+  styles,
+  ariaCurrent,
+  ariaLabel,
+  ariaLabelledBy,
+  ariaDescribedBy,
+  ariaHidden,
+  tabIndex,
+  download,
```
DmitryB-NF: Does it make sense to remove this to pass to GiLink with `rest`?

apps/docs/content/3-components/2-library/link-button/design.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+# LinkButton
+
+LinkButton renders a native anchor element (`<a>`) styled like a Button. Use it for button-styled navigation links where the destination is a URL.
+It keeps the implicit `link` role so browser link behaviours (right-click "Open in new tab", Ctrl/Cmd+click) work correctly.
+
+<Section current="design">
+
+<ComponentPreview>
+  <LinkButton href="https://www.gov.ie">This is a link button</LinkButton>
```
DmitryB-NF: Let's use the DS docs home page

DmitryB-NF (commented):
The implementation looks nice and clean!
Can we check Playwright and demo links?

apps/docs/content/3-components/2-library/link-button/react.mdx
```diff
@@ -0,0 +1,144 @@
...
+          ofType: 'number'
+          description: 'Controls the tab order. Use -1 to remove from the natural tab order in composite widgets.'
+          required: false
+        - name: 'lang'
+          ofType: 'string'
+          description: 'Language of the link text, when it differs from the surrounding content. Maps to the HTML lang attribute.'
+          required: false
+        - name: 'onClick'
+          ofType: '(event: any) => void'
+          description: 'Click event handler.'
```
Ademsk1, line 62: What's the idea around an `onClick` event handler for a link? I'm not sure it makes sense to me. Are there scenarios we want to support around this?

apps/docs/content/3-components/2-library/link-button/react.mdx
```diff
@@ -0,0 +1,144 @@
...
+        - name: 'ariaLabelledBy'
+          ofType: 'string'
+          description: 'Points to element id(s) whose content labels the link button. Maps to aria-labelledby.'
+          required: false
+        - name: 'ariaDescribedBy'
+          ofType: 'string'
+          description: 'Points to element id(s) whose content describes the link button. Maps to aria-describedby.'
+          required: false
+        - name: 'ariaHidden'
+          ofType: 'boolean | "true" | "false"'
```
Ademsk1: Should this just be `ofType: 'boolean'`

packages/core/atoms/Button.lite.tsx
```diff
@@ -332,23 +347,26 @@ export const buttonBaseStyles = tv({
...
-  Object.values(Variant).includes(x) ? x : Variant.PRIMARY;
-const getAppearance = (x: Props['appearance']) =>
-  x === Appearance.LIGHT || x === Appearance.DARK ? x : Appearance.DEFAULT;
-const getSize = (x: Props['size'] = ButtonSize.MD) => (Object.values(ButtonSize).includes(x) ? x : ButtonSize.MD);
+export const getSize = (x: Props['size'] = ButtonSize.MD) => clamp(x, ButtonSize, ButtonSize.MD);
+
+export const getVariant = (x: (typeof Variant)[keyof typeof Variant] | undefined) => clamp(x, Variant, Variant.PRIMARY);
+
+export const getAppearance = (x: (typeof Appearance)[keyof typeof Appearance] | undefined) =>
+  clamp(x, Appearance, Appearance.DEFAULT);
```
Ademsk1, line 372: It makes sense to keep this in Button! Nice!

packages/react/src/atoms/LinkButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+      style={props.styles}
+      target={props.target ?? (props.external ? '_blank' : undefined)}
+      rel={props.rel ?? (props.external ? 'noreferrer noopener' : undefined)}
+      download={props.download}
+      aria-current={props.ariaCurrent}
+      aria-label={props.ariaLabel}
+      aria-labelledby={props.ariaLabelledBy}
+      aria-describedby={props.ariaDescribedBy}
+      aria-hidden={props.ariaHidden}
+      tabIndex={props.ariaHidden ? -1 : props.tabIndex}
```
Ademsk1, line 48: It's worth mentioning in the docs for `tabIndex` that its set to -1 when `ariaHidden` is true.

packages/react/src/atoms/LinkButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+      lang={props.lang}
+      onClick={(event) => props.onClick && props.onClick(event)}
+      onFocus={(event) => props.onFocus && props.onFocus(event)}
+      onBlur={(event) => props.onBlur && props.onBlur(event)}
+      onKeyDown={(event) => props.onKeyDown && props.onKeyDown(event)}
+      onKeyUp={(event) => props.onKeyUp && props.onKeyUp(event)}
```
Ademsk1, line 54: I know we've done it this way for other components - but should we use optional chaining instead for better readability? 

```suggestion
      onClick={(event) => props.onClick?.(event)}
      onFocus={(event) => props.onFocus?.(event)}
      onBlur={(event) => props.onBlur?.(event)}
      onKeyDown={(event) => props.onKeyDown?.(event)}
      onKeyUp={(event) => props.onKeyUp?.(event)}
```

apps/docs/content/3-components/2-library/link-button/react.mdx
ramonmulia: `onClick` forwards the native anchor click event, it is useful for analytics tracking and others usecases,we're just exposing what the DOM already provides, not adding custom behaviour.

packages/react/src/atoms/LinkButton.tsx
ramonmulia: hmm actually optional chaining works fine for React, but not for Angular. Since Mitosis generates code for both frameworks, in Angular it changes `@Output` with `EventEmitter.emit()` to `@Input` with a direct function call, breaking the Angular event binding. 

the `&&` guard pattern is necessary and intentional, it is safe and ensures correct output for both targets.

Ademsk1: Good to know!

DmitryB-NF (approved):

PR #986, opened 2026-06-15
feat(AB#39534): remove filled prop and references

Ademsk1:
## Description

Removes `filled` prop from the `<Icon/>` component

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist


packages/react/src/icon/icon.tsx
```diff
@@ -246,7 +243,6 @@ export const Icon = forwardRef<HTMLSpanElement, IconProps>(
         )}
         style={{
           fontSize,
-          fontVariationSettings: `'FILL' ${filled ? 1 : 0}, 'wght' 400, 'GRAD' 0, 'opsz' ${fontSize}`,
```
Ademsk1, line 249: Because there was an issue with how `fontVariationSettings` parsed `fontSize`, this line never worked. Hence the `fontVariationSettings` defaulted to `normal` - which is the standard for if `fontVariationSettings` isn't specified in the first place.

packages/react/src/icon/icon.tsx
```diff
@@ -97,8 +97,6 @@ export type IconProps = {
   dataTestId?: string;
   /** @deprecated Use `IconButton` to add interactivity to your icons */
   onClick?: MouseEventHandler<HTMLSpanElement>;
-  /** @deprecated Work is planned to introduce SVG rendered filled icons in-place of Material symbols. This prop forces a Material Symbols font fallback to enable filled icons to be rendered. The package will stop including the Material Symbols stylesheet; import fonts manually if you still need this. See [here](https://ds.services.gov.ie/components/library/icon/react/) for more details.  */
-  filled?: boolean;
```
DmitryB-NF, line 101: Does it make sensу to leave the prop but change the message as unsupported anymore due to SVG migration?

DmitryB-NF (commented):
The fix is correct, but we may consider having exactly the same public API to avoid type errors in case of accidential usage

packages/react/src/icon/icon.tsx
```diff
@@ -207,7 +207,6 @@ export const Icon = forwardRef<HTMLSpanElement, IconProps>(
       className,
       onClick,
       useFontIcon,
-      filled,
```
DmitryB-NF, line 210: ⚠️ it is better to keep this as it is. Otherwise, the prop will be passed further with `{...props}`

Ademsk1: Good catch!

DmitryB-NF (approved):
Nicely done!

PR #987, opened 2026-06-16
feat(AB#39308): svg icons docs

Ademsk1:
## Description

Generated all icons from the SVG mitosis source to the public folder for download. Adds in search for icons. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#39308](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/39308)


apps/docs/public/icons/apps.svg
Ademsk1, line 1: These changes are accurate

apps/docs/public/icons/Accessibility.svg
DmitryB-NF, line 1: ⚠️ The original proposal was to generate all SVG icons on the fly based on React SVG components to have a single source of truth. Can we investigate this solution?

apps/docs/content/3-components/2-library/buttons/design.mdx
```diff
@@ -95,13 +113,14 @@ Buttons are interactive elements that enable users to trigger actions or navigat
...
       <Button>
-        <Icon icon="mail" />
+        <MailIcon/>
         <span>Send mail</span>
       </Button>
     </div>
     <div className="gi-mt-2 gi-text-sm gi-text-gray-600">
       Descriptive label with relevant icon that enhances understanding
     </div>
+
```
DmitryB-NF: Excessive line?
I suppose the only change in the file is to apply 2 Icons. Does it make sense to avoid reformatting for 2 valuable lines?

apps/docs/content/3-components/2-library/icon-button/design.mdx
```diff
@@ -36,18 +38,43 @@ Icon buttons are ideal for quick, easily recognised actions, but they rely on fa
...
 ## Accessibility
 
 - Icon buttons have no visible text, so assistive technologies rely on a programmatic accessible name to announce the button's purpose. Without one, screen readers announce the button as simply "button" with no context.
 - Every icon button **must** have an accessible name. Use one of the following approaches:
 
-| Approach         | When to use                                             | Example                                                       |
-| ---------------- | ------------------------------------------------------- | ------------------------------------------------------------- |
-| `ariaLabel`      | Most common. Provide a short text description directly. | `<IconButton ariaLabel="Close dialog">...</IconButton>`       |
-| `ariaLabelledBy` | When a visible label exists elsewhere on the page.      | `<IconButton ariaLabelledBy="section-title">...</IconButton>` |
+<Table layout="fixed">
```
DmitryB-NF, line 51: Is there an issue with converting the original table here? I am not sure that for the MD file, it would be a better approach

apps/docs/src/components/document/common/mdx.tsx
```diff
@@ -106,6 +106,87 @@ import {
   Tooltip,
   generateSvgPlaceholderDataUrl,
 } from '@ogcio/design-system-react';
+
+import { AccessibilityIcon } from '@ogcio/design-system-react/icons';
```
DmitryB-NF: We can import all the icons at once if we need like
```
import * as Icons from '@ogcio/design-system-react/icons';

{
  ...Icons
}
```

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,39 +1,61 @@
...
+            setIconFilter(event.target.value);
+          }}
+          value={iconFilter}
+          placeholder="Accessibility"
+          type="search"
+          className="gi-max-w-md"
+        />
+      </FormField>
+      <ul className="flex flex-wrap gap-2 p-0 gi-not-prose">
+        {_.toPairs(AllIcons).map(([name, Icon]) => {
```
DmitryB-NF: could be simplified as `_.map(AllIcons, (Icon, name) => {...})`

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,39 +1,61 @@
...
+          if (!new RegExp(iconFilter, 'gi').exec(name)) {
+            return null;
+          }
```
DmitryB-NF: ❓ Why do we need this check?

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,39 +1,61 @@
...
+          onChange={(event) => {
+            setIconFilter(event.target.value);
+          }}
+          value={iconFilter}
+          placeholder="Accessibility"
+          type="search"
+          className="gi-max-w-md"
+        />
+      </FormField>
+      <ul className="flex flex-wrap gap-2 p-0 gi-not-prose">
```
DmitryB-NF: Btw, we can use `Grid` component to avoid adding flex styles manually

DmitryB-NF (changes requested):
We may need to get rid of duplicated SGVs in docs if possible

The search is great, btw 🚀

apps/docs/content/3-components/2-library/buttons/design.mdx
Ademsk1: Good spot, not sure what made the formatter do this but I've fixed it to remove them. 

We may as well change it now, it will have to be formatted at some point.

apps/docs/content/3-components/2-library/icon-button/design.mdx
Ademsk1: It was actually broken previously, I thought a Table would be a better fit. See [here](https://ds.services.gov.ie/components/library/icon-button/design/). 

Couldn't find out how to render it properly with markdown tables - seems like contentlayers markdown doesn't support it?

DmitryB-NF: Yeah, I see that the markup is broken. If adding the `Table` usage is the best approach, we should keep it here

apps/docs/src/components/document/resources/icon-grid-layout.tsx
Ademsk1: Ah, I needed to filter and was using a lot of Regexp, but I think I'll just use `includes` instead.

apps/docs/src/components/document/resources/icon-grid-layout.tsx
Ademsk1: Going to use `reduce` here as I need to filter by the search

apps/docs/public/icons/Accessibility.svg
Ademsk1: I've made changes now to download-icon-button to accept a component as a prop. Now if an Icon is passed in, the URI is generated `onClick` of the `IconButton`. 

I removed the `a` tag. I was playing around a lot with it, and it **was** working with using `setState(uri)` for the `href` attribute when clicked. but I removed it as I felt that between clicking the a tag, and **then** setting the URI for the download, while also running `preventDefault` etc etc all felt quite flaky. However, to still enable the download, I create a link element programmatically, attach the href to that, and then simulate the `click`.

To keep consistency, I did this for all url props (text / href props) that could be passed for `<DownloadIconButton/>`. 

For rendering HTML from a React component, I'm following the [React docs](https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code) suggestion to use `createRoot` on the fly, read the computed innerHTML, and then generate a `data:` url. 

There was an alternative to all of this - use `renderToString` on the component **instead** of using `createRoot` - however that was not encouraged by the React documentation as its would unnecessarily increase bundle size. Arguably it makes the code a **lot** cleaner though : rather than adding all of this logic into `DownloadIconButton` I could simply pass in `renderToString(<Icon/>)` into the `text` attribute and it would work. 

Let me know your feelings on this implementation.

Ademsk1: also let me know if you'd like me to remove the svgs in `public/icons` as they are pretty redundant now... but maybe we want some URL for reaching them from other services? 

The icon duplication you mentioned I'm not sure what you mean ... could you expand on this? I have regenerated all of the icons now - so everything there **should** be completely new.

DmitryB-NF: Using the link approach is also fine if this improves behaviour. I still think that the `renderToString` could be a simple and straightforward solution, especially for icons in NextJS, where we can remove all excessive 'use client' declarations to generate the page on the backend, where the bundle size will not be affected. But I would use the current solution to avoid unnecessary refactoring.

And we need to remove all SVG files (old and newly created) from the `docs` app to use the `ds-react` package directly (`/icons`) to make sure we have a single source of truth for icons (fetch exactly the lib ones but not manual copies)

apps/docs/public/icons/apps.svg
DmitryB-NF: I think we need to remove this entirely

apps/docs/src/components/document/resources/download-icon-button.tsx
```diff
@@ -1,22 +1,56 @@
...
 
-export function DownloadIconButton({ name, href, text }: SvgIcon | TextIcon) {
+type SourceIcon = {
+  IconComponent: ComponentType<IconProps>;
+  name: string;
+  href?: never;
+  text?: never;
+};
+
+const encodedSvgURL = (text: string) => {
```
DmitryB-NF: It is better to move all private function after the public ones to improve readability

DmitryB-NF: Btw, the function could be inlined

apps/docs/src/components/document/resources/icon-grid-layout.tsx
DmitryB-NF: `ul` is still redundant

DmitryB-NF (changes requested):

apps/docs/src/components/document/resources/download-icon-button.tsx
```diff
@@ -28,32 +62,40 @@ export function DownloadIconButton({ name, href, text }: SvgIcon | TextIcon) {
...
-        appearance="light"
-        className={cn({
-          'text-green-600': downloaded,
-          'text-gray-600': !downloaded,
-        })}
-      />
-    </a>
+    <IconButton
+      onClick={handleClick}
+      size="small"
```
DmitryB-NF: it would be nice to use a new format for sizes (`sm`)

apps/docs/public/icons/Accessibility.svg
Ademsk1: Done! :D

apps/docs/src/components/document/resources/download-icon-button.tsx
Ademsk1: inlined and placed below :D

apps/docs/src/components/document/resources/icon-grid-layout.tsx
```diff
@@ -1,39 +1,70 @@
...
+      <Grid container gap={2}>
+        {_.reduce(
+          AllIcons,
+          (acc, Icon, name) => {
+            if (!name.toLowerCase().includes(iconFilter.toLowerCase())) {
+              return acc;
+            }
+            const iconName = titleCase(name).replace('Icon', '');
+            return [
+              ...acc,
```
DmitryB-NF: 💡 Overall, `reduce` is not recommended for the sake of readability, but recreation of the array on each step can affect performance. `_.filter().map()` is preferable here. Or we can also do `_.map().filter(Boolean)` to return `null` if icon is filtered out

DmitryB-NF (commented):
Finally, all SVG duplicates are removed! 
We just need to clean up the Grid a bit

DmitryB-NF (approved):

DmitryB-NF (approved):

PR #988, opened 2026-06-16
feat(AB#38538): Adding Footer atom

ramonmulia:
## Description

Adding Footer atom

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[38538](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/38538)


packages/react/src/footer/footer.tsx
```diff
@@ -49,7 +49,10 @@ export function Footer({
   logo,
   className = '',
   dataTestid,
-  ...props
```
DmitryB-NF, line 52: ⚠️ This could be a potential breaking change where props can stop working accidentally (it extends `HTMLDivElement`). Can we make sure we do not need anything else here?

packages/core/atoms/Footer.lite.tsx
```diff
@@ -0,0 +1,33 @@
...
+      data-testid={props.dataTestId}
+    >
+      {props.children}
+    </footer>
+  );
+}
+
+const footerStyles = tv({
+  base: 'gi-font-primary',
+});
```
DmitryB-NF: I think we may not need `font` style here as well as the `tv`

packages/react/src/footer/footer.tsx
ramonmulia: good call! since the `GiFooter` atom only accepts typed props, there's no way to forward all the HTML attributes from `{...props}`, so to be safe, I'll revert the legacy wrapper back to a raw `<footer>` and we can keep the atom usage in the `/next` entry point only, that way existing consumers won't break.

packages/core/atoms/Footer.lite.tsx
ramonmulia: true, removed.

DmitryB-NF (approved):

PR #992, opened 2026-06-18
fix(AB#39533): selectnext rich text

Ademsk1:
## Description

For SelectItemNext, the `inputText` was rendering `[object Object]` as it was directly reading the children and converting it to string. Now, we use a recursive method to find all text content within the children, operating similarly to the `.textContent` property for HTML elements. Storybook tests added and utility tests as well. 

## Type of Issue

- [ ] 🚀 Feature
- [x] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#39533](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/39533)


packages/react/src/utilities.ts
```diff
@@ -1,4 +1,4 @@
-import type { ReactElement, ReactNode } from 'react';
+import type { JSX, ReactElement, ReactNode } from 'react';
```
DmitryB-NF: unused?

packages/react/src/utils/utilities.ts
```diff
@@ -112,3 +112,38 @@ export function getSpecialComponentType(child: ReactNode): string | null {
...
+ * ```
+ * getTextContent(
+ *  <div>
+ *    <p>hello</p>
+ *    there
+ * </div>)
+ * returns "hello there"
+ * ```
+ */
+export const getTextContent = (node: ReactNode, _textContent: string[] = []): string => {
```
DmitryB-NF: I do not really think that the optimisation of having the `_textContent` makes a lot of sense here, especially when this is not supposed to be used in the large node trees.
So we can simplify this with an inlined `join`
```
export function getTextContent(node: ReactNode): string {
  if (node == null || typeof node === 'boolean') {
    return '';
  }

  if (typeof node === 'string' || typeof node === 'number') {
    return String(node);
  }

  if (Array.isArray(node)) {
    return node.map((child) => getTextContent(child)).join('');
  }

  if (isValidElement(node)) {
    return getTextContent((node.props as { children?: ReactNode }).children);
  }

  return '';
}
```

packages/react/src/select/select-next.stories.tsx
```diff
@@ -933,3 +937,100 @@ export const TestToggleDropdown: StoryObj = {
...
+          <SelectItemNext value="select-option" hidden>
+            Select Option
+          </SelectItemNext>
+          <SelectItemNext value="value-1">
+            <code>code block</code>
+          </SelectItemNext>
+          <SelectItemNext value="value-2">
+            <Paragraph>
+              <MenuIcon className="gi-inline" /> with icon
+            </Paragraph>
```
DmitryB-NF: ⚠️ I would suggest avoiding such examples, especially when the existing `SelectNext` is not really supposed to use such examples. I would suggest adding the default `<strong>` `<Text>` tags there to aply basic styles rather than having complex examples

packages/react/src/select/select-next.stories.tsx
```diff
@@ -933,3 +937,100 @@ export const TestToggleDropdown: StoryObj = {
...
+    expect(defaultSelect.getAttribute('value')).toBe('Alert');
+    await userEvent.click(searchSelect);
+    const iconOption = await canvas.findByTestId('option-value-2');
+    const alertOption = await canvas.findByTestId('option-value-3');
+    await userEvent.clear(searchSelect);
+    await userEvent.type(searchSelect, 'with', { delay: 200 });
+    await waitFor(() => {
+      expect(alertOption).not.toBeInTheDocument();
+    });
+    await userEvent.click(iconOption);
```
DmitryB-NF: The test is good, but we may add extra line breaks to have readable blocks for every check

packages/react/src/select/select-next.tsx
```diff
@@ -95,7 +96,7 @@ export const SelectNext = forwardRef<HTMLInputElement, SelectNextProps>(
       }
 
       if (found) {
-        const foundValue = found.props.children?.toString() || '';
+        const foundValue = getTextContent(found);
         setInputValue(foundValue);
```
DmitryB-NF: `foundValue` is redundant 

```suggestion
        setInputValue(getTextContent(found));
```

packages/react/src/select/select-next.stories.tsx
```diff
@@ -933,3 +937,100 @@ export const TestToggleDropdown: StoryObj = {
...
+  tags: ['skip-playwright'],
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Rich text can be rendered within the SelectItemNext components, operating similarly to the `.textContent` HTMLElement method; traversing through the children and concatenating the content inside.',
+      },
+    },
+  },
+  render: () => (
```
DmitryB-NF: ⚠️ Render should have a `_props` argument, even if unused, to make sure Storybook "Show Code" functionality works as expected

```suggestion
  render: function Render (_props) {...}
```

DmitryB-NF (changes requested):
The fix is good, but we need to clean up the Storybook examples to make it clear for users

packages/react/src/utils/utilities.ts
Ademsk1: Applied your code, its good!

packages/react/src/select/select-next.stories.tsx
Ademsk1: I'm using <Text/> now and also a simple anon function call :D

packages/react/src/utils/utilities.ts
```diff
@@ -112,3 +112,39 @@ export function getSpecialComponentType(child: ReactNode): string | null {
...
+  if (node == null || typeof node === 'boolean') {
+    return '';
+  }
+
+  if (typeof node === 'string' || typeof node === 'number') {
+    return String(node);
+  }
+
+  if (Array.isArray(node)) {
+    return node.map((child) => getTextContent(child)).join('');
```
ramonmulia, line 138: as an alternative, you can use `_.compact` here. It removes empty values before join, which helps avoid extra spaces when `{null}` children are rendered.

packages/react/src/utils/utilities.ts
```diff
@@ -1,4 +1,4 @@
-import type { ReactNode } from 'react';
+import type { ReactNode, JSX, ReactElement, PropsWithChildren } from 'react';
```
ramonmulia: `JSX` is imported but never used.

Ademsk1: oops I did it again

packages/react/src/utils/utilities.ts
Ademsk1: good function to know! I don't think this is useful here though, as if the node **is** null, then empty string is returned ''.

packages/react/src/utils/utilities.ts
Ademsk1: I simplified it further and removed the node == null || typeof node === 'boolean' check

packages/react/src/select/select-next.stories.tsx
```diff
@@ -933,3 +936,113 @@ export const TestToggleDropdown: StoryObj = {
...
+            <SelectItemNext value="value-3">
+              <Paragraph>
+                {(() => {
+                  return (
+                    <Text dataTestId="fn-call" className="gi-font-bold">
+                      Function call
+                    </Text>
+                  );
+                })()}
+              </Paragraph>
```
DmitryB-NF: ❓ Is it expected? It would be nice to have only text-rich text inside Select Items

packages/react/src/select/select-next.stories.tsx
```diff
@@ -933,3 +936,113 @@ export const TestToggleDropdown: StoryObj = {
...
+          <FormFieldLabel>Default</FormFieldLabel>
+          <SelectNext aria-label="Select" defaultValue="value-2" data-testid="default">
+            <SelectItemNext value="select-option" hidden>
+              Select Option
+            </SelectItemNext>
+            <SelectItemNext value="value-1">
+              <code>code block</code>
+            </SelectItemNext>
+            <SelectItemNext value="value-2">
+              <Paragraph>
```
DmitryB-NF: Do we need a Paragraph here?

DmitryB-NF (commented):
The code is good, but it is better to fix the public example

packages/react/src/select/select-next.stories.tsx
Ademsk1: Removed and used a simple <Text> with bold font class

DmitryB-NF (approved):

PR #993, opened 2026-06-19
feat(AB#34759): Adding "multiple" prop in Autocomplete

ramonmulia:
## Description

Adding "multiple" prop in Autocomplete

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [x] 🌐 Docs

## Checklist

## Related Issues
[AB#34759](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/34759)


packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -150,6 +171,20 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     [children, dispatch, name, onAutocompleteChange, onSelectItem],
   );
 
+  const handleOnToggleItem = useCallback(
+    (toggleValue: string) => {
+      dispatch({ type: TOGGLE_SELECTED_ITEM, payload: toggleValue });
+      setTimeout(() => inputRef.current?.focus(), 0);
```
Ademsk1: What's the reason for the `setTimeout`?

packages/react/src/autocomplete/use-autocomplete-controller.ts
```diff
@@ -70,6 +68,21 @@ const reducer = (state: AutocompleteState, action: AutocompleteAction): Autocomp
     case SET_OPTION_TYPE: {
       return { ...state, optionType: action.payload };
     }
+    case TOGGLE_SELECTED_ITEM: {
+      const next = new Set(state.selectedValues);
+      if (next.has(action.payload)) {
+        next.delete(action.payload);
+      } else {
+        next.add(action.payload);
+      }
```
Ademsk1:
```suggestion
      if (!next.delete(action.payload)) {
              next.add(action.payload);
       }
```
the `.delete` method returns `true` or `false` depending on whether the element was there when it was "deleted". See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete)

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -70,6 +79,18 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
...
   }, [value]);
 
+  useEffect(() => {
+    if (multiple) {
+      onSelectionChange?.([...state.selectedValues]);
+    }
+  }, [state.selectedValues]);
+
+  useEffect(() => {
+    if (selectedValues && selectedValues.length < state.selectedValues.size) {
```
Ademsk1: Why the < sign?

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -165,33 +200,38 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
...
-            if (state.highlightedIndex === -1) {
-              return direction === -1 ? 0 : -1;
-            }
-            return state.highlightedIndex;
-          };
-
           const direction = event.key === 'ArrowDown' ? 1 : -1;
-          const newIndex = cycleEnabledIndex(state.autocompleteOptions, getStartingPoint(), direction);
-          dispatch({ type: SET_HIGHLIGHTED_INDEX, payload: newIndex });
+          const optionIndices = state.autocompleteOptions.map((_: unknown, index: number) => index);
```
Ademsk1: I don't think we need to put a type on the placeholder variable here

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -165,33 +200,38 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
...
+          const items = hasSelections ? [CLEAR_ALL_INDEX, ...optionIndices] : optionIndices;
+          const currentPos = items.indexOf(state.highlightedIndex);
+          const startPos = currentPos === -1 ? (direction === 1 ? -1 : 0) : currentPos;
+          const next = items[(startPos + direction + items.length) % items.length];
```
Ademsk1, line 235: why `+ items.length` ?

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -47,15 +50,21 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     name,
     value,
     id,
+    multiple,
+    onSelectionChange,
```
DmitryB-NF: I'm wondering if we may not need 2 separate props for the same action, `onSelectItem` and `onSelectionChange`. We may still use one function as a single select is basically a subset of multiselect
Or at least make the naming similar: `onSelectItem` and `onSelectChange`

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -70,6 +79,18 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     }
   }, [value]);
 
+  useEffect(() => {
+    if (multiple) {
+      onSelectionChange?.([...state.selectedValues]);
+    }
+  }, [state.selectedValues]);
```
DmitryB-NF: ⚠️ this seems like an antipattern to set the state and add effects. It is better to set the state and call the callback at the same handler to avoid re-renders and avoid visual glitches (useEffect is called after the full rendering of the entire tree and the parent component will have a correct state on thу first render)

packages/react/src/autocomplete/autocomplete.tsx
DmitryB-NF, line 235: Btw, we can use the `Array.at` function that accepts negative numbers
```
items.at(tartPos + direction) // [0,1,2].at(-1) -> 2
```

apps/docs/content/3-components/2-library/autocomplete/react.mdx
```diff
@@ -53,6 +53,22 @@ component:
...
           description: 'Controls the open/closed state of the dropdown (use for controlled mode).'
           required: false
+        - name: 'multiple'
+          ofType: 'boolean'
+          description: 'When true, allows selecting multiple options. Selected items are shown as a count badge in the input field.'
+          required: false
+        - name: 'onSelectionChange'
+          ofType: 'function'
+          description: 'Callback triggered when selections change in multiple mode. Receives an array of selected values.'
+          required: false
```
Ademsk1, line 63: Should this be required only if `multiple` is true?

packages/react/src/autocomplete/use-autocomplete-controller.ts
ramonmulia: good call!

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: good call! we use it to move focus back to the input after clicking a dropdown option, so without it, the browser moves focus to the clicked `<li>` and react has not finished re-rendering yet, but I ended up replacing it to `requestAnimationFrame` it is a better option on this case, basically it waits until the render is done before refocusing.

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: the `<` check is needed to make bidirectional sync work, It basically lets the parent update the internal state only when items are removed (e.g., from Chips), and ignores updates when it’s just syncing after an item was added.

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: It prevents negative index.. so by adding `items.length` before `%`, it wraps to the last item instead of going out of bounds.

apps/docs/content/3-components/2-library/autocomplete/react.mdx
ramonmulia: good call! but not sure if it needs to be required cause the component already handles selections internally, I improved the ts types to have this only when `multiple` is true.

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: renamed `onSelectionChange` to `onSelectChange` to align with the `onSelect*` naming pattern, they stay separate because `onSelectItem` gives the clicked value, while `onSelectChange` gives the full selection array.

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: good one, but `at()` handles negative indices nicely, but it doesn't wrap past the end (e.g. [0,1,2].at(3)   returns undefined instead of 0), since we need circular navigation in both directions, I think the modular arithmetic is necessary here.

packages/react/src/select/select-menu.tsx
```diff
@@ -182,8 +229,24 @@ export const SelectMenuOption = ({
       data-testid={dataTestid || `option-${value}`}
       {...props}
     >
-      <span className="gi-text-sm">{children}</span>
-      {selected && <Check className="gi-shrink-0" />}
+      {multiple ? (
+        <>
```
ramonmulia, line 243: using a visual only checkbox instead of `<InputCheckbox>` because placing a native `<input>` inside   `<li role="option">` is a nested interactive a11y violation, even with `tabIndex={-1}`, `aria-hidden="true"`, and `readOnly`, axe still flags it.. a native `<input>` is always considered interactive. 
The fix for this is is to not render an `<input>` at all.

Ademsk1 (commented):
The issue with the a11y is somewhat awkward. I see that now. Are there any downsides tro using the visual only checkbox? Is aria-selected ok?

packages/react/src/select/select-menu.tsx
```diff
@@ -101,7 +117,29 @@ export const SelectMenu = forwardRef<HTMLDivElement, SelectMenuProps>(
...
+                aria-selected={false}
+                className={cn('gi-select-option-item', {
+                  'gi-select-option-item-highlighted': clearAllHighlighted,
+                })}
+                tabIndex={-1}
+                onClick={onClearAll}
+              >
+                <Text size="sm">{t('autocomplete.clearAll', { defaultValue: 'Clear all selections' })}</Text>
+              </li>
+              <li role="presentation" className={dividerStyles()} />
```
DmitryB-NF: Does it make sense to directly use the `Divider` component here, not to expose styles separately?

packages/react/src/select/select-menu.tsx
```diff
@@ -101,7 +117,29 @@ export const SelectMenu = forwardRef<HTMLDivElement, SelectMenuProps>(
           })}
         </div>
       ) : (
-        <ul aria-label="options" role="listbox">
+        <ul
+          id={listboxId}
+          aria-label={listboxLabel || 'options'}
```
DmitryB-NF: I think we can use the default `ariaLabel` and `id` props here, as the list is the main component. It is better to add a custom label for no results separately

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -70,6 +79,12 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     }
   }, [value]);
 
+  useEffect(() => {
+    if (selectedValues && selectedValues.length !== state.selectedValues.size) {
+      dispatch({ type: SET_SELECTED_VALUES, payload: selectedValues });
+    }
```
DmitryB-NF: ⚠️ This check is quite dangerous because `selectedValues` is a public prop where the user can programmatically _replace_ one value with another. This way, this check will not handle it properly. 
We need a proper comparison meachanism here

packages/react/src/ChipGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+      role="group"
+      ariaLabel={t('chipGroup.label', { defaultValue: 'Selected items' })}
+      className={className}
+    >
+      {visibleItems.map((item) => (
+        <Chip key={item.value} label={item.label} onClose={() => onRemove(item.value)} />
+      ))}
+      {shouldCollapse && (
+        <Text className="gi-text-xs gi-font-bold gi-py-0.5 gi-px-1.5">
+          {t('chipGroup.moreItems', { count: hiddenCount, defaultValue: `+${hiddenCount} more` })}
```
DmitryB-NF: 💡 it would be nice to avoid adding i18n directly to components and use it as a prop where possible to add a proper layer of translations later on

packages/design/tailwind/css/components/input-checkbox.css
```diff
@@ -37,6 +37,16 @@
...
 
+/* Visual-only checkbox indicator for use inside interactive containers (e.g. role="option") where a real <input> would cause nested-interactive a11y violations. */
+.gi-input-checkbox-visual {
+  @apply gi-flex-none gi-rounded-sm gi-border-sm gi-border-solid gi-border-color-border-system-neutral-default gi-w-6 gi-h-6 gi-shrink-0;
+}
+
+.gi-input-checkbox-visual[data-checked='true']::before {
+  @apply gi-block gi--rotate-45 gi-relative gi-border-l-lg gi-border-b-lg gi-border-color-border-system-neutral-default gi-w-4 gi-h-2 gi-left-0.5 gi-top-1;
+  content: '';
+}
```
DmitryB-NF: ⚠️ we need to use the TV directly in components to avoid usage of the tailwing package. HTML lib will be reflected as a side effect

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -230,10 +268,20 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
...
         aria-label={labelText}
-        aria-labelledby={srOnlyLabelId}
+        aria-expanded={state.isOpen}
+        aria-controls={listboxId}
+        aria-activedescendant={
+          state.highlightedIndex === CLEAR_ALL_INDEX
+            ? `${listboxId}-clear-all`
+            : state.highlightedIndex >= 0
+              ? `${listboxId}-option-${state.highlightedIndex}`
+              : undefined
```
DmitryB-NF, line 292: ⚠️ it looks different from `${listboxId}-option-${optionProps.value}` for `SelectMenuOption` in https://github.com/ogcio/govie-ds/pull/993/changes#diff-450fd22df1057de4c44323d5f3c9a4edfb74d9df08abd2cbb24ed939d168fd3bR369. It can break the a11y for lists

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -150,6 +165,25 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     [children, dispatch, name, onAutocompleteChange, onSelectItem],
   );
 
+  const handleOnToggleItem = useCallback(
+    (toggleValue: string) => {
+      dispatch({ type: TOGGLE_SELECTED_ITEM, payload: toggleValue });
+      const next = new Set(state.selectedValues);
+      if (!next.delete(toggleValue)) {
+        next.add(toggleValue);
+      }
```
DmitryB-NF, line 180: This logic duplicates `TOGGLE_SELECTED_ITEM` handler. Does it make sense to calculate it here once and post to SET_VALUES with exactly the same value as for `onSelectChange`?

packages/react/src/ChipGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const hiddenCount = items.length - visibleCount;
+
+  return (
+    <Stack
+      direction="row"
+      gap={3}
+      wrap
+      align="center"
+      role="group"
+      ariaLabel={t('chipGroup.label', { defaultValue: 'Selected items' })}
```
DmitryB-NF: ⚠️ it is better to pass a `ariaLabel` prop

packages/react/src/select/select-menu.tsx
```diff
@@ -101,7 +117,29 @@ export const SelectMenu = forwardRef<HTMLDivElement, SelectMenuProps>(
...
+          role="listbox"
+          aria-multiselectable={multiselectable || undefined}
+        >
+          {onClearAll && (
+            <>
+              <li
+                id={listboxId ? `${listboxId}-clear-all` : undefined}
+                role="option"
```
DmitryB-NF: ❓ is it correct to set `option` for the action? I would suggest moving this out of the list to make sure the list includes only actual options and avoids a11y mismatches

packages/react/src/select/select-menu.tsx
DmitryB-NF: This is the correct approach. I think we may create a separate component as `Checkbox` as well as `InputCheckbox` to use it for general visual purposes, not only for Forms. This way, styles will be encapsulated in React TW and could be moved to HTML separatelly

packages/react/src/ChipGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const ChipGroupOverflow = {
+  Collapse: 'collapse',
+  Wrap: 'wrap',
+} as const;
+
+export type ChipGroupProps = {
+  items: { label: string; value: string }[];
+  onRemove: (value: string) => void;
+  overflow?: (typeof ChipGroupOverflow)[keyof typeof ChipGroupOverflow];
```
DmitryB-NF: According to the logic, `overflow` is excessive here because the `visibleCount` can control it the same way if passed:
```
overflow === ChipGroupOverflow.Collapse && items.length > visibleCount
```
vs
```
overflow === visibleCount && items.length > visibleCount
```
You can name it as `maxVisible` to indicate that this is the collapse max length

apps/docs/content/3-components/2-library/autocomplete/react.mdx
```diff
@@ -144,6 +188,38 @@ The Autocomplete component allows users to search and select from a list of opti
...
+
+When `multiple` is enabled, users can select more than one option. A count badge in the input field shows how many items are selected. A "Clear all selections" action is available in the dropdown.
+
+<StorybookFrame
+  heightClassName="h-[550px]"
+  componentId="autocomplete-react"
+  story="with-multiple"
+/>
+
+### Multi-select with Chips (Wrap)
```
DmitryB-NF: I think we can remove `(Wrap)` and make this behaviour by default, rather than `collapse`, to be more usable. In the collapse mode, values are not visible and have no tooltips, which makes it not a preferable way of displaying

DmitryB-NF (changes requested):
The Multiselect look good overall!
We need to fix values processing and a11y, and simplify components a bit

packages/react/src/select/select-menu.tsx
ramonmulia: good point, but the reason for `listboxId/listboxLabel` is that `SelectMenu` renders a wrapper `div` around the `ul role="listbox"`... So `id` and `aria-label` on the component would apply to the `div`, not the `listbox`, these custom props target the inner `ul` specifically (the combobox aria-controls must reference the listbox id, not the wrapper).

 we could improve this by passing them via a `listboxProps` object maybe?, or by restructuring so the `ul` is the top level element, but this approach would require refactoring the component, not sure if it is expected here, maybe we can have a follow-up task?

packages/react/src/autocomplete/autocomplete.tsx
ramonmulia: good catch, fixed.

packages/design/tailwind/css/components/input-checkbox.css
ramonmulia: since we haven't touched `InputCheckbox` yet, I thought it would be fine to include the styles in the `.css` file and then have a proper `checkbox` atom, but good point, that would affect the html pkg as a side effect, replaced by `CheckboxVisual` component.

packages/react/src/CheckboxVisual.tsx
```diff
@@ -0,0 +1,54 @@
+import { tv } from 'tailwind-variants';
+
+const checkboxVisualStyles = tv({
```
DmitryB-NF, line 13: Move private after public?

packages/react/src/autocomplete/autocomplete.stories.tsx
```diff
@@ -376,12 +398,207 @@ export const Test: Story = {
...
+        <FormFieldLabel>Multi Select (Chips Collapse)</FormFieldLabel>
+        <Stack gap={3}>
+          <Autocomplete {...props} multiple selectedValues={selectedValues} onSelectChange={setSelectedValues}>
+            {labelOptions.map(({ value, label }) => (
+              <AutocompleteItem value={value} key={value}>
+                {label}
+              </AutocompleteItem>
+            ))}
+          </Autocomplete>
+          <ChipGroup
```
DmitryB-NF, line 641: I am wondering if we can have a separate ticket to create a Storybook for the `ChipGroup` component. It is a good candidate for Mitosis and can be reused elsewhere as a independent component

ramonmulia: yes, I think that's a good idea.
I created a task for it and tagged you there.

packages/core/atoms/Divider.lite.tsx
```diff
@@ -28,7 +28,7 @@ export default function Divider(props: Props) {
   );
 }
 
-const dividerStyles = tv({
+export const dividerStyles = tv({
```
DmitryB-NF: I suppose it is not used publicly anymore

packages/react/src/autocomplete/autocomplete.stories.tsx
```diff
@@ -376,12 +399,252 @@ export const Test: Story = {
...
+      const secondOption = canvas.getByRole('option', { name: labelOptions[1].label });
+      await userEvent.click(secondOption);
+      await waitFor(() => {
+        expect(canvas.getByText('2')).toBeInTheDocument();
+      });
+    });
+  },
+};
+
+const getLabel = (value: string) => labelOptions.find((option) => option.value === value)?.label ?? value;
```
DmitryB-NF: Just in case, it could be inlined for 2 uses or declared right before use for better readability.
You can also use lodash for inlining `_.find(labelOptions, { value })`

packages/react/src/autocomplete/autocomplete.tsx
DmitryB-NF: @ramonmulia could you check this one as well?

packages/react/src/autocomplete/autocomplete.tsx
```diff
@@ -165,33 +203,40 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>((pro
     propagateOnBlur(onAutocompleteBlur, name)(current);
   };
 
+  const effectiveOnChange = multiple ? handleOnToggleItem : handleOnSelectItem;
```
DmitryB-NF: Just in case, I think `handleChange` or `onChange` (as it is) could work a bit better because the component is huge and `effectiveOnChange` besides all other `onChange` handlers a bit readable to track

packages/react/src/autocomplete/use-autocomplete-controller.ts
```diff
@@ -70,6 +68,19 @@ const reducer = (state: AutocompleteState, action: AutocompleteAction): Autocomp
     case SET_OPTION_TYPE: {
       return { ...state, optionType: action.payload };
     }
+    case TOGGLE_SELECTED_ITEM: {
```
DmitryB-NF: If removing the duplication logic in `handleOnToggleItem` (https://github.com/ogcio/govie-ds/pull/993/changes#diff-450fd22df1057de4c44323d5f3c9a4edfb74d9df08abd2cbb24ed939d168fd3bR176), we can remove the handler at all and just use `SET_SELECTED_VALUES` directly

packages/react/src/select/select-menu.tsx
```diff
@@ -101,9 +118,31 @@ export const SelectMenu = forwardRef<HTMLDivElement, SelectMenuProps>(
...
+            <>
+              <div
+                role="button"
+                tabIndex={-1}
+                className={cn('gi-select-option-item', {
+                  'gi-select-option-item-highlighted': clearAllHighlighted,
+                })}
+                onClick={onClearAll}
+              >
+                <Text size="sm">{t('autocomplete.clearAll', { defaultValue: 'Clear all selections' })}</Text>
```
DmitryB-NF: It would be really nice to have a new prop to declare this message outside in the client application. If a client has its own i18n system, it would be a good extension to use it instead of a built-in one

packages/react/src/select/select-menu.tsx
```diff
@@ -101,9 +118,31 @@ export const SelectMenu = forwardRef<HTMLDivElement, SelectMenuProps>(
...
         </div>
       ) : (
-        <ul aria-label="options" role="listbox">
-          {filteredOptions}
-        </ul>
+        <>
+          {onClearAll && (
+            <>
+              <div
+                role="button"
```
DmitryB-NF, line 128: I think we may need to add `id={listboxId ? `${listboxId}-clear-all` : undefined}` here to still highlight the button if it is possible to navigate to it. This way, adding an active like
```
aria-activedescendant={
   state.highlightedIndex === CLEAR_ALL_INDEX
      ? `${listboxId}-clear-all`
      : state.highlightedIndex >= 0
        ? `${listboxId}-option-${state.highlightedIndex}`
        : undefined
}
```
will improve the a11y a bit to make sure that this element is actually active (it could be selected by arrows but not by tab)

DmitryB-NF: As a separate option, we can convert this div into a proper button to be fully selectable (could be useful) but we need to make sure it works properly with the menu

packages/react/src/ChipGroup.tsx
```diff
@@ -0,0 +1,33 @@
+'use client';
```
DmitryB-NF: Client can be removed as the component will work fully on the the server as there is no state

packages/react/src/autocomplete/autocomplete.tsx
DmitryB-NF: @ramonmulia this one is still valuable as ` id={listboxId ? `${listboxId}-option-${optionProps.value}` : undefined}` has a different id in https://github.com/ogcio/govie-ds/pull/993/changes#diff-450fd22df1057de4c44323d5f3c9a4edfb74d9df08abd2cbb24ed939d168fd3bR369

DmitryB-NF (changes requested):
The implementation looks good. 
We have only a potential issue with `id`s for Menu options. 
I also added a couple of potential improvements for readability and maintenance as an optional action

packages/react/src/chip/chip.tsx
```diff
@@ -25,6 +25,7 @@ export const Chip = ({ label, className, onClose = () => null }: ChipProps) => {
 
   return (
     <div
+      role="group"
```
DmitryB-NF: I think this is redundant for a single element as there is a `group` Chip inside a `group`  ChipGroup

ramonmulia: good call!

apps/docs/content/3-components/2-library/autocomplete/react.mdx
```diff
@@ -144,6 +196,36 @@ The Autocomplete component allows users to search and select from a list of opti
...
+  componentId="autocomplete-react"
+  story="with-multiple-chips"
+/>
+
+### Multi-select with Chips (Collapse)
+
+When `maxVisible` is set, only that many chips are shown. A `formatOverflow` function provides the overflow label text (e.g. "+2 more").
+
+<StorybookFrame
+  heightClassName="h-[550px]"
```
DmitryB-NF: 💡 I am not sure this works as expected because there is no `heightClassName` prop in the `govie-ds/apps/docs/src/components/document/components/storybook-frame.tsx`. So the output still has a 128px height. Should it be passed as a `className`?

ramonmulia: good call, It had been there for a while in a few other stories and I didn't notice. replaced it with `className`

DmitryB-NF (approved):
Nice and clean!

PR #996, opened 2026-06-25
feat(AB#39585): mitosis footer logo wrapper

Ademsk1:
## Description

Adds Footer Logo atom component to react and angular through Mitosis. 

## Type of Issue

- [x] 🚀 Feature
- [ ] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#39585](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/39585)


packages/core/atoms/FooterLogo.lite.tsx
```diff
@@ -0,0 +1,24 @@
...
+export type Props = {
+  children: any;
+  id?: string;
+  className?: string;
+  dataTestId?: string;
+};
+
+export default function FooterLogo(props: Readonly<Props>) {
+  return (
+    <GiBox id={props.id} className={styles({ class: props.className })} dataTestId={props.dataTestId}>
```
ramonmulia: since FooterLogo has no variants, `tv()` adds overhead without benefit here, I think a template literal would be sufficient: ```className={`gi-min-w-fit md:gi-ml-auto gi-mt-8 md:gi-mt-0 ${props.className || ''}`}```
 
this matches how `Footer.lite` handles it (no tv(), just direct class forwarding).

packages/core/atoms/FooterLogo.lite.tsx
```diff
@@ -0,0 +1,24 @@
...
+useMetadata({ angular: { selector: 'gi-footer-logo' } });
+
+export type Props = {
+  children: any;
+  id?: string;
+  className?: string;
+  dataTestId?: string;
+};
+
+export default function FooterLogo(props: Readonly<Props>) {
```
ramonmulia: is `Readonly` necessary here?

packages/core/atoms/FooterLogo.lite.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { tv } from 'tailwind-variants';
+import GiBox from './Box.lite';
+
+useMetadata({ angular: { selector: 'gi-footer-logo' } });
+
+export type Props = {
+  children: any;
+  id?: string;
+  className?: string;
+  dataTestId?: string;
```
ramonmulia, line 12: `styles?` is missing from Props, all sibling atoms expose it, I think it could be added here too.

packages/core/atoms/FooterLogo.lite.tsx
Ademsk1: Sonarqube was complaining about it ... I can remove if we like?

packages/core/atoms/FooterLogo.lite.tsx
Ademsk1: This is true, but it won't handle the tailwind merging of classes. So consumers won't be able to override the margin values.

packages/core/atoms/FooterLogo.lite.tsx
ramonmulia: doesn't hurt to have it, I guess.

Ademsk1: @DmitryB-NF interested on your opinion here ... we should stay consistent and be in agreement  for future components

packages/core/atoms/FooterLogo.lite.tsx
DmitryB-NF: TV merge will not work even if adding `tv` usage. We need to explicitly add a configuration for merging custom `gi-` styles. Otherwise, it is an overhead without direct benefits. We can use it for consistency but I would suggest avoding it where possible because we still need a better solution in place to apply proper setup

packages/core/atoms/FooterLogo.lite.tsx
DmitryB-NF: Readonly Props is a general recommendation, but it is a false alarm for Mitosis, as it is not a direct React code. I would say that consistency is better, especially when `Readonly<Props>` is verbose without any potential benefits

DmitryB-NF (dismissed):
The component is nice and clean.
@Ademsk1 could you check unresolved threads?

packages/core/atoms/FooterLogo.lite.tsx
```diff
@@ -0,0 +1,30 @@
...
+      styles={props.styles}
+      dataTestId={props.dataTestId}
+    >
+      {props.children}
+    </GiBox>
+  );
+}
+
+const footerStyles = tv({
+  base: 'gi-w-fit md:gi-ml-auto gi-mt-8 md:gi-mt-0',
```
DmitryB-NF, line 29: ❓ `gi-rounded-sm` is not presented here, and there is a change `lg:gi-ml-auto` → `md:gi-ml-auto`. Is it expected for the new Logo?

Ademsk1: Yes, the change is expected - I aimed to match the Figma perfectly, and for the examples @LexaGallery provided, we have an auto margin on screen widths `md` and greater - not `lg` as it was previously. 

For `gi-rounded-sm`, I don't see a good reason for keeping it. I believe the previous implementation added it so that when the footer logo (when it was a link) was focused, the outline edges wouldn't be so sharp.

packages/core/atoms/FooterLogo.lite.tsx
Ademsk1: removing :D

packages/core/atoms/FooterLogo.lite.tsx
Ademsk1: When we **do** find a configuration solution for this, we will still need to merge them so that the last one wins right? If I did not use merging, here's the example I had in mind:
1. A user passed in `mt-4` to override the default margin for screen widths < md
2. Final className = `... gi-mt-8 mt-4`
3. Configuration fixed, styles treat `gi-mt-8` the same as `mt-8`
4. className= `mt-8 mt-4`
5. mt-8 still wins because it appears later on in the stylesheet

Ademsk1: Unless its the case that we don't want the user to override these base styles at all...

DmitryB-NF (approved):

PR #999, opened 2026-06-30
fix(AB#40245): divider orientation

Ademsk1:
## Description

Divider orientation on Angular was breaking the rest of the DOM structure for Angular. Enforced aria orientation to either be horizontal or vertical - per the [mdn docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation). 

Updated the story so we catch this if it happens again. Matched it with React. 

## Type of Issue

- [ ] 🚀 Feature
- [x] 🐛 Bug
- [ ] 🔧 Chore
- [ ] 🌐 Docs

## Checklist

## Related Issues

[AB#40245](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/40245)


packages/angular/src/stories/Divider.stories.ts
```diff
@@ -16,16 +17,16 @@ export const Horizontal: StoryObj = {
   ...stories.Horizontal,
   render: (props) => ({
     props,
-    moduleMetadata: { imports: [Box, Divider, Stack] },
+    moduleMetadata: { imports: [Box, Divider, Stack, Link] },
     template: `
       <gi-stack [gap]="2" [direction]="orientation === '${Orientation.VERTICAL}' ? 'row' : 'column'" className="gi-font-primary gi-text-sm">
-        <gi-box>Content</gi-box>
+        <gi-box><gi-link href="#" variant="inline">Above</gi-link></gi-box>
```
DmitryB-NF: ⚠️ Does it make sense to create a separate Story to test this? I am not sure we should chnage the base ones especially because they are presented in the Docs website and the view and the link may not work as expected

packages/react/src/stories/Divider.stories.tsx
```diff
@@ -27,9 +28,13 @@ export const Horizontal: Story = {
   ...stories.Horizontal,
   render: (props: DividerProps) => (
     <Stack gap={2} direction={props.orientation === Orientation.VERTICAL ? 'row' : 'column'}>
-      <Box>Content</Box>
+      <Box>
+        <Link href="#">Above</Link>
```
DmitryB-NF: ⚠️ Does it make sense to create a separate Story to test this? I am not sure we should change the base ones, especially because they are presented on the Docs website, and the view and the link may not work as expected

packages/core/atoms/Divider.lite.tsx
```diff
@@ -18,7 +18,7 @@ export default function Divider(props: Props) {
     <hr
       id={props.id}
       data-testid={props.dataTestId}
-      aria-orientation={getOrientation(props.orientation) === Orientation.VERTICAL ? Orientation.VERTICAL : undefined}
+      aria-orientation={getOrientation(props.orientation)}
```
DmitryB-NF, line 21: @ramonmulia could you review this? Do you remember the original reason for having this logic?

packages/angular/src/stories/Divider.stories.ts
```diff
@@ -49,3 +50,22 @@ export const Vertical: StoryObj = {
     `,
   }),
 };
+
+export const RichText: StoryObj = {
+  ...stories.Vertical,
+  render: (props) => ({
```
DmitryB-NF, line 56: ⚠️ Does it make sense to move this into the Core Storybook, not to extend `Vertical` and avoid any unexpected dependencies

packages/react/src/stories/Divider.stories.tsx
```diff
@@ -27,9 +28,13 @@ export const Horizontal: Story = {
   ...stories.Horizontal,
   render: (props: DividerProps) => (
     <Stack gap={2} direction={props.orientation === Orientation.VERTICAL ? 'row' : 'column'}>
-      <Box>Content</Box>
+      <Box>
+        <Link href="#">Content</Link>
+      </Box>
```
DmitryB-NF: ⚠️ Should we revert this as well?

DmitryB-NF (changes requested):

packages/core/atoms/Divider.lite.tsx
ramonmulia: that's because it already has a default semantic role that includes this orientation natively (horizontal), and also because browsers already expects and communicates that a separator is horizontal, adding `aria-orientation="horizontal"` for `hr` adds no new information for screen readers.

for example: `aria-orientation="horizontal"` would be necessary if we have something like:
 `<div role="separator" aria-orientation="horizontal">` as it is a `div` with role separator it doesn't have `aria-orientation` by default.

I think we should use `aria-*` only when we really need to, to not add any reduntant information for assistive technologies.

DmitryB-NF: According to the docs, the `<hr>` tag has a native separator role and has `horizontal` as an `aria-orientation` default value. So passing the default value explicitly has no effect but simplifies the code and the Mitosis logic.

DmitryB-NF (approved):

